PR #36, opened 2015-01-03
working app listing check

aaustin:
@qinweigong 

This works differently than the Android version. Instead of harvesting a list, it will check an existing list of URI schemes that we have in the database. Currently, it's a very short and simple list stored in the API. We'll use the Android harvesting to accumulate a rich data set of URI schemes


Branch-SDK/Branch-SDK/BNCConfig.h
```diff
@@ -24,7 +24,7 @@
 #endif
 
 #ifdef BNC_DEV_ENV
-#define BNC_API_BASE_URL        @"http://localhost:3001"
+#define BNC_API_BASE_URL        @"http://localhost:3000"
```
qinweigong: so dashboard is using port 3001?

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1208,7 +1236,9 @@ - (void)serverCallback:(BNCServerResponse *)response {
                     [BNCPreferenceHelper setInstallParams:[response.data objectForKey:DATA]];
                 }
             }
-            
+            if ([BNCPreferenceHelper getNeedAppListCheck]) {
+                [self processListOfApps];
+            }
```
qinweigong: why do we need to call this again here? (same as line 1210-1212?)

Branch-SDK/Branch-SDK/BNCConfig.h
aaustin: I just use localhost 3000 for my API when I'm testing. No big deal if you prefer to keep it at 3001

Branch-SDK/Branch-SDK/Branch.m
aaustin: It gets checked and added in both 'install' and 'open' callbacks, as scans the apps every 4 days or so

qinweigong: Got it!

qinweigong:
:+1:

PR #38, opened 2015-01-14
Moved app key to plist; added ChangeLog

qinweigong:
@aaustin


aaustin:
So this is close but it doesn't solve the actual problem. The problem statement is that our current SDK implementation stores the app id in NSUserDefaults, which can be cleared. If a user clears the settings for one of our apps, it will start sending bnc_no_value in app key (and every other field). Does that make sense?

Facebook and other SDKs have the user just put the app key into the plist file, and then the SDK internal retrieve it from the plist automatically. It's never stored in NSUserDefaults. The developer never needs to retrieve it from the plist file themselves. Does that make sense?

Note, this also applies to identity, session and fingerprint ids as well, but it's probably fine to leave at is for now. Eventually, we should be saving those to a file instead of defaults so users can't clear it.

qinweigong:
Oh, got it!
So, I also need to change the getAppKey and setAppKey functions in BNCPreferenceHelper, correct?

aaustin:
Let's make sure to support legacy implementations for now if we can't get
the key from the plist file - so don't gut it

qinweigong:
I have the backward compatibility code in getAppKey. Please review my latest commit.

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -100,20 +100,9 @@ @implementation Branch
...
-    
-    if (!currInstance) {
-        [Branch initInstance];
-    }
-    
-    return currInstance;
-}
-
 + (Branch *)getInstance {
     if (!currInstance) {
```
aaustin: can we add some key validation and warning code? I want to make sure forgetting the plist step is really obvious and the the logs will tell them what to do

qinweigong: Good point!

aaustin:
:+1:

qinweigong:
I've added more change logs. Please take a peek.

aaustin:
:+1: looks good!

PR #40, opened 2015-01-24
Cache link for same params

qinweigong:
@aaustin


aaustin:
Getting some errors when I have no connection - doesn't crash though
2015-01-26 17:35:04.722 Branch-TestBed[3475:2492154] -[BNCLinkData encodeWithCoder:]: unrecognized selector sent to instance 0x170099f00

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -775,40 +772,44 @@ - (void)generateShortUrl:(NSArray *)tags andAlias:(NSString *)alias andType:(Bra
...
+            if (self.initFinished || !self.hasNetwork) {
+                self.lastRequestWasInit = NO;
+                [self processNextQueueItem];
+            } else if (self.initFailed || self.initNotCalled) {
+                [self handleFailure:[self.requestQueue size]-1];
+            }
+        } else {
+            if (callback) {
+                dispatch_async(dispatch_get_main_queue(), ^{
+                    if (callback) callback([self.linkCache objectForKey:post], nil);
```
shortstuffsushi: Double checking for callback here, it appears.

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -775,40 +772,44 @@ - (void)generateShortUrl:(NSArray *)tags andAlias:(NSString *)alias andType:(Bra
...
-            [self.requestQueue enqueue:req];
-        }
+        [post setupParams:args];
         
-        if (self.initFinished || !self.hasNetwork) {
-            self.lastRequestWasInit = NO;
-            [self processNextQueueItem];
-        } else if (self.initFailed || self.initNotCalled) {
-            [self handleFailure:[self.requestQueue size]-1];
+        if (![self.linkCache objectForKey:post]) {
```
shortstuffsushi: The `callback` is never called on this branch of the if.

Branch-SDK/Branch-SDK/BNCLinkCache.m
```diff
@@ -0,0 +1,41 @@
...
+- (id)init {
+    self = [super init];
+    if (self) {
+        self.cache = [[NSMutableDictionary alloc] init];
+    }
+    return self;
+}
+
+- (void)setObject:(NSString *)anObject forKey:(BNCLinkData *)aKey {
+    for (id key in [self.cache allKeys]) {
```
shortstuffsushi: With this check, the cache can never be updated. Not sure if that's an issue or not, but the old object will always be kept for the same key.

Branch-SDK/Branch-SDK/BNCLinkCache.m
```diff
@@ -0,0 +1,41 @@
...
+            return;
+        }
+    }
+    [self.cache setObject:anObject forKey:aKey];
+}
+
+- (NSString *)objectForKey:(BNCLinkData *)aKey {
+    for (id key in [self.cache allKeys]) {
+        BNCLinkData *data = (BNCLinkData *)key;
+        if ([data hash] == [aKey hash]) {
```
shortstuffsushi: You could push the value of `[aKey hash]` out of this loop, so it doesn't have to be re-evaluated each time around.

qinweigong: Good idea! Not to rely on Apple's compiler to optimize it :)

Branch-SDK/Branch-SDK/Branch.m
qinweigong: In this case, the callback will be called asynchronously.

Branch-SDK/Branch-SDK/BNCLinkCache.m
qinweigong: Hmm, that's valid. Though with the cache check before calling API to generate link, it will never need to overwrite value for an existing key.

shortstuffsushi: True. With that in mind, is there a reason to implement this as a dictionary like structure? An array might be a better structures, since really all you need is an array of the hashes.

Branch-SDK/Branch-SDK/BNCLinkCache.m
```diff
@@ -0,0 +1,36 @@
...
+    }
+    return self;
+}
+
+- (void)setObject:(NSString *)anObject forKey:(BNCLinkData *)aKey {
+    [self.cache setObject:anObject forKey:aKey];
+}
+
+- (NSString *)objectForKey:(BNCLinkData *)aKey {
+    NSUInteger hashcode = [aKey hash];
```
shortstuffsushi: With your change to `setObject:forKey`, will it be necessary to have the for loop at all here?

qinweigong: Yes, it's still needed. 
return [self.cache objectForKey:aKey]; won't work actually.

shortstuffsushi: Could you explain why? You're inserting the object keyed on `aKey` in the method above, why wouldn't you be able to retrieve it the same way?

qinweigong: Something odd with Objective C's NSDictionary. It calls the hash function and returns the same hashcode, but [self.cache objectForKey:aKey] just returns nil

qinweigong: Or maybe I am not implementing it correctly? Any advice?

shortstuffsushi: Hmm, tbh I guess I'm not sure. It seems like that should work to me, but I haven't had a chance to try it yet. However, if you instead just keyed on the hash value (`@([aKey hash])`), you could avoid the need for that check entirely (and it would be O(1) vs O(n)).

qinweigong: Good idea!

shortstuffsushi:
I prefer the `@(numberValue)` syntax to `[NSNumber numberWithXxx:`, but other than that, this appears good to go from what I can tell :+1:

Branch-SDK/Branch-SDK/Branch.m
shortstuffsushi: Forgot to ever respond to this, but makes sense I think.

aaustin:
:+1:

PR #41, opened 2015-01-26
check if long press has been added already

qinweigong:
@aaustin


Branch-TestBed/Branch-TestBed.xcodeproj/project.pbxproj
```diff
@@ -208,7 +203,7 @@
 				ORGANIZATIONNAME = "Branch Metrics";
 				TargetAttributes = {
 					6700165F1940F51400A9E103 = {
-						DevelopmentTeam = 6B2DFWM843;
```
shortstuffsushi: Probably don't want to change this.

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1419,4 +1427,78 @@ - (void)serverCallback:(BNCServerResponse *)response {
...
+
+- (void)bnc_addCancelDebugGestureRecognizer {
+    [self bnc_addGesterRecognizer:@selector(bnc_endDebug:)];
+}
+
+- (void)bnc_addGesterRecognizer:(SEL)action {
+    BNCLongPress = [[UILongPressGestureRecognizer alloc] initWithTarget:self action:action];
+    BNCLongPress.cancelsTouchesInView = NO;
+    BNCLongPress.minimumPressDuration = BNCDebugTriggerDuration;
+    if (![BNCSystemObserver isSimulator]) {
```
shortstuffsushi: If you're making use of both conditions here (is and isn't), why check "not" simulator? Just put the simulator scenario first.

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -835,6 +841,8 @@ - (void)applicationDidBecomeActive {
             }
         });
     }
+    
+    [self bnc_addDebugGestureRecognizer];
```
shortstuffsushi: You'll need a corresponding remove in the `applicationWillResignActive` to avoid multiple gestureRecognizers being added.

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1419,4 +1427,78 @@ - (void)serverCallback:(BNCServerResponse *)response {
...
+- (void)bnc_addDebugGestureRecognizer {
+    [self bnc_addGesterRecognizer:@selector(bnc_connectToDebug:)];
+}
+
+- (void)bnc_addCancelDebugGestureRecognizer {
+    [self bnc_addGesterRecognizer:@selector(bnc_endDebug:)];
+}
+
+- (void)bnc_addGesterRecognizer:(SEL)action {
+    BNCLongPress = [[UILongPressGestureRecognizer alloc] initWithTarget:self action:action];
```
shortstuffsushi: I'm not 100% sure how this works, but if you're adding this to the window, I believe that other gesture recognizers added by the app developers will be ignored. I'm not sure if the "cancelsTouchesInView" will pass through and call both yours and theirs or not.

aaustin: Maybe Qinwei can chime in here, but I believe he found this to be used by Mixpanel for their debugging option

Branch-TestBed/Branch-TestBed.xcodeproj/project.pbxproj
qinweigong: This is changed by Xcode. How to prevent it from being changed?

shortstuffsushi: You can't prevent Xcode from changing it (unless you have that team available in Xcode), but you can certainly just not commit ;)

qinweigong: What's the consequence of having this changed and committed? We've never paid attention to this value, and always let it be so far.

shortstuffsushi: If you guys don't have one specific team that you use, then it's not a big deal. At that point, just a churn things. I usually try not to commit things that are not related to the pull request. In this case, the team isn't related to anything, and probably shouldn't be committed at all, or could be committed as "None" to prevent this churn.

Branch-SDK/Branch-SDK/Branch.m
qinweigong: From the documentation: If a gesture recognizer doesn’t recognize its gesture or if the value of this property is NO, the view receives all touches in the multi-touch sequence.

Branch-SDK/Branch-SDK/Branch.m
qinweigong: added

Branch-SDK/Branch-SDK/Branch.m
qinweigong: Good catch! Replaced with gesture recognizer delegate method shouldRecognizeSimultaneouslyWithGestureRecognizer.

aaustin:
:+1:

PR #43, opened 2015-01-28
UIActivityItemProvider Subclass

scotthasbrouck:
Adds Branch class methods which return a UIActivityItemProvider subclass, with arguments for Branch params and properties. The subclass generate a Branch short URL that is auto tagged with the channel based on the sharing activity  the user selected. The callback for the short URL is fired on a thread other than the main thread, however, the main thread waits for the callback to return and provide the UIActivityItem. This implementation boils down sharing Branch URLs with a UIActivityView share sheet to just a few lines of code.


aaustin:
Oh my - this is kick ass

aaustin:
:+1: super excited

qinweigong:
Nice! :+1:

PR #53, opened 2015-02-18
Empty tests

qinweigong:
Pre-mature PR, not intended for review!


qinweigong:
@aaustin @shortstuffsushi 
Need some final touches and, of course, more test cases moving forward.

aaustin:
Awesome! Looking forward to it

Branch-SDK/Branch-SDK/BNCPreferenceHelper.h
```diff
@@ -94,5 +94,6 @@ static NSString *NO_STRING_VALUE = @"bnc_no_value";
 + (void)sendScreenshot:(NSData *)data;
 + (void)setDebugConnectionDelegate:(id<BNCDebugConnectionDelegate>) debugConnectionDelegate;
 + (void)keepDebugAlive;
++ (BOOL)getRemoteDebug;
```
shortstuffsushi: Rather than making a method for this, couldn't you just use `extern`?

Branch-SDK/Branch-SDK/Branch.h
```diff
@@ -87,6 +87,7 @@ typedef enum {
                                                            andAlias:(NSString *)alias;
 
 + (void)setDebug;
+- (void)simulateInitFinished;
```
shortstuffsushi: I'm sure this is only used for testing, but I feel like we shouldn't expose this method. Is there another way we could trigger this without exposing it?

Branch-TestBed/Branch-SDK Debugger Tests/Branch_SDK_Debugger_Tests.m
```diff
@@ -0,0 +1,65 @@
...
+
+#import <UIKit/UIKit.h>
+#import <XCTest/XCTest.h>
+#import "BNCPreferenceHelper.h"
+#import "BranchServerInterface.h"
+#import "Nocilla.h"
+
+@interface Branch_SDK_Debugger_Tests : XCTestCase {
+    
+@private
```
shortstuffsushi: Why not properties?

Branch-TestBed/Branch-SDK Debugger Tests/Branch_SDK_Debugger_Tests.m
```diff
@@ -0,0 +1,65 @@
...
+}
+
+@end
+
+@implementation Branch_SDK_Debugger_Tests
+
+- (void)setUp {
+    [super setUp];
+    
+    [[LSNocilla sharedInstance] start];
```
shortstuffsushi: Note that you're starting and stopping Nocilla between each test. In the Github example, it suggests doing so at the start and stop of the suite (the `+ (void)setup` instead of `- (void)setup`.

Branch-TestBed/Branch-SDK Debugger Tests/Branch_SDK_Debugger_Tests.m
```diff
@@ -0,0 +1,65 @@
...
+    
+    XCTAssertTrue([BNCPreferenceHelper getRemoteDebug]);
+}
+
+- (void)testConnectFail {
+    NSDictionary *responseDict = @{@"error": @{@"code": @465, @"message": @"Server not listening"}};
+    NSData *responseData = [BNCServerInterface encodePostParams:responseDict];
+    
+    stubRequest(@"POST", [BNCPreferenceHelper getAPIURL:@"debug/connect"])
+    .andReturn(465)
```
shortstuffsushi: This shouldn't be the response code, I don't think. It should probably be a 400.

Branch-TestBed/Branch-SDK Load Tests/Branch_SDK_Load_Tests.m
```diff
@@ -0,0 +1,78 @@
...
+    
+    stubRequest(@"POST", [BNCPreferenceHelper getAPIURL:@"url"])
+    .andReturn(200)
+    .withHeaders(@{@"application/json": @"Content-Type"})
+    .withBody(responseData);
+    
+    XCTestExpectation *getShortURLExpectation = [self expectationWithDescription:@"Test getShortURL"];
+    
+    for (int i = 0; i < 1000; i++) {
+        [branch getShortURLWithParams:nil andChannel:[NSString stringWithFormat:@"%d", i] andFeature:nil andCallback:^(NSString *url, NSError *error) {
```
shortstuffsushi: Could you explain what this test is actually doing? It seems to be executing a 'get link' a thousand times, but I don't think that in itself is terribly valuable. The expectation is that this will return in < 1 second, but this won't work for two reasons.

* The for loop itself makes 1000 async requests, which doesn't guarantee completion before the final get url is called, so the only thing you're actually assuring there is that the last request finishes in less than one second
* Testing time is a bad idea in general because any system quirk (say the CPU on the machine is hung for a second on any arbitrary process) could cause the test to fail -- it's not a 'determinate' test.

Branch-TestBed/Branch-SDK test/Branch_SDK_Debugger_Tests.m
```diff
@@ -0,0 +1,65 @@
...
+}
+
+@end
+
+@implementation Branch_SDK_Debugger_Tests
+
+- (void)setUp {
+    [super setUp];
+    
+    [[LSNocilla sharedInstance] start];
```
shortstuffsushi: This pattern seems to be repeated through a number of tests, could a Test Util class be created instead?

Branch-TestBed/Branch-SDK test/Branch_SDK_Debugger_Tests.m
```diff
@@ -0,0 +1,65 @@
...
+}
+
+- (void)tearDown {
+    [[LSNocilla sharedInstance] clearStubs];
+    [[LSNocilla sharedInstance] stop];
+    
+    [super tearDown];
+}
+
+- (void)testConnectSucceed {
```
shortstuffsushi: Is this the same set of tests as above?

Branch-TestBed/Pods/Headers/Private/Nocilla/ASIHTTPRequestStub.h
```diff
@@ -0,0 +1 @@
+../../../Nocilla/Nocilla/Hooks/ASIHTTPRequest/ASIHTTPRequestStub.h
```
shortstuffsushi: I'm of the opinion that the Pods directory shouldn't be checked in. I'll leave that open ended, but I don't really think it should be.

Branch-TestBed/Branch-TestBed/ViewController.h
```diff
@@ -10,4 +10,7 @@
 
 @interface ViewController : UIViewController
 
+@property (nonatomic, weak) IBOutlet UIButton *refreshUrlButton;
```
shortstuffsushi: Why did these move from private to public?

Branch-TestBed/Branch-SDK Debugger Tests/Branch_SDK_Debugger_Tests.m
qinweigong: Hah, copied from apple developer site

Branch-TestBed/Branch-SDK Debugger Tests/Branch_SDK_Debugger_Tests.m
qinweigong: Haha, this is the special error code returned from API server side indicating that it's not listening.

Branch-TestBed/Branch-SDK Load Tests/Branch_SDK_Load_Tests.m
qinweigong: No, it's not testing time. It's a load test mainly on the queue. Though they are all async requests, remember we have a queue and all requests are queued and persisted. As long as the last request completes successfully, that means the queue hasn't crashed in the load test.

Branch-TestBed/Branch-SDK test/Branch_SDK_Debugger_Tests.m
qinweigong: Will optimize later.

Branch-TestBed/Branch-SDK test/Branch_SDK_Debugger_Tests.m
qinweigong: Yes, initially I put this test class in the same test target as the functionality tests. But seems they will interfere each other. So I disabled it here and moved out to a separate test target. However, I couldn't find a way to delete unnecessary test class from a test target!

Branch-TestBed/Branch-TestBed/ViewController.h
qinweigong: Was trying to manipulate them in the tests.

Branch-TestBed/Branch-SDK Debugger Tests/Branch_SDK_Debugger_Tests.m
shortstuffsushi: Is that the Branch status, though, or the HTTP Status? Afaik, 465 doesn't exist anywhere (I couldn't find reference to it), could you point me towards it?

qinweigong: Yes, it's Branch status.

shortstuffsushi: In that case, you should use an HTTP status code for the `andReturn` method, since that's going to be presented as the response status.

Branch-TestBed/Branch-SDK Load Tests/Branch_SDK_Load_Tests.m
shortstuffsushi: Hmm, I'm still not sure there is going to be any value in this test. Even though you're not aiming to test speed, you technically are, because that `waitForExpectations` method will cause the test to fail if it takes more than a second (which, as mentioned, should theoretically never happen, but it's not determinate). Also, what features of the Queue are you testing by filling it with requests; that it can hold 1001 requests? I'm just not sure this test is going to actually prove anything for us.

Branch-TestBed/Branch-SDK test/Branch_SDK_Debugger_Tests.m
shortstuffsushi: I'm not sure I understand this (I still need to checkout this Branch locally to see how it runs instead of just examining code), but it seems that you've created two targets, and now have code duplicated between both. Is that correct? And if so, can it be removed?

Branch-TestBed/Branch-SDK Debugger Tests/Branch_SDK_Debugger_Tests.m
qinweigong: Actually, it doesn't matter what status code here if I use stubbing.

Branch-TestBed/Branch-SDK test/Branch_SDK_Debugger_Tests.m
qinweigong: That's exactly what I have been searching for: how to remove a test class from a test target! Seems there's no remove button!

Branch-TestBed/Branch-SDK Debugger Tests/Branch_SDK_Debugger_Tests.m
shortstuffsushi: It is still important to return a valid status code, as the HTTP handling library that sends these requests that are stubbed still builds up a response as though this was a real request. Say, for example, that you wanted to test a "failure" where Branch return a 299 status (not a valid HTTP code), but say as well that the 299 status of Branch indicated failure. The underlying HTTP lib would say "299 is < 400, so that request must have succeeded" and would fall into your success handler.

Branch-TestBed/Branch-SDK test/Branch_SDK_Debugger_Tests.m
shortstuffsushi: Hmm, I'm not sure exactly what you're referring to. I'll check out the actual branch quick and take a look.

Branch-TestBed/Branch-SDK Load Tests/Branch_SDK_Load_Tests.m
qinweigong: I am not sure about the timeout either, but run it and you will see how it works. Yah, the goal is load test the queue.

qinweigong:
Seems CircleCI requires a Podfile, otherwise it'll fail, as you can see.

aaustin:
Qinwei - can we test with different versions, like iOS6?

shortstuffsushi:
Due to some issues with the current project architecture, ability to test this project is somewhat limited. After experimenting for a couple of days trying to get things running in a single project/scheme/etc, @qinweigong and I have just decided to let it be the way it is for now. We will need to do some additional work in the future, though.

aaustin:
This looks great to me! Can you go in asana and make a list of all tests that are top of mind that we should fill this framework with? :+1:

PR #54, opened 2015-02-18
Make sync calls inside interface

qinweigong:
@aaustin @shortstuffsushi

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCServerInterface.m
+++ b/Branch-SDK/Branch-SDK/BNCServerInterface.m
@@ -18,7 +18,7 @@
     [self getRequestAsync:params url:url andTag:requestTag log:YES];
 }
 
-// this is actually a synchronous call, since its callers are NOT on the main queue
+// this is actually a synchronous call; it should NOT be called from the main queue
 - (void)getRequestAsync:(NSDictionary *)params url:(NSString *)url andTag:(NSString *)requestTag log:(BOOL)log {
     BNCServerResponse *serverResponse = [self genericSyncHTTPRequest:[self prepareGetRequest:params url:url log:log] withTag:requestTag andLinkData:nil];
     if (self.delegate) [self.delegate serverCallback:serverResponse];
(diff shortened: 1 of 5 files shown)


Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -18,8 +18,10 @@ - (void)getRequestAsync:(NSDictionary *)params url:(NSString *)url andTag:(NSStr
     [self getRequestAsync:params url:url andTag:requestTag log:YES];
 }
 
+// this is actually a synchronous call, since its callers are NOT on the main queue
```
shortstuffsushi: If these two calls are actually synchronous, renaming them would probably be best. They're not publicly exposed anyway, right?

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -886,12 +886,13 @@ - (void)applyReferralCode:(NSString *)code andCallback:(callbackWithParams)callb
...
-        req.tag = REQ_TAG_GET_CUSTOM_URL;
-        BNCLinkData *post = [self prepareLinkDataFor:tags andAlias:alias andType:type andChannel:channel andFeature:feature andStage:stage andParams:params];
-        
-        if (![self.linkCache objectForKey:post]) {
+    
+    BNCServerRequest *req = [[BNCServerRequest alloc] init];
+    req.tag = REQ_TAG_GET_CUSTOM_URL;
+    BNCLinkData *post = [self prepareLinkDataFor:tags andAlias:alias andType:type andChannel:channel andFeature:feature andStage:stage andParams:params];
+    
+    if (![self.linkCache objectForKey:post]) {
```
shortstuffsushi: Just a note on shorthand, in case I hadn't mentioned this to you before, `self.linkCache[post]` could be used in place of `objectForKey:`

qinweigong: Haha, you paranoid :-) Yup, I know this shorthand for NSDictionary, but self.linkCache here is actually a BNCLinkCache object.

Branch-SDK/Branch-SDK/BNCServerInterface.m
qinweigong: Hmm, makes sense.

shortstuffsushi:
@qinweigong and I discussed this a bit further in person, and determined that a larger refactor may be in order to make all calls "sync" in that they no longer take in a delegate. However, that is not going to be a part of this pull request, so :thumbsup: for the rest of this.

PR #56, opened 2015-02-26
added app_id param to get requests

qinweigong:
@dmitrig01 @aaustin


dmitrig01:
I can't actually speak to the functionality, but:

@aaustin - we need to include the app id on all requests going forward, so we know which shard to queries to. for now, we will send legacy requests to all three shards and see which one gives us back data.

aaustin:
ok cool - that sounds really ugly

aaustin:
:+1:

dmitrig01:
yep see https://github.com/BranchMetrics/branch-backend/pull/688/files#diff-54ee6cd60658735b0afa26e8af585600R733

PR #59, opened 2015-02-28
Check iOS version

qinweigong:
@scotthasbrouck @aaustin


aaustin:
:+1:

shortstuffsushi:
Verified on a 6.0 device. Works fine, but time from press of share button to time of display of share dialog is quite slow. Not blocking, but something to revisit.

:+1:

PR #60, opened 2015-03-02
add duration to iOS SDK for Chosen

aaustin:
@qinweigong urgent


Branch-SDK/Branch-SDK/BNCLinkData.h
```diff
@@ -20,6 +20,7 @@
 @property (nonatomic, strong) NSString *feature;
 @property (nonatomic, strong) NSString *stage;
 @property (nonatomic, strong) NSString *params;
+@property (nonatomic, assign) NSInteger duration;
```
qinweigong: To make it clear to partners, would it be better to declare it as NSUInteger?

aaustin:
Changed to NSU and realized I forgot update the link data @qinweigong

qinweigong:
:+1:  Could you also change it in BNCLinkData.h property declaration?

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -606,108 +606,125 @@ - (NSDictionary *)getLatestReferringParams {
...
+    [self generateShortUrl:tags andAlias:nil andType:BranchLinkTypeUnlimitedUse andMatchDuration:0 andChannel:channel andFeature:BRANCH_FEATURE_TAG_SHARE andStage:nil andParams:[BranchServerInterface encodePostToUniversalString:[self sanitizeQuotesFromInput:params]] andCallback:callback];
 }
 
 - (void)getContentUrlWithParams:(NSDictionary *)params andChannel:(NSString *)channel andCallback:(callbackWithUrl)callback {
-    [self generateShortUrl:nil andAlias:nil andType:BranchLinkTypeUnlimitedUse andChannel:channel andFeature:BRANCH_FEATURE_TAG_SHARE andStage:nil andParams:[BranchServerInterface encodePostToUniversalString:[self sanitizeQuotesFromInput:params]] andCallback:callback];
+    [self generateShortUrl:nil andAlias:nil andType:BranchLinkTypeUnlimitedUse andMatchDuration:0 andChannel:channel andFeature:BRANCH_FEATURE_TAG_SHARE andStage:nil andParams:[BranchServerInterface encodePostToUniversalString:[self sanitizeQuotesFromInput:params]] andCallback:callback];
```
shortstuffsushi: Is defaulting this to 0 the right thing to do?

shortstuffsushi: I guess in that case, it will just be ignored. Do we want to have a default match duration that is non-zero?

shortstuffsushi:
Is there any explanation as to what this "duration" parameter is, or what it's used for? Should that be added to the Readme or anything?

Branch-SDK/Branch-SDK/Branch.m
aaustin: It's a default set on the backend - there's no need to send/hardcode it client side

aaustin:
When someone clicks a link, there is an outstanding click object created that is eligible for match when an app opens up. This duration is the amount of time that the click object is eligible for match

shortstuffsushi:
So, this would be relevant when you're *creating* a link, allowing you to specify how long that link, when clicked, should keep an open match. Gotcha.

PR #62, opened 2015-03-08
[BNCServerInterface encodePostToUniversalString] cleanup

shortstuffsushi:
Large rework of `BNCServerInterface encodePostToUniversalString:needSource:]` method.

* Removing spaces at start/end of object, since they're not in the middle anyway. No need for extra bytes for those two spaces.
* Rather than executing `[params objectForKey:]` a dozen times inside the for loop, just storing off into a local var.
* Updating encoding of nested dictionaries to *not* include source attribute. Doesn't make sense to add this at any level other than root.
* Adding support for null (and adding note to add more support in the future).
* Adding `continue` in else case for types. If we haven't encoded it at this point, we end up trying to append nil which causes the crash mentioned in BranchMetrics/Branch-iOS-Invite-SDK#19.
* Cleaning up append calls, no need to have so many -- a couple of simple `appendFormat`s does the trick.

Also added some tests around this functionality to avoid regression. These have been lumped in with the stubless tests, which is probably not the right place, but the `Branch-SDK Tests` target is not included in the main project, doesn't have a scheme, and fails all the tests when I run it. This should be re-evaluated in the future.

CC @qinweigong and @aaustin

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCConfig.h
+++ b/Branch-SDK/Branch-SDK/BNCConfig.h
@@ -9,7 +9,7 @@
 #ifndef Branch_SDK_Config_h
 #define Branch_SDK_Config_h
 
-#define SDK_VERSION             @"0.4.6.2"
+#define SDK_VERSION             @"0.4.7"
 
 #define BNC_PROD_ENV
 //#define BNC_STAGE_ENV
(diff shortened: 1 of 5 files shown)


aaustin:
:+1: awesome and thanks for making the tests

PR #63, opened 2015-03-20
Branch Key

qinweigong:
@aaustin @shortstuffsushi 

- [ ] The branch_key in plist is still temp. Need to change it once backend is deployed, and fix tests.
- [ ] Update README with sample code for setting both live and test branch key and switch between them.


aaustin:
same comments as Android version - looks great otherwise

shortstuffsushi:
Hey guys,

The title and commit message don't really give me much to go on here; could you explain a bit more what the goal of changing this is? This will presumably be a breaking change for all of our users, but I'm not sure what the motivation is (and I wasn't able to find it in chat either). Also, should we deprecate the old methods and have some backwards compat at all?

shortstuffsushi:
Also, I think this will require a `readme` update.

qinweigong:
@shortstuffsushi Sorry for the lack of info here. This PR is part of the bigger project to 1) enable partners to switch between live/test apps, 2) build a public/private key system. Details is Asana.

Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
```diff
@@ -182,21 +182,37 @@ + (NSInteger)getRetryCount {
...
+}
+
++ (NSString *)getBranchKey:(BOOL)isLive {
+    NSString *key = nil;
+    
+    id ret = [[[NSBundle mainBundle] infoDictionary] objectForKey:KEY_BRANCH_KEY];
+    if (ret) {
+        if ([ret isKindOfClass:[NSString class]]) {
+            key = ret;
+        } else if ([ret isKindOfClass:[NSDictionary class]]) {
```
shortstuffsushi: Is there ever a case where this could be saved as a dictionary? I don't see any code here that inserts it in that manner.

shortstuffsushi: My mistake; you're inserting it into the PList that way, optionally.

shortstuffsushi:
Two things:

Like we discussed, I think supporting the old key with a deprecation warning should be done for one or two versions.

Tests are failing, those should be fixed before merging.

Otherwise, I think things look good. Assuming you support the old version (`bnc_app_key`), this won't be a breaking API change, so this could probably just be a minor version bump. The deprecated key should be removed in the next major version (`0.6.0` I believe).

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1208,6 +1240,11 @@ - (void)handleFailure:(unsigned int)index {
     
     if (req) {
         if ([req.tag isEqualToString:REQ_TAG_REGISTER_INSTALL] || [req.tag isEqualToString:REQ_TAG_REGISTER_OPEN]) {
+                self.initFailed = YES;
+                //                for (int i = 0; i < [self.requestQueue size]-1; i++) {
```
shortstuffsushi: Remove commented out add here.

shortstuffsushi:
Rather than passing the app key around to all the methods still, I was thinking you would just basically alias those calls (so, if they have an app_id, when we try to pass app_key we use that app_id in place). Will that not work with the backend though?

Also, this branch would really benefit from my "add app id on all request" branch, since that limits the number of location that you actually have that key. Instead of every single api request, it's centralized to the server interface.

derrickstaten:
@shortstuffsushi you guys should merge the branch in at some pt

shortstuffsushi:
I think that this all seems good now :+1:

PR #66, opened 2015-03-26
Add version to update check

Sarkar:
Logic:

If new install, stored version will be nil, so yes, update. 
If open -> check stored pref. If version same, no update. If version higher (can only go higher in iOS), yes update. 

@aaustin @qinweigong


shortstuffsushi:
Big :+1: for `pragma mark`s, but can you put a `-` between them? So `#pragma mark - foo`? It creates a nicer grouping the method drop down for the file.

Branch-SDK/Branch-SDK/BNCSystemObserver.m
```diff
@@ -121,13 +121,16 @@ + (NSString *)getDeviceName {
 }
 
 + (NSNumber *)getUpdateState {
-    NSString *bundleRoot = [[NSBundle mainBundle] bundlePath];    NSFileManager *manager = [NSFileManager defaultManager];
```
shortstuffsushi: If this returns an `NSNumber` if you need an update, and `nil` if you don't, wouldn't it just be better to make this a `BOOL` function, with a more appropriate name like "needsUpdate" or something?

Branch-SDK/Branch-SDK/BNCSystemObserver.m
```diff
@@ -73,7 +73,7 @@ + (NSString *)getURIScheme {
 }
 
 + (NSString *)getAppVersion {
-    return [[[NSBundle mainBundle] infoDictionary] objectForKey:@"CFBundleVersion"];
+    return [[[NSBundle mainBundle] infoDictionary] objectForKey:@"CFBundleShortVersionString"];
```
shortstuffsushi: Two comments here:

* We may want to keep using the CFBundleVersion, as that refers to the more-descript version identifier -- the build number, instead of the App Store version number. Not sure, open to debate on that
* We should use the constant, if possible. For CFBundleVersion, that's `kCFBundleVersionKey`

Branch-SDK/Branch-SDK/BNCSystemObserver.m
```diff
@@ -121,13 +121,16 @@ + (NSString *)getDeviceName {
...
-    if ((int)([[attrs fileCreationDate] timeIntervalSince1970]/(60*60*24)) == (int)([[attrs fileModificationDate] timeIntervalSince1970]/(60*60*24))) {
-        return nil;
-    } else {
+
+    NSString *storedAppVersion = [BNCPreferenceHelper getAppVersion];
+    NSString *currentAppVersion = [BNCSystemObserver getAppVersion];
+
+    if (storedAppVersion == nil ||
+        ([storedAppVersion compare:currentAppVersion options:NSNumericSearch] == NSOrderedAscending)) {
+        [BNCPreferenceHelper setAppVersion:currentAppVersion];
```
aaustin: So, maybe I'm just tired, but I think this doesn't actually work. Here are the 3 scenarios:
1. branch old update branch new
stored version something, current version different = update
2. fresh install with branch
stored version nil, current version something = update (but should be install)
3. no branch old update branch new 
stored version nil, current version something = update

shortstuffsushi: What's the difference between scenario two and three there?

Also, wouldn't it just be easiest to say `if (![storedAppVersion equal:currentAppVersion])`, since basically any time they aren't the same, we need an update?

aaustin: 2 is a real install from the app store, 3 is an update after integrating branch and the app was previously installed on the phone

shortstuffsushi: Hmm, is there any way for Branch to know that it's not an "install" at that point?

aaustin: Read the code that was deleted

shortstuffsushi: Ha, neat little trick there. Didn't think of that.

qinweigong:
BTW, what's the purpose of this PR? Is it meant to fix any bug?

aaustin:
Trusper saw 22k installs on the Branch dashboard yesterday but iTunesConnect reports 2100. My method of estimating an install vs an upgrade has a discrepancy and I thought we could try to do a better job.

qinweigong:
And the 22k installs were all from iOS? Hmm, the 10X difference seems too big to be all caused by mis-calculating "updates" as "installs".

Branch-SDK/Branch-SDK/BNCSystemObserver.m
Sarkar: Like these suggestions. Original usage for CFBundleShortVersionString is because an app update requires a higher version to be approved. Thoughts?

Branch-SDK/Branch-SDK/BNCSystemObserver.m
Sarkar: I like this, but I wanted to maintain whatever the original function signature was. Can our server accept BOOL values in the JSON? Havent tested it myself

shortstuffsushi: Was this previously being sent to the server? I thought you were adding that as a part of this PR.

Branch-SDK/Branch-SDK/BNCSystemObserver.m
shortstuffsushi: Huh, that is a fair point. I don't think anything is enforced with the Build Number except that it be different than the previous.

Branch-SDK/Branch-SDK/BNCSystemObserver.m
Sarkar: looks like I didnt fully understand the logic for update myself -- my bad. 

heres what i should modify this to: 

if we have a nil value for storedAppVersion at all times, this is a fresh install -> no update
if we have an existing value match -> no update
if we have differing values -> compare (if scenario one, this check is already no update), update

Branch-SDK/Branch-SDK/BranchServerInterface.m
```diff
@@ -70,6 +70,8 @@ - (void)registerOpen:(BOOL)debug {
     if (osVersion) [post setObject:osVersion forKey:@"os_version"];
     NSString *uriScheme = [BNCSystemObserver getURIScheme];
     if (uriScheme) [post setObject:uriScheme forKey:@"uri_scheme"];
+    NSNumber *updateState = [BNCSystemObserver getUpdateState];
+    if (updateState) [post setObject:updateState forKeyedSubscript:@"update"];
     [post setObject:[NSNumber numberWithBool:[BNCSystemObserver adTrackingSafe]] forKey:@"ad_tracking_enabled"];
```
Sarkar: also wanted to get your guys thought on this. this adds the check to register open - I don't think I should be doing that after your guys' feedback?

Branch-SDK/Branch-SDK/BNCSystemObserver.m
Sarkar: Yeah, it was. In that case I'll go back and change it to a BOOL val

Branch-SDK/Branch-SDK/BNCSystemObserver.m
aaustin: but what about the case when they first integrate Branch on a prexisting app with a large install base? We can't show all updates as installs

The problem with today is that the app developer had 2k actual itunesconnect downloads and 200k upgrades. We counted 20k incorrectly as installs instead of updates

Sarkar: There wouldn't be anything saved because branch would actively track after
that fact, so that wouldn't count as an update, no?

I'd love to whiteboard the possible scenarios here tomorrow btw, think I'm
missing a few back end pieces here

Branch-SDK/Branch-SDK/BNCSystemObserver.m
```diff
@@ -121,13 +121,25 @@ + (NSString *)getDeviceName {
...
+    NSString *currentAppVersion = [BNCSystemObserver getAppVersion];
+    NSString *bundleRoot = [[NSBundle mainBundle] bundlePath];
+    NSFileManager *manager = [NSFileManager defaultManager];
     NSDictionary* attrs = [manager attributesOfItemAtPath:bundleRoot error:nil];
-    if ((int)([[attrs fileCreationDate] timeIntervalSince1970]/(60*60*24)) == (int)([[attrs fileModificationDate] timeIntervalSince1970]/(60*60*24))) {
+
+    int fileCreationDate = (int)([[attrs fileCreationDate] timeIntervalSince1970]/(60*60*24));
+    int fileModificationDate = (int)([[attrs fileModificationDate] timeIntervalSince1970]/(60*60*24));
+
+    if (([storedAppVersion compare:currentAppVersion])
```
shortstuffsushi: Could you add a comment here describing what the logic here is, or use intention variables?

Branch-SDK/Branch-SDK/BNCSystemObserver.m
```diff
@@ -121,13 +121,25 @@ + (NSString *)getDeviceName {
...
+        || (storedAppVersion == nil)
+        || (fileCreationDate == fileModificationDate)) {
+        [BNCPreferenceHelper setAppVersion:currentAppVersion];
         return nil;
     } else {
```
shortstuffsushi: No need for a dangling else here.

PR #67, opened 2015-03-27
Callback immediately if initFailed

qinweigong:
@aaustin @shortstuffsushi

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCConfig.h
+++ b/Branch-SDK/Branch-SDK/BNCConfig.h
@@ -9,7 +9,7 @@
 #ifndef Branch_SDK_Config_h
 #define Branch_SDK_Config_h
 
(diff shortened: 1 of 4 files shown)


aaustin:
can we move the if callback with main dispatch to a separate function? Just to save on lines of code and characters

qinweigong:
Yes, wil do!

qinweigong:
Actually, maybe not! 
Because callbacks are of different types, i can pass them to the separate function as an id, but in the function i lose the context and don't know how to cast it back, and can't just use an id to call.

Or is there a way?

aaustin:
ahh yea that's a wrench in the spokes for sure - sorry for the wild goose chase

aaustin:
Oops forgot to :+1:

PR #68, opened 2015-03-29
new method for app creation date

derrickstaten:
@aaustin @shortstuffsushi


Branch-SDK/Branch-SDK/BNCSystemObserver.m
```diff
@@ -123,16 +123,21 @@ + (NSString *)getDeviceName {
 + (NSNumber *)getUpdateState {
     NSString *storedAppVersion = [BNCPreferenceHelper getAppVersion];
     NSString *currentAppVersion = [BNCSystemObserver getAppVersion];
-    NSString *bundleRoot = [[NSBundle mainBundle] bundlePath];
     NSFileManager *manager = [NSFileManager defaultManager];
-    NSDictionary* attrs = [manager attributesOfItemAtPath:bundleRoot error:nil];
+    
+    // for creation date
+    NSURL *documentsDirRoot = [[[NSFileManager defaultManager] URLsForDirectory:NSDocumentDirectory inDomains:NSUserDomainMask] lastObject];
```
shortstuffsushi: No need for another `defaultManager` call here, already as a local variable available above.

Branch-SDK/Branch-SDK/BNCSystemObserver.m
```diff
@@ -123,16 +123,21 @@ + (NSString *)getDeviceName {
...
+    NSDictionary *documentsDirAttributes = [manager attributesOfItemAtPath:documentsDirRoot.path error:nil];
+    int appCreationDay = (int)([[documentsDirAttributes fileCreationDate] timeIntervalSince1970]/(60*60*24));
```
shortstuffsushi: This would be better named, "number of days since creation."

Branch-SDK/Branch-SDK/BNCSystemObserver.m
```diff
@@ -123,16 +123,21 @@ + (NSString *)getDeviceName {
...
 
-    int fileCreationDate = (int)([[attrs fileCreationDate] timeIntervalSince1970]/(60*60*24));
-    int fileModificationDate = (int)([[attrs fileModificationDate] timeIntervalSince1970]/(60*60*24));
+    // for modification date
+    NSString *bundleRoot = [[NSBundle mainBundle] bundlePath];
+    NSDictionary *bundleAttributes = [manager attributesOfItemAtPath:bundleRoot error:nil];
+    int appModificationDay = (int)([[bundleAttributes fileModificationDate] timeIntervalSince1970]/(60*60*24));
```
shortstuffsushi: As with `appCreationDay`, better named "number of days since update."

Branch-SDK/Branch-SDK/BNCSystemObserver.m
```diff
@@ -123,16 +123,21 @@ + (NSString *)getDeviceName {
...
 
     if (!storedAppVersion) {
         [BNCPreferenceHelper setAppVersion:currentAppVersion];
-        if ([attrs fileCreationDate] && [attrs fileModificationDate] && (fileCreationDate != fileModificationDate)) {
+        if ([documentsDirAttributes fileCreationDate] && [bundleAttributes fileModificationDate] && (appCreationDay != appModificationDay)) {
```
shortstuffsushi: Is there any reason to check the existence of these vars here? You've already attempted to use them by the time you get to this comparison.

shortstuffsushi:
Am I understanding correctly that you're comparing the creation date of the documents directory with the updated date of the bundle directory?

If you're making the change from comparing creation date and update date of bundle, is there any reason you don't just compare creation date of both?

Branch-SDK/Branch-SDK/BNCSystemObserver.m
aaustin: You can check here and earlier or just here, but for cleanliness, I think it's better here alone.

aaustin:
On your last point, the thought that the creation date might not always equal the update date. It seems more like a bug that it does now than anything else

Branch-SDK/Branch-SDK/BNCSystemObserver.m
shortstuffsushi: Is it necessary at all, though? Is it possible for them to be nil? And if you didn't check, you'd end up w/ 0 values for both, in which case the result of this function would just be that it was not an update.

aaustin: We haven't observed either one in testing, but it's the only explanation for the second problem. It must be possible for modified date or creation date to be not set for the bundle on some phones. One or the other would return nil and would have a date of 0, resulting in the modified date to not equal the creation date.
-> This results in the installs being improperly classified as updates (which we've observed in the server logs)

shortstuffsushi: Hmm, I'd really like to see this behavior documented somewhere or something, especially if we're unable to reproduce the same thing. It would just be nice to have definitive things in our repo as much as possible. However, does this check actually fix the issue? In the case that either is nil, we'll return nil, which would mean it's an update instead of an install still.

PR #69, opened 2015-03-30
Sdk version on install

Sarkar:
Want to send SDK version on installOpen: and registerOpen: 

@aaustin @shortstuffsushi


aaustin:
:+1:

PR #70, opened 2015-03-31
Add retry number to requests

shortstuffsushi:
Reviewers @qinweigong @derrickstaten @Sarkar


Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -14,14 +14,35 @@
 @implementation BNCServerInterface
 
 // make a generalized get request
-- (void)getRequestAsync:(NSDictionary *)params url:(NSString *)url andTag:(NSString *)requestTag {
-    [self getRequestAsync:params url:url andTag:requestTag log:YES];
+- (void)getRequestAsync:(NSDictionary *)params url:(NSString *)url andTag:(NSString *)requestTag retryCount:(NSInteger)retryCount {
+    [self getRequestAsync:params url:url andTag:requestTag retryNumber:0 retryCount:retryCount log:YES];
 }
 
 // this is actually a synchronous call; it should NOT be called from the main queue
```
Sarkar: this comment applies to the function below it?

shortstuffsushi: This comment (and the other in the file) apply to all of the functions in this class. We do not currently make *any* async requests.

Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -29,30 +50,51 @@ - (BNCServerResponse *)getRequestSync:(NSDictionary *)params url:(NSString *)url
...
+
+    // Retry the request if appropriate
+    if (retryCount > 0 && isRetryableStatusCode) {
+        [NSThread sleepForTimeInterval:[BNCPreferenceHelper getRetryInterval]];
+
+        if (log) {
+            [BNCPreferenceHelper log:FILE_NAME line:LINE_NUM message:@"Replaying request with tag %@", requestTag];
+        }
+
+        [self postRequestAsync:post url:url andTag:requestTag andLinkData:linkData retryNumber:++retryNumber retryCount:--retryCount log:log];
```
qinweigong: Only concern with this recursive call is, theoretically it could lead to stackoverflow if someone sets retry count to a big number.

Branch-SDK/Branch-SDK/BNCServerInterface.m
Sarkar: slightly confusing since these methods have Async appended to them but oh well!

shortstuffsushi: Fortunately these are internal methods, not really public facing. I agree that they are confusingly named, though, and have plans to revisit them with the callback structure we had discussed.

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1650,7 +1639,9 @@ - (void)serverCallback:(BNCServerResponse *)response {
                 }
             }
             [self handleFailure:[self.requestQueue size]-1];
-        } else if (status != 200) {
+        }
+        // Anything else not between 200 and 400 indicates a server error or iOS error (like connectivity)
+        else if (status < 200 || status > 399) {
```
qinweigong: Do we want to cover status > 200 && status <= 399 ?

shortstuffsushi: Those are all considered "successful" and fall into the appropriate handler below this block.

Branch-SDK/Branch-SDK/BNCServerInterface.m
shortstuffsushi: Hmm, that's true, but I think we'd run into other problems either way if they set their retry count to 99999999999, even with the old method ;)

Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -14,14 +14,35 @@
...
+    
+    // Retry the request if appropriate
+    if (retryCount > 0 && isRetryableStatusCode) {
+        [NSThread sleepForTimeInterval:[BNCPreferenceHelper getRetryInterval]];
+
+        if (log) {
+            [BNCPreferenceHelper log:FILE_NAME line:LINE_NUM message:@"Replaying request with tag %@", requestTag];
+        }
+
+        [self getRequestAsync:params url:url andTag:requestTag retryNumber:++retryNumber retryCount:--retryCount log:log];
```
qinweigong: Can we use the fact that retryNumber = [BNCPreferenceHelper getRetryCount] - retryCount, and consolidate these two params into one? Just a bit confusing to have both retryNumber and retryCount in all these places :-)

shortstuffsushi: Actually, yeah. I was thinking about how to do that, and just now realized a better solution. Will refactor with that shortly.

qinweigong:
:+1:

Sarkar:
:+1:

PR #72, opened 2015-04-01
Centralize and extend encoding

shortstuffsushi:
Reviewers: @qinweigong @Sarkar @derrickstaten


Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -239,26 +176,21 @@ - (NSURLRequest *)preparePostRequest:(NSDictionary *)post url:(NSString *)url re
...
-            if (!convError) {
-                serverResponse.data = jsonData;
-            }
-        }
-    } else {
-        serverResponse = [[BNCServerResponse alloc] initWithTag:requestTag andStatusCode:[NSNumber numberWithInteger:error.code]];
+        serverResponse.data = [BNCEncodingUtils decodeJsonDataToDictionary:data];
+    }
+    else {
+        serverResponse = [[BNCServerResponse alloc] initWithTag:requestTag andStatusCode:@(error.code)];
```
Sarkar: minor, but since we create this twice, maybe have this outside the if / else block, and have a setter/getter for the status code accordingly?

shortstuffsushi: Yeah, missed that. Updated.

Sarkar:
this refactor is freaking awesome, man. :+1: super excited for this :bowtie:

qinweigong:
Looking good! :+1:

derrickstaten:
massive refactor! this is awesome!

two questions.

first, are we definitely okay with iso8601StringFromDate? @qinweigong @Sarkar 

second, this is probably not important, but there's some shared code between encodeArray and encodeDictionary. it's small though

Sarkar:
@derrickstaten re: iso8601StringFromDate - yeah, we had chatted about it in slack but afaik server accepts iso8601. others should chime in though

shortstuffsushi:
@derrickstaten and @Sarkar, the Slack chat ended on a somewhat inconclusive note, and I had already implemented this. We could change it in the future, but it seems that some of our API calls (`/referralcode`, `/applycode`, and `/credithistory` specifically) are already returning ISO strings for dates.

derrickstaten:
@shortstuffsushi ISO strings of what format? I'd almost rather force people to use an NSDateFormatter / have an assert. But in production that's probably not a good idea

shortstuffsushi:
ISO8601 are what come back from Branch. I think that's probably the defacto standard, and unless our customers say otherwise, I think it's a pretty safe format to go with (and, like you said with asserts, production safe :wink:). If this format is *not* what people are expecting, they can always run it through their own date formatter as well, as you mentioned. I will add a note to the `readme` about that quick, though.

derrickstaten:
sounds good 👍

PR #73, opened 2015-04-01
Uri scheme updates

### LINKED ISSUE #51: SDK needs to support apps with multiple URL schemes
Presently the SDK incorrectly assumes that the Branch URL scheme is the first URL scheme in an app's Info.plist. Docs need to be updated to indicate this for apps adding Branch support that may likely already have one or more URL schemes already defined.

shortstuffsushi:
Reviewers: @qinweigong @Sarkar @derrickstaten.

This will fix #51.


shortstuffsushi:
Open question: is it valuable / correct to be saving some of the "configuration" preference items into User Defaults? These things are going to be embedded into apps, typically in the AppDelegate, and will be set on every single launch. Additionally, if they were to push a future build that no longer called the `set` method, the old value would *still* come through because it's stored in defaults.

qinweigong:
That's exactly my question for this PR. Imagine a partner uses this new setUriScheme for a test uri scheme during development, while keeping his prod uri scheme still in the plist.
When he's done with dev, he removes this setUriScheme call. But at that time we will still read uri scheme from user defaults instead of from plist.
This won't be a problem for the parter's users, but the partner is likely to get confused and send us unnecessary support requests complaining about this.
After all, is it necessary to give partner this ability to set an alternative uri scheme?

shortstuffsushi:
See the referenced issue (#51). I believe that we *do* need to support this, as an app can have multiple Uri schemes, and we assume it to be the first. That's correct in most cases, but not always, which is why we should allow them to specify.

Your scenario is exactly my worry as well, so I think we will need to revisit this before merging. The same is true for the other config items as well (`retryCount`, `retryInterval`, and `networkTimeout`).

qinweigong:
Hah, now I remember how i "solved" issue #51 : I added following to the README :-)
"Make sure to keep this URI scheme as the first item (item 0) under URL schemes"

Yah, we need a better solution for this.

shortstuffsushi:
That explains why I couldn't find it -- I was looking through code for a related change, not the readme :smile:

qinweigong:
Actually I was thinking about the same thing: just saving those configuration stuff in memory rather than in UserDefaults! So I am cool on doing so for retryCount, retryInterval and timeout.

However, I have another thought about Uri Scheme. I don't think we have been doing it correctly.
The thing is we are not utilizing the identifier (i.e. CFBundleURLName) of the Uri scheme setup in plist!

Take the following for example, where I added CFBundleURLName ("io.branch.sdk"), and deliberately moved it to the second position.
 
<key>CFBundleURLTypes</key>
	<array>
		<dict>
			<key>CFBundleTypeRole</key>
			<string>Editor</string>
			<key>CFBundleURLName</key>
			<string>com.facebook</string>
			<key>CFBundleURLSchemes</key>
			<array>
				<string>1234</string>
			</array>
		</dict>
        <dict>
            <key>CFBundleTypeRole</key>
            <string>Editor</string>
            <key>CFBundleURLName</key>
            <string>io.branch.sdk</string>
            <key>CFBundleURLSchemes</key>
            <array>
                <string>branchtest</string>
            </array>
        </dict>
</array>

Then we just need to change our getUriScheme method to:
+ (NSString *)getURIScheme {
    NSArray *urlTypes = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleURLTypes"];
    for (NSDictionary *urlType in urlTypes) {
        if ([urlType[@"CFBundleURLName"] isEqualToString:@"io.branch.sdk"]) {
            return [urlType objectForKey:@"CFBundleURLSchemes"][0];
        }
    }
    return nil;
}

I believe this is how other SDKs (e.g. facebook) are doing. The only thing is we need to let existing partners to migrate to this by adding the scheme identifier.

qinweigong:
Shoot, XML won't display correctly in my previous comment!

qinweigong:
:+1: only concerned about backward compatibility

derrickstaten:
Did we actually add the following here? 

```
+@property (assign, nonatomic) NSInteger retryCount;
+@property (assign, nonatomic) NSInteger retryInterval;
+@property (assign, nonatomic) NSInteger timeout;
```

It seems like there's non-URI scheme code being marked as "added" by git

shortstuffsushi:
Yes, those were added as a part of this pull request to respond to the comment that configuration items should not be persisted into NSUserDefaults, as they can then potentially be incorrectly carried along between versions.

aaustin:
A couple high level thoughts:
1. For most folks, going in and editing the Plist is hectic. Before this proposal, we're asking them to just add a URI scheme to the list, right along side where they've likely done Facebook or others. This change is asking them to create a new hierarchy in the Plist file, label it for Branch and then put the URI scheme in.

This is a behavior change from the norm and complexity to offload for the developer. That's a big ask. From what I can tell, the only benefit to doing this would be so that we can more accurately determine what the URI scheme was. Do we have data to indicate that this is an issue? For what gain in accuracy of URI scheme reading can we weigh this complexity.

2. setUriScheme doesn't actually set the URI scheme - it only tells Branch what it is. This is extremely confusing. So now they need to set the URI scheme in both the PLIST and in the Branch SDK? But if they only do it in the Branch SDK, it won't work. Going back to my previous point. What do we gain from this? How much will our accuracy of reading the URI scheme be improved.

Fundamentally, that URI scheme method to retrieve it was added purely as a convenience, so when they loaded up the dashboard to go through the onboarding flow, they would find their URI scheme pre-populated. If it wasn't there, or was wrong, it would be changed. While setting up Branch, the primary method of setting the URI scheme is on the dashboard, where you configure all the links. We don't want to put this in the SDK

PR #74, opened 2015-04-02
App id on all requests

shortstuffsushi:
@qinweigong @Sarkar @derrickstaten


qinweigong:
Looks good, after fixing the tests :+1:

PR #77, opened 2015-04-07
Activity item bypass first click

derrickstaten:
@shortstuffsushi @aaustin 

I changed ignoreFirstClick to a string and sent up the user agent for a UIWebView. You should test this in conjunction with the backend PR that I'm about to make.


Branch-SDK/Branch-SDK/BranchActivityItemProvider.m
```diff
@@ -31,15 +31,27 @@ - (id)initWithParams:(NSDictionary *)params
...
-        return self.branchURL;
+        
+        // Because Facebook immediately scrapes URLs, we add an additional parameter to the existing list, telling the backend to ignore the first click
+        BOOL ignoreFirstClick = [channel isEqualToString:@"facebook"];
+        if (ignoreFirstClick && self.userAgentString) {
+            self.branchURL = [[Branch getInstance] getShortURLWithParams:self.params andTags:self.tags andChannel:channel andFeature:self.feature andStage:self.stage andAlias:self.alias ignoreFirstClick:self.userAgentString];
+        } else {
+            self.branchURL = [[Branch getInstance] getShortURLWithParams:self.params andTags:self.tags andChannel:channel andFeature:self.feature andStage:self.stage andAlias:self.alias ignoreFirstClick:nil];
+        }
+        NSLog(@"url: %@", self.branchURL);
```
aaustin: remove this

Branch-SDK/Branch-SDK/BNCLinkData.h
```diff
@@ -21,6 +21,7 @@
 @property (nonatomic, strong) NSString *stage;
 @property (nonatomic, strong) NSString *params;
 @property (nonatomic, assign) NSUInteger duration;
+@property (nonatomic, assign) NSString *ignoreFirstClick;
```
shortstuffsushi: Not sure why this was changed to a string; if you're passing user agent here instead, then you should probably rename this to an entirely new variable.

Branch-SDK/Branch-SDK/BNCLinkData.m
```diff
@@ -30,6 +30,7 @@ - (id)copyWithZone:(NSZone *)zone {
     copy.params = [_params copyWithZone:zone];
     copy.type = _type;
     copy.duration = _duration;
+    copy.ignoreFirstClick = _ignoreFirstClick;
```
aaustin: need copy with zone like the strings above

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1100,6 +1106,7 @@ - (BNCLinkData *)prepareLinkDataFor:(NSArray *)tags andAlias:(NSString *)alias a
     [post setupStage:stage];
     [post setupAlias:alias];
     [post setupMatchDuration:duration];
+    if (ignoreFirstClick) [post setupIgnoreFirstClick:ignoreFirstClick];
```
shortstuffsushi: Note that the other items here have the `if` within the setup method itself. This should be move there for consistency. Also, I really prefer bracketed `if`s, even for one liners like this.

Branch-SDK/Branch-SDK/BranchActivityItemProvider.m
```diff
@@ -31,15 +31,27 @@ - (id)initWithParams:(NSDictionary *)params
         self.branchURL = url;
     }
     
+    dispatch_async(dispatch_get_main_queue(), ^{
```
shortstuffsushi: You're dispatching *to* the main thread (read: locking the main thread) here. Any reason you're trying to push this into a block instead of just evaluating directly?

aaustin:
make sure to sync too for the latest versiom

Branch-SDK/Branch-SDK/BNCLinkData.h
Sarkar: i second this. reading through it wasnt very clear what ignoreFirstClick contained since it was treated as bool inside an if statement. could you please rename this?

PR #79, opened 2015-04-10
Branch key

qinweigong:
@shortstuffsushi


Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -982,6 +965,22 @@ - (void)applyReferralCode:(NSString *)code andCallback:(callbackWithParams)callb
 
 // PRIVATE CALLS
 
++ (Branch *)getBranchInstance:(BOOL)isLive {
+    if (!currInstance) {
+        NSString *branchKey = [BNCPreferenceHelper getBranchKey:isLive];
+        if (!branchKey || [branchKey isEqualToString:NO_STRING_VALUE]) {
+            NSLog(@"Branch Warning: Please enter your branch_key in the plist!");
+        }
+        NSString *appId = [BNCPreferenceHelper getAppKey];
```
shortstuffsushi: This won't quite work -- if you have one or the other, you shouldn't see a warning. This currently will show you a warning for each one you don't have (so, if you specified none, you'd see both warnings :smile:) We also probably don't want to tell them to set the old app key.

README.md
```diff
@@ -83,13 +83,17 @@ Alternatively, you can add the URI scheme in your project's Info page.
 
 ### Add your Branch Key to your project
 
-After you register your app, your app key can be retrieved on the [Settings](https://dashboard.branch.io/#/settings) page of the dashboard. Now you need to add it to YourProject-Info.plist (Info.plist for Swift).
+After you register your app, your Branch Key can be retrieved on the [Settings](https://dashboard.branch.io/#/settings) page of the dashboard. Now you need to add it to YourProject-Info.plist (Info.plist for Swift).
 
 1. In plist file, mouse hover "Information Property List" which is the root item under the Key column.
 1. After about half a second, you will see a "+" sign appear. Click it.
-1. In the newly added row, fill in "bnc_app_key" for its key, leave type as String, and enter your Branch Key obtained in above steps in its value column.
+1. In the newly added row, fill in "branch_key" for its key, leave type as String*, and enter your app's Branch Key obtained in above steps in the value column.
```
shortstuffsushi: I don't think we need the * on String.

README.md
```diff
@@ -83,13 +83,17 @@ Alternatively, you can add the URI scheme in your project's Info page.
...
 1. Save the plist file.
 
+* If you want to add key for both your live and test apps at the same time, you need change this type to Dictionary, and add two entries inside:
```
shortstuffsushi: A screenshot would be helpful for explaining this.

README.md
```diff
@@ -118,21 +122,36 @@ This deep link routing callback is called 100% of the time on init, with your li
...
+```
+
+Or
+
+```objc
+Branch *branch = [Branch getInstance:@"your test branch key"];  // replace with your actual branch key
+```
+
+Either way, we recommend you put a #warning directive to remind you to change back to live app during deployment later.
+Also, note the Branch object is singleton, so you can and should still use `[Branch getInstance]` in all the other places (see examples below).
```
shortstuffsushi: Hmm, this seems confusing to me. Calling `getTestInstance` once will cause all future calls to `getInstance` to return the test instance... feels a little strange.

README.md
```diff
@@ -171,6 +190,21 @@ func application(application: UIApplication, openURL url: NSURL, sourceApplicati
...
+let branch: Branch = Branch.getTestInstance()
+```
+
+Or
+
+```swift
+let branch: Branch = Branch.getInstance("your test branch key");  // replace with your actual branch key
+```
+
+Either way, we recommend you put a //TODO: or //FIXME landmark to remind you to change back to live app during deployment later.
```
shortstuffsushi: You recommended using a #warning above

Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -121,12 +121,10 @@ - (NSURLRequest *)prepareGetRequest:(NSDictionary *)params url:(NSString *)url r
     fullParamDict[@"retryNumber"] = @(retryNumber);
     
     NSString *appId = [BNCPreferenceHelper getAppKey];
-    // TODO re-add this
-    //    NSString *branchKey = [BNCPreferenceHelper getBranchKey];
-    //    if (![branchKey isEqualToString:NO_STRING_VALUE]) {
-    //        fullParamDict[KEY_BRANCH_KEY] = branchKey;
-    //    }
-    //    else if (![appId isEqualToString:NO_STRING_VALUE]) {
```
shortstuffsushi: Note that this was an `else if` here -- Dmitri had requested that we only send one value or the other.

README.md
qinweigong: seems #warning is only for objc, doesn't work in swift.

README.md
qinweigong: Yes, will add it today.

README.md
shortstuffsushi: Ah, I missed that these were different langs. It would be cool to find some consistency between the two, though. I'm still not real familiar with Swift, but it's unfortunate that those differ.

README.md
shortstuffsushi: And just as a reminder, Dmitri had requested that we start pushing these into GitHub instead of S3.

shortstuffsushi:
I think this looks alright now :+1:

PR #80, opened 2015-04-10
Callback all the things v1

shortstuffsushi:
@aaustin @qinweigong @derrickstaten @Sarkar 

This is largely a rewrite of most of the internals of Branch.m. It is step one of what I'm hoping to accomplish in separating a lot of the queuing logic from Branch itself.

Some other things coming out of this:
  * Errors are more specific to their failure, not just a generic message.
  * Queue is not persisted on *every* request, but instead on a time basis.
  * Fix infinite loop of logging/encoding in debug mode.
  * Requests are actually asynchronous, no need to have `dispatch_async` everywhere anymore.
  * Requests now have a specific callback rather than the big switch statement.
  * When not initialized and a request is made, an initialization requests will be pushed into the queue.
  * Initialization is managed by a single boolean instead of five.
  * Logic around failure has changed some:
    * All requests call back to their callbacks, every time.
    * All requests are dequeued
    * `setIdentity` and `userCompletedAction` requests are re-enqueued, minus their callback.

This will need manual review, you're not just going to be able to read the diff, since it's such a huge rework.


derrickstaten:
yeah I'm probably not going to complete going through this today. also, does this have to be all or nothing? i'm worried if we have to revert this is going to become a giant mess. but i guess there's no going back once we push this.

shortstuffsushi:
@derrickstaten I have mixed feeling about the size as well. Unfortunately, a couple of the underpinning changes here are the kind that touch the whole sdk. Specifically, the change to callbacks for the BNCServerInterface and actual async requests primarily.

There are a couple of things that could be pulled out of this, but I think they're much lower impact. Changing to more accurate error messages and the queue persistence change don't necessarily need to be part of this, but seem unlikely to break anything, and are easy enough to revert if they do cause problems.

As mentioned above, rather than trying to diff this, I think extensive testing of the various cases locally will be our best bet. I've tried to do so as best as possible, but would appreciate an additional review from everyone on this PR.

aaustin:
I'll need some time to review this - will try tonight or tomorrow

Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -11,109 +11,130 @@
...
-        }
-
-        [self getRequestAsync:params url:url andTag:requestTag retryNumber:++retryNumber log:log];
-    }
-    // Otherwise, let the delegate handle it
-    else if (self.delegate) {
-        [self.delegate serverCallback:serverResponse];
-    }
+    [self genericHTTPRequest:request withTag:requestTag andLinkData:nil retryNumber:retryNumber log:log callback:callback retryHandler:^NSURLRequest *(NSInteger lastRetryNumber) {
+        return [self prepareGetRequest:params url:url retryNumber:++lastRetryNumber log:log];
```
qinweigong: The only difference between the prepareGetRequest call here and above is the incremented retryNumber. Feeling kinda wasted to rebuild the whole NSURLRequest again just for this diff.

Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -11,109 +11,130 @@
...
+    NSURLRequest *request = [self preparePostRequest:post url:url retryNumber:retryNumber log:log];
 
-        [self postRequestAsync:post url:url andTag:requestTag andLinkData:linkData retryNumber:++retryNumber log:log];
-    }
-    // Otherwise, let the delegate handle it
-    else if (self.delegate) {
-        [self.delegate serverCallback:serverResponse];
-    }
+    [self genericHTTPRequest:request withTag:requestTag andLinkData:linkData retryNumber:retryNumber log:log callback:callback retryHandler:^NSURLRequest *(NSInteger lastRetryNumber) {
+        return [self preparePostRequest:post url:url retryNumber:++lastRetryNumber log:log];
```
qinweigong: same here.

Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -11,109 +11,130 @@
...
-    NSError * error = nil;
-    NSData * POSTReply = [NSURLConnection sendSynchronousRequest:request returningResponse:&response error:&error];
+- (void)genericHTTPRequest:(NSURLRequest *)request withTag:(NSString *)requestTag andLinkData:(BNCLinkData *)linkData retryNumber:(NSInteger)retryNumber log:(BOOL)log callback:(BNCServerCallback)callback retryHandler:(NSURLRequest *(^)(NSInteger))retryHandler {
+    [NSURLConnection sendAsynchronousRequest:request queue:[NSOperationQueue mainQueue] completionHandler:^(NSURLResponse *response, NSData *responseData, NSError *error) {
+        BNCServerResponse *serverResponse = [self processServerResponse:response data:responseData error:error tag:requestTag andLinkData:linkData];
+        NSInteger status = [serverResponse.statusCode integerValue];
+        BOOL isRetryableStatusCode = status >= 500;
+        
+        // Retry the request if appropriate
+        if (retryNumber < [BNCPreferenceHelper getRetryCount] && isRetryableStatusCode) {
```
qinweigong: isRetryableStatusCode is not used again down there, maybe just use && status >= 500 here for better readability?

Branch-SDK/Branch-SDK/BNCServerRequestQueue.m
```diff
@@ -157,9 +157,21 @@ - (BOOL)containsClose {
...
 
-- (void)persist {
-    NSUserDefaults *defaults = [NSUserDefaults standardUserDefaults];
+- (void)persistEventually {
+    if (!self.writeTimer.valid) {
+        self.writeTimer = [NSTimer scheduledTimerWithTimeInterval:BATCH_WRITE_TIMEOUT target:self selector:@selector(persistToDisk) userInfo:nil repeats:NO];
+    }
+}
+
+- (void)persistImmediately {
```
qinweigong: So currently this one is not called/used anywhere?

qinweigong:
Why I can't see the diffs in Branch.m? It just displays "846 additions, 1,060 deletions not shown".

Branch-SDK/Branch-SDK/BNCServerInterface.m
shortstuffsushi: Yeah, I know. I'm not a fan of it either, but since the URL Request includes the actual request body / query string, it has to be separate, unless you've got any better thoughts for that.

Branch-SDK/Branch-SDK/BNCServerRequestQueue.m
shortstuffsushi: This is a public method, and is called in a couple places within the `Branch.m` file, like when the app goes to background.

shortstuffsushi:
@qinweigong, the diff is probably too large for Github to render. However, I mentioned above that the diff will be somewhat meaningless anyway, as it is largely a rewrite of that entire file -- you'd be better off actually inspecting it locally.

Branch-SDK/Branch-SDK/BNCServerInterface.m
shortstuffsushi: This is called an intention variable (struggling to find a good link), and is used for the sake of readability. If I was to wander through this code base and find a `&& status > 500`, it wouldn't really be clear to me why a status greater than 500 should cause this check to pass. In this case, we're basically defining statuses greater than 500 as retryable statuses.

Sarkar: Agree with graham here. For the sake of readability, this makes much more sense as to the reader it's clearer than `status > 500`

aaustin:
I hate to dump this on you but can you add tests for the 4 scenarios i mentioned in slack yesterday:
1. Session init with internet, which is dropped thereafter for further requests. Then close with no internet, reopen with internet and should recover. Also make sure all callbacks are executed in every scenario with proper errors.
2. Session init without internet, make further requests, close with no internet, reopen with internet and make sure recovers. Also make sure all callbacks are executed in every scenario with proper errors.
3. Session init with internet, which is dropped thereafter for further requests, then reenable internet and let recover
4.  Session init without internet, make further requests, reenable internet and make sure recovers.

Sarkar:
Just FYI, i've tested with a tool called network link conditioner, found here http://nshipster.com/network-link-conditioner/

It's a bit tricky to re-create, but can be done

aaustin:
I'm worried about re-queueing the userCompletedActions that failed. Is that the method of not disposing of them? With events, order matters a lot because it's how we calculate the conversion funnels on the backend. If you pull it out and requeue, it's possible that signup could get called after purchase in some strange connectivity scenario. Can you just not remove it from the queue?

aaustin:
looks good otherwise! Can you add those tests I mentioned above? I'd feel a lot better if we did before pushing this out. Sorry to poo poo it

shortstuffsushi:
@aaustin the requests are currently being re-enqueued so that other requests get a chance to play. The alternative is to stop processing once we hit a failure due to network or Branch server errors. We can't continue processing items in the queue if we don't remove a failed request; it'll just keep wanting to use it because it's at the front of the queue. Additionally, per what you're saying, it wouldn't make sense to continue, as there may be other order-sensitive requests later in the queue. So, should I take that to mean that we should stop procecessing the queue once this sort of request failure occurs?

shortstuffsushi:
Regarding testing, I will try to find a way to add test cases around that, but what you've described above might be somewhat tricky to simulate. I've documented the scenarios in a Google Doc, and have run the test app through those scenarios to verify the behavior is as expected, though.

derrickstaten:
If we get an internet failure or a branch server error, why would we pull a request off the queue and move on to the next one? Shouldn't we just stop the queue and wait for the internet/server to be back online per @aaustin's suggestion

This seems superior to changing the order of requests. The only scenario in which this is bad is with a malformed request that's always going to get a 500. It would be pretty awful to have that hold up the queue forever. 

This is trickier than I realized

shortstuffsushi:
500s indicate server error -- think you mean 400s. I am actually dequeueing and not re-enqueueing 400s specifically -- if the request is malformed, it should not be reattempted later. I can update the code to not move the request and just stop processing, but should that happen for non-re-enqueueable requests as well? So:

If I call /referrals (loadActionCounts, not re-enqueueable), and it fails w/ 500 (for example), the callback will be called indicating failure, but should further processing be stopped at this point as well?

If I call /event (userCompletedAction, re-enqueueable), and it fails w/ 500 (for example), the callback will be called indicating failure, but should further processing be stopped at this point as well?

shortstuffsushi:
@derrickstaten and @aaustin I also want to point out that there is not currently any logic to "wait until the connection comes back online," the SDK simply waits until the next request to continue. In most cases, this would probably happen if the user backgrounded/foregrounded the app.

shortstuffsushi:
@aaustin the scenario tests have been added, as requested. They are failing on CircleCI at the moment, as they tend to do from time to time, but they work just fine, and cover the scenarios described via some reflection in the SDK. All of these tests can be cleaned up significantly in future version by separating out the actual server interface layer, which will also have the added benefit of avoiding these problems with CircleCI.

aaustin:
Alex-test 2: Fail.

init session in airplane mode
call userCompletedAction
watch logs

shortstuffsushi:
Hey @aaustin. I'm not at my computer right now to check, but I'm guessing you're referring to to it making endless requests to userCompletedAction when the request fails. This was something I had asked for input for, and didn't get feedback on. What do you think should happen in the scenario of a failed complete action?

Should it just stop trying to process queue items?
Should it re-enqueue and try the next?
If it does try the next, but there are two userCompletedActions in the queue, how should it prevent them from looping?

I'm aware of the scenario, but was looking for feedback. Thoughts?

aaustin:
It should leave it in the queue and stop processing after it fails - similar to the existing SDK. It can retry again once another request is made. We decided in that meeting a while back that only userCompletedAction and setIdentity would be left in the queue.

Can you let me know about any other land mines that aren't finished so we can find get them fixed before I test it?

shortstuffsushi:
@aaustin, if any other request fails, should the queue stop processing as well? This is not how it behaves (on this branch) currently; it will continue after a failure. Should this be different for 400s, though, which indicate a user error and not request failure?

Also, I believe this is the only remaining thing that I wasn't sure about. Anything else is probably just a bug.

aaustin:
Ah - you mean continue to process any new requests that were added to the queue after the failure callback? That's what causes us to try to execute the queue again in the live SDK and should do the same in the update. We just don't want to execute any more retries after the count has been hit for a given request in that cycle..

400s we should stop and dequeue (and call the callback if applicable) as that indicates there's an SDK bug trying to submit incorrect data to the API.

aaustin:
Found another issue.

1. drop connection
2. open app
3. call a bunch of branch functions including /v1/event
4. got to home screen
5. see call to /v1/event where /v1/close should be. Expect to see nothing since session never opened

aaustin:
Nice one :+1: 

Let's hope we don't suffer too much from some extreme edge cases

derrickstaten:
i'm terrified. such a massive change. i've become a lover of incremental change

Sarkar:
god speed - let's do this!

On Fri, May 8, 2015 at 3:41 PM, derrickstaten <notifications@github.com>
wrote:

> i'm terrified. such a massive change. i've become a lover of incremental
> change
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/BranchMetrics/Branch-iOS-SDK/pull/80#issuecomment-100387280>
> .
>



-- 

*Sahil Verma*

*Branch Metrics *

w branch.io | c 510-386-1750 | s sahilroyverma

PR #82, opened 2015-04-16
Calling setValue:forKey: on an immutable dictionary

shortstuffsushi:
The issue is not that the session id is nil, but that we're calling setValue:forKey: on an immutable dictionary. This is addressed on a separate branch, but I'm pulling the fix here as well.

@qinweigong @aaustin


aaustin:
:+1:

PR #84, opened 2015-04-20
Separating set/get of update state.

shortstuffsushi:

Previously, this was being updated when a get was called, which feels somewhat unnatural.
Additionally, in the case of an install or open "failing," the update state shouldn't be updated.
UpdateState should only be set when the open/install call completes successfully.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCSystemObserver.h
+++ b/Branch-SDK/Branch-SDK/BNCSystemObserver.h
@@ -20,7 +20,8 @@
 + (NSString *)getOSVersion;
 + (NSNumber *)getScreenWidth;
 + (NSNumber *)getScreenHeight;
-+ (NSNumber *)getUpdateState:(BOOL)updatePrefs;
++ (NSNumber *)getUpdateState;
++ (void)setUpdateState;
 + (NSString *)getDeviceName;
 + (NSDictionary *)getListOfApps;
 + (BOOL)isSimulator;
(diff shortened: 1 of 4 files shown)


derrickstaten:
So far so good!

One comment: I think looking at an entire day is super unnecessary. I changed the code on my end so we're dealing with increments of ten seconds instead of a day. 

    int appCreationDay = (int)([[documentsDirAttributes fileCreationDate] timeIntervalSince1970]/(10));
    int appModificationDay = (int)([[bundleAttributes fileModificationDate] timeIntervalSince1970]/(10));

These should probably be renamed if we go this route. Thoughts?

derrickstaten:
:+1:

shortstuffsushi:
Regarding the whole day comparison, three things:
  * If we were going to make that change, I would prefer it be a separate PR, as it isn't really related to the changes being made here.

  * You're actually making additional operations here that the *device* doesn't care about. While as a human, comparing small numbers is simpler, the device doesn't care -- `int` comparisons are the same between any size `int`s, so the actual `/ 10` operation is an extra step that isn't necessary.

  * Dividing by 10 would actually *break* the functionality here. Say the first value is 10000001 and the second is 10000000. Those values divided by 10 (as `int`s) are equal. Instead, if you were to do modulus comparison, you'd only be comparing the last 10 seconds, but then 10000001 and 20000001 would be equal.

So, I propose we not make that change :wink:

shortstuffsushi:
Actually, that does bring up an interesting point, though. Why aren't we just comparing the date equality here? There's no need to change from `NSDate` to `int`.

derrickstaten:
so for a future PR, we should move to a new system.

PR #85, opened 2015-04-21
Adding long User URL to getShortUrl failure logic.

shortstuffsushi:

Returning User URL by itself isn't correct, but you can add all the params and make it a long url.
Note that we don't have the items available in the handleFailure: method, so we don't even try (also note this won't be the case in the callbacks branch).
Updating generateLongUrl to share the same method, since they're accomplishing the same thing.

### DIFF
--- a/Branch-SDK/Branch-SDK/Branch.m
+++ b/Branch-SDK/Branch-SDK/Branch.m
@@ -1089,8 +1089,8 @@ static Branch *currInstance;
         [longUrl appendFormat:@"stage=%@&", stage];
     }
     
-    [longUrl appendFormat:@"type=%lld&", (long long)type];
-    [longUrl appendFormat:@"matchDuration=%lld&", (long long)duration];
+    [longUrl appendFormat:@"type=%ld&", (long)type];
+    [longUrl appendFormat:@"matchDuration=%ld&", (long)duration];
     
     NSData *jsonData = [BNCEncodingUtils encodeDictionaryToJsonData:params];
     NSString *base64EncodedParams = [BNCEncodingUtils base64EncodeData:jsonData];
--- a/Branch-SDK/Branch-SDK/Branch.m
+++ b/Branch-SDK/Branch-SDK/Branch.m
@@ -991,9 +991,11 @@ static Branch *currInstance;
             } else {
                 if (callback) {
                     NSString *failedUrl = nil;
-                    if (![[BNCPreferenceHelper getUserURL] isEqualToString:NO_STRING_VALUE]) {
-                        failedUrl = [BNCPreferenceHelper getUserURL];
+                    NSString *userUrl = [BNCPreferenceHelper getUserURL];
+                    if (![userUrl isEqualToString:NO_STRING_VALUE]) {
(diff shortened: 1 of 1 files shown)


Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1048,41 +1050,53 @@ - (NSString *)generateShortUrl:(NSArray *)tags andAlias:(NSString *)alias andTyp
...
+    if ([feature length]) {
+        [longUrl appendFormat:@"feature=%@&", feature];
+    }
+    
+    if ([stage length]) {
+        [longUrl appendFormat:@"stage=%@&", stage];
     }
     
-    return longURL;
+    [longUrl appendFormat:@"type=%lld&", (long long)type];
```
qinweigong: we need long long type here?

shortstuffsushi: `type` is an enum, and you can't just append that to a string. The nearest is to cast it to an NSInteger (long long) and append that.

qinweigong: Thought NSInterger is (long)?

qinweigong:
Was not involved in any previous discussion, so not sure if returning the long url is correct business logic wise. Other than that, looks good to me :+1:

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1048,41 +1050,53 @@ - (NSString *)generateShortUrl:(NSArray *)tags andAlias:(NSString *)alias andTyp
...
+    if ([appIdentifier isEqualToString:NO_STRING_VALUE]) {
+        NSLog(@"No Branch Key specified, cannot create a long url");
+        return nil;
+    }
     
-    NSData *jsonData = [BNCEncodingUtils encodeDictionaryToJsonData:params];
-    NSString *base64EncodedParams = [BNCEncodingUtils base64EncodeData:jsonData];
+    NSString *baseLongUrl = [NSString stringWithFormat:@"%@/a/%@", BNC_LINK_URL, appIdentifier];
+
+    return [self longUrlWithBaseUrl:baseLongUrl params:params tags:tags feature:feature channel:nil stage:stage alias:alias duration:0 type:BranchLinkTypeUnlimitedUse];
```
Sarkar: just curious -- `BranchLinkTypeUnlimitedUse` is the default parameter here right?

shortstuffsushi: I believe so, that's the same value used in the GenerateShortUrl methods above.

Branch-SDK/Branch-SDK/Branch.m
shortstuffsushi: Hmm, it looks like that's right. I'm not sure why I thought it was this way. If I can't find anything supporting that, I'll switch to just `%ld`.

shortstuffsushi: Either way, adds another quick task to the back log -- make enums more defined. It looks like Apple introduced a new syntax for declaring them to help make the type def more explicit (see [NSHipster](http://nshipster.com/ns_enum-ns_options/))

Sarkar:
looks good to me! :+1:

PR #86, opened 2015-04-21
switched to time interval

derrickstaten:
**Untested / do not merge!**

@shortstuffsushi it should look something like this, if I've done the comparison operation correctly

We can do this in the future, no need to implement today or even this month

Do not thumb up. I just wanted you to see this.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCSystemObserver.m
+++ b/Branch-SDK/Branch-SDK/BNCSystemObserver.m
@@ -131,15 +131,15 @@
     // for creation date
     NSURL *documentsDirRoot = [[[NSFileManager defaultManager] URLsForDirectory:NSDocumentDirectory inDomains:NSUserDomainMask] lastObject];
     NSDictionary *documentsDirAttributes = [manager attributesOfItemAtPath:documentsDirRoot.path error:nil];
-    int appCreationDay = (int)([[documentsDirAttributes fileCreationDate] timeIntervalSince1970]/(60*60*24));
-
+    NSDate *creationDate = [documentsDirAttributes fileCreationDate];
+    
     // for modification date
     NSString *bundleRoot = [[NSBundle mainBundle] bundlePath];
     NSDictionary *bundleAttributes = [manager attributesOfItemAtPath:bundleRoot error:nil];
-    int appModificationDay = (int)([[bundleAttributes fileModificationDate] timeIntervalSince1970]/(60*60*24));
-
+    NSDate *modificationDate = [bundleAttributes fileModificationDate];
+    
     if (!storedAppVersion) {
-        if ([documentsDirAttributes fileCreationDate] && [bundleAttributes fileModificationDate] && (appCreationDay != appModificationDay)) {
+        if (creationDate && modificationDate && [modificationDate timeIntervalSinceDate:creationDate] > 60) {
             return [NSNumber numberWithInt:2];
         }
         return nil;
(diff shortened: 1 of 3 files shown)


Branch-SDK/Branch-SDK/BNCSystemObserver.m
```diff
@@ -131,15 +131,15 @@ + (NSNumber *)getUpdateState {
...
-    int appModificationDay = (int)([[bundleAttributes fileModificationDate] timeIntervalSince1970]/(60*60*24));
-
+    NSDate *modificationDate = [bundleAttributes fileModificationDate];
+    
     if (!storedAppVersion) {
-        if ([documentsDirAttributes fileCreationDate] && [bundleAttributes fileModificationDate] && (appCreationDay != appModificationDay)) {
+        if (creationDate && modificationDate && [modificationDate timeIntervalSinceDate:creationDate] > 60) {
             return [NSNumber numberWithInt:2];
         }
         return nil;
```
shortstuffsushi: We probably shouldn't ever return `nil` here, right?

Branch-SDK/Branch-SDK/BNCSystemObserver.m
```diff
@@ -131,15 +131,15 @@ + (NSNumber *)getUpdateState {
...
     // for modification date
     NSString *bundleRoot = [[NSBundle mainBundle] bundlePath];
     NSDictionary *bundleAttributes = [manager attributesOfItemAtPath:bundleRoot error:nil];
-    int appModificationDay = (int)([[bundleAttributes fileModificationDate] timeIntervalSince1970]/(60*60*24));
-
+    NSDate *modificationDate = [bundleAttributes fileModificationDate];
+    
     if (!storedAppVersion) {
-        if ([documentsDirAttributes fileCreationDate] && [bundleAttributes fileModificationDate] && (appCreationDay != appModificationDay)) {
+        if (creationDate && modificationDate && [modificationDate timeIntervalSinceDate:creationDate] > 60) {
```
shortstuffsushi: This comparison would is comparing 60 seconds, is that correct? I thought you may have said as much as minutes.

shortstuffsushi:
No longer untested, and potentially mergeable given additional review (@aaustin).

Branch-SDK/Branch-SDK/BNCSystemObserver.m
```diff
@@ -131,23 +131,34 @@ + (NSNumber *)getUpdateState {
...
+    NSDate *modificationDate = [bundleAttributes fileModificationDate];
+    
+    // No stored version
     if (!storedAppVersion) {
-        if ([documentsDirAttributes fileCreationDate] && [bundleAttributes fileModificationDate] && (appCreationDay != appModificationDay)) {
-            return [NSNumber numberWithInt:2];
+        // Modification and Creation date are more than 60 seconds different indicates an update
+        // This would be the case that they had un-installed and re-installed the app, since this
+        // value isn't currently in NSDefaults.
+        if (creationDate && modificationDate && [modificationDate timeIntervalSinceDate:creationDate] > 60) {
```
aaustin: do we need to absolute value this?

aaustin:
Looks great! :+1:

Branch-SDK/Branch-SDK/BNCSystemObserver.m
shortstuffsushi: I also considered that, but came to the same conclusion -- the modification date of the bundle couldn't be before the creation date of the the documents directory, right?

derrickstaten:
so abs is really only needed for simulator, but should definitely be in there. love the switch to @1, @2 notation

:+1:

derrickstaten:
if you figure out something with keychain, let's get that in this PR as well

shortstuffsushi:
The keychain research I did indicates that the keychain *is* persisted through uninstall and reinstall. However, we can't grab the creation date of an item in the keychain because we haven't inserted one, in the scenario that this is the first addition of the Branch SDK. We could reevaluate this in the future, but I don't think it's a necessary addition for the time being.

PR #87, opened 2015-04-22
Updating Branch to use NS_ENUMs to represent enum constants.

shortstuffsushi:
@derrickstaten @qinweigong @aaustin @Sarkar 

Also prefixing all enums with Branch. This will be temporarily backwards compatible.

### DIFF
--- a/Branch-SDK/Branch-SDK/Branch.h
+++ b/Branch-SDK/Branch-SDK/Branch.h
@@ -53,9 +53,9 @@ typedef NS_ENUM(NSUInteger, BranchLinkType) {
 };
 
 // Backwards compat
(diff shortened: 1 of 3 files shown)


aaustin:
Why is it temporarily backwards compatible? It still just takes an integer, right?

aaustin:
as long as it doesn't make people go in and change code :+1:

shortstuffsushi:
@aaustin the enums are effectively just constants, yes, but if the developer holds a local reference to any of them (we do so in the [TestBed App here]() for example), that reference would break. I would like to remove this in a future version, perhaps with the callbacks Branch (which I see as a "major" release, since it's a larger rework of the SDK).

PR #89, opened 2015-04-29
placeholder is now an nsurl

derrickstaten:
@shortstuffsushi this is needed for the new Facebook app. without this Facebook is no longer an option in share sheets

### DIFF
--- a/Branch-SDK/Branch-SDK/BranchActivityItemProvider.m
+++ b/Branch-SDK/Branch-SDK/BranchActivityItemProvider.m
@@ -37,21 +37,18 @@
 }
 
 - (id)item {
-    if ([self.placeholderItem isKindOfClass:[NSString class]]) {
-        NSString *channel = [BranchActivityItemProvider humanReadableChannelWithActivityType:self.activityType];
-        
-        // Because Facebook immediately scrapes URLs, we add an additional parameter to the existing list, telling the backend to ignore the first click
-        BOOL ignoreFirstClick = [channel isEqualToString:@"facebook"];
-        if (ignoreFirstClick) {
-            self.branchURL = [[Branch getInstance] getShortURLWithParams:self.params andTags:self.tags andChannel:channel andFeature:self.feature andStage:self.stage andAlias:self.alias ignoreUAString:self.userAgentString];
-        }
-        else {
-            self.branchURL = [[Branch getInstance] getShortURLWithParams:self.params andTags:self.tags andChannel:channel andFeature:self.feature andStage:self.stage andAlias:self.alias];
-        }
-
-        return [NSURL URLWithString:self.branchURL];
+    NSString *channel = [BranchActivityItemProvider humanReadableChannelWithActivityType:self.activityType];
+    
+    // Because Facebook immediately scrapes URLs, we add an additional parameter to the existing list, telling the backend to ignore the first click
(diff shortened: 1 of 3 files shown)


derrickstaten:
Also, we need to push out an update to Cocoapods when we merge this, it's needed for at least one partner and should probably be there for all of them.

shortstuffsushi:
The code looks OK here, but I haven't had a chance to try on a physical device. If you have, and you've tried the various core offerings (Twitter, Facebook, Mail, etc), then you can go ahead and merge :+1: 

As far as tagging, you can either do that when you merge, or I can do it later on today. Keep in mind that there have been a number of updates since the last tag when you're doing release notes. All in all, though, I think they're pretty minor changes, so 0.6.2 would still be an appropriate tag.

derrickstaten:
I'll test across everything I have on my phone now. I'll let you do the actual push since I think you're probably faster at it and know what else belongs in the release notes

derrickstaten:
@shortstuffsushi actually you should probably revisit this. currently we set the placeholder item at an NSString. Then in the item method, we check whether the placeholder item is a string. If it is, we know that we still need to generate a short URL. Perhaps there's a better way to do this?

I tried a few variations but decided that you're probably better suited to do this. Also happy to talk about it and do it myself after a short back-and-forth

derrickstaten:
@shortstuffsushi this is ready now. please test extensively before incorporating

shortstuffsushi:
I think this looks good, and haven't been able to find any issue while testing. I think it would be good if @scotthasbrouck took a look though, since he wrote this initially.

Branch-SDK/Branch-SDK/BranchActivityItemProvider.m
```diff
@@ -37,21 +37,18 @@ - (id)initWithParams:(NSDictionary *)params
...
-            self.branchURL = [[Branch getInstance] getShortURLWithParams:self.params andTags:self.tags andChannel:channel andFeature:self.feature andStage:self.stage andAlias:self.alias];
-        }
-
-        return [NSURL URLWithString:self.branchURL];
+    NSString *channel = [BranchActivityItemProvider humanReadableChannelWithActivityType:self.activityType];
+    
+    // Because Facebook immediately scrapes URLs, we add an additional parameter to the existing list, telling the backend to ignore the first click
+    BOOL ignoreFirstClick = [channel isEqualToString:@"facebook"];
+    if (ignoreFirstClick) {
+        self.branchURL = [[Branch getInstance] getShortURLWithParams:self.params andTags:self.tags andChannel:channel andFeature:self.feature andStage:self.stage andAlias:self.alias ignoreUAString:self.userAgentString];
```
shortstuffsushi: Is there a reason to even have `self.branchURL` as a property?

derrickstaten: i'm not actually sure. i was wondering this myself when we first added it

scotthasbrouck: Totally correct, `self.branchURL` is not needed. Pushing up a commit that removes it now. The rest of this looks solid.

scotthasbrouck:
Ok, sorry for the confusion. That commit removes the `if` check on placeholder, and returns the placeholder to being an NSURL, and is what I meant to commit.

PR #91, opened 2015-05-04
App list opt out

shortstuffsushi:

Adding the ability to opt out of the app list check. Really quick, simple change.


Sarkar:
Looks good to me! :+1:

PR #92, opened 2015-05-04
check for nil to prevent crash in NSJSONSerialization

jeremiah-de:
This prevents NSJSONSerialization from trying to use a nil string returned from dataUsingEncoding. Another option would be to use [NSString dataUsingEncoding:allowLossyConversion:] with allowLossyConversion set to YES.


jeremiah-de:
According to the NSString docs, dataUsingEncoding "Returns nil if ... the receiver can’t be converted without losing some information (such as accents or case)."

shortstuffsushi:
Hey @jeremiah-de, thanks for this! I'm gonna add a couple of test cases around this, and will merge it in tomorrow.

shortstuffsushi:
Hey @jeremiah-de. I added some tests, but pushing them pushed a new branch instead of merging to this pull request. Can we continue on #93?

jeremiah-de:
@shortstuffsushi Go ahead with #93, thanks

PR #93, opened 2015-05-05
Iheartradio encoding fix

shortstuffsushi:
@derrickstaten @Sarkar @jeremiah-de


AdibContractorCC:
:thumbsup:

shortstuffsushi:
@jeremiah-de or @AdibContractorCC -- if you guys have any better test scenarios for constructing strings that will not decode properly, please let me know. The test cases I wrote certainly work, but are not very "real," so a better real-world example would be preferred. Thanks again for submitting this, though, I'll merge it in and release and update now.

PR #96, opened 2015-05-11
LinkData cleanup

shortstuffsushi:

Updating LinkData items (constants, enum) to be in the proper header file.
Removing setObject:forKeyedSubscript: which isn't needed.


qinweigong:
:+1:

PR #97, opened 2015-05-11
Decouple branch dependencies

shortstuffsushi:

Making it possible to inject the several items into the Branch initializer, primarily for the sake of testing.
Removing as much Nocilla as possible to prevent bugs with CircleCI.
The only actual code changes are in the Branch.h and .m files. The rest is tests and CocoaPods.


Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -997,16 +996,15 @@ - (void)applyReferralCode:(NSString *)code andCallback:(callbackWithParams)callb
...
-        NSString *branchKey = [BNCPreferenceHelper getBranchKey:isLive];
-        if (!branchKey || [branchKey isEqualToString:NO_STRING_VALUE]) {
-            NSLog(@"Branch Warning: Please enter your branch_key in the plist!");
-        }
-        
-        [Branch initInstance];
-    }
-    return currInstance;
++ (Branch *)getInstanceInternal {
+    static Branch *branch;
```
aaustin: If it's a static variable scoped to the method, won't this be garbage collected if we're not created in the AppDelegate? Let's say someone inits Branch in VC1, then navigates a few VCs down the funnel causing VC1 to be GCed. My understanding is that a static variable in the class will remain in memory but a method variable could be GC. Unsure here though.

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -91,59 +91,58 @@ @interface Branch() <BNCDebugConnectionDelegate, UIGestureRecognizerDelegate, BN
...
-
 #pragma mark - Public methods
 
 
 #pragma mark - GetInstance methods
 
 + (Branch *)getInstance {
-    return [Branch getBranchInstance:YES];
+    NSString *branchKey = [BNCPreferenceHelper getBranchKey:YES];
+    if (!branchKey || [branchKey isEqualToString:NO_STRING_VALUE]) {
```
aaustin: we should check app key as well here? I thought we had that in prod

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -91,59 +91,58 @@ @interface Branch() <BNCDebugConnectionDelegate, UIGestureRecognizerDelegate, BN
...
+        NSLog(@"Branch Warning: Please enter your branch_key in the plist!");
+    }
+
+    return [Branch getInstanceInternal];
 }
 
 + (Branch *)getTestInstance {
-    return [Branch getBranchInstance:NO];
+    NSString *branchKey = [BNCPreferenceHelper getBranchKey:NO];
+    if (!branchKey || [branchKey isEqualToString:NO_STRING_VALUE]) {
```
aaustin: same here as well

Branch-SDK/Branch-SDK/Branch.m
shortstuffsushi: @aaustin A static variable like this will not be cleaned up, Check out [this question and answer on SO](http://stackoverflow.com/a/15924867/1252541), and feel free to test it out yourself. This is how I create most of my singletons, so there can't be any accidental direct references elsewhere in the code.

Branch-SDK/Branch-SDK/Branch.m
shortstuffsushi: Hmm, that's strange. If you look at line 1003 in the deleted diff, we weren't actually checking there either. We probably should here, though.

Branch-SDK/Branch-SDK/Branch.m
shortstuffsushi: I don't think that we should check app key here, though, because we don't have a concept of "test" app keys.

aaustin: It's probably ok but this can help us prevent people from making the dumb mistake of forgetting to put the key in no matter the instance they choose

Branch-SDK/Branch-SDK/Branch.m
aaustin: Cool sounds good - scratch my concern

Branch-SDK/Branch-SDK/Branch.m
shortstuffsushi: I'm moving these checks to the init session now. We were checking Branch Key stuff in a couple different places within the Branch file.

shortstuffsushi: Actually, having a check here just to have the appropriate warnings would be good. If they specify get test instance, but they have a live key, that's not gonna work.

shortstuffsushi: Hmm, on second look, we're not really doing this right. We never actually specifically make sure we're using a test instance, and we never use that in the server interface. I will make a separate PR to address this.

aaustin:
Looks great :+1:

PR #99, opened 2015-05-12
Clearing Items when Branch Key Changes

shortstuffsushi:


shortstuffsushi:
Currently, existing requests will still potentially have bad items built into them. We can either ensure that all of these requests are removed, or we can just leave them for now. In a future pull request, I'll be making the requests more "dynamic," so they won't actually build up their request until the point of actually being made, which will fix this problem.

Since this isn't a very common use case, and isn't really something we've had much trouble with currently, I'd prefer to hold off on adding logic to remove them, as it won't be necessary in the near future.

aaustin:
We should clear the persisted queue as well just in case

shortstuffsushi:
@aaustin I debated that, but I wasn't sure. Throwing the whole thing out means you lose any pending requests. Given the use case for changing the key, that's probably appropriate, though.

aaustin:
Yea -  I think it's an easy one. Will approve after that's done

aaustin:
:+1:

PR #103, opened 2015-05-19
Remove tag from server interface

shortstuffsushi:
@aaustin @derrickstaten @Sarkar @qinweigong

Title pretty much says it all. Tag isn't needed anymore, removing from server interface and response.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCServerInterface.h
+++ b/Branch-SDK/Branch-SDK/BNCServerInterface.h
@@ -18,16 +18,16 @@ static NSString *REQ_TAG_DEBUG_DISCONNECT = @"t_debug_disconnect";
 
 @interface BNCServerInterface : NSObject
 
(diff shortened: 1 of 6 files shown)


derrickstaten:
what was tag before? so curious

qinweigong:
:+1:

shortstuffsushi:
@derrickstaten In the past, the request tag was being passed into all the method in order to be added to the response so it could fall through the big ol `if` tree to find the appropriate handling. This was no longer necessary with the callbacks branch. Could probably have been removed there, but I missed it and it was already a huge diff.

shortstuffsushi:
@derrickstaten For reference, [this](https://github.com/BranchMetrics/Branch-iOS-SDK/blob/3e045d90290c18ea2ca753d49ee6bfcf1acf0a4b/Branch-SDK/Branch-SDK/Branch.m#L1618)

derrickstaten:
ah, cool. thanks.

PR #104, opened 2015-05-19
Remove pod dependencies

shortstuffsushi:

Now that CircleCI is more stable, we don't need to have these checked in anymore. Instead, we can just go back to relying on the `Podfile.lock`.


qinweigong:
Cool :+1:

PR #107, opened 2015-05-20
Fix key usage

shortstuffsushi:

Rather than letting BNCServerInterface choose a key, it should be provided with one for each request.
This means adding key to all the methods (yay, more injection).
Prior to this change, when trying to use test mode, the server interface would never have actually sent the test key.
Not entirely confident that the Preference helper is doing this right, but that's going to be another change.

Also removed some unused methods from the BranchServerInterface.
Consolidated param prep within BNCServerInterface into a single method.


aaustin:
Looks good :+1:

PR #108, opened 2015-05-20
Rework debug logic

shortstuffsushi:

Updating PreferenceHelper debug to only use two booleans, once indicating whether the app is in debug mode, when indicating with it is connected to remote debug.
Removing bnc_ prefixes from a bunch of methods in Branch.m.
Making all the variables related to debug instance variables (and setDebug an instance method).
Updating open/install debug checks to just use `isDebug`.

### DIFF
--- a/Branch-SDK/Branch-SDK/Branch.m
+++ b/Branch-SDK/Branch-SDK/Branch.m
@@ -215,6 +215,10 @@ static int BNCDebugTriggerFingersSimulator = 2;
 
 #pragma mark - Configuration methods
 
++ (void)setDebug {
+    [[Branch getInstance] setDebug];
+}
+
 - (void)setDebug {
     [BNCPreferenceHelper setDebug];
 }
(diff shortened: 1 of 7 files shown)


qinweigong:
Looks good :+1:

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -216,8 +215,8 @@ + (BranchActivityItemProvider *)getBranchActivityItemWithParams:(NSDictionary *)
 
 #pragma mark - Configuration methods
 
-+ (void)setDebug {
-    [BNCPreferenceHelper setDevDebug];
+- (void)setDebug {
```
aaustin: Can we leave the static method with a deprecation note?

shortstuffsushi: Because `setDebug` is a non-permanent method (for testing purposes only), I figured we could just directly remove it. Thoughts?

aaustin: but there are likely hundreds of apps with static method commented out or in some debug logic that will now break the next time they update the SDK files. There's nothing worse than that when you're in a hurry

aaustin:
:+1:

PR #109, opened 2015-05-21
Remove all instances of NO_STRING_VALUE.

shortstuffsushi:
The server doesn't want this, and internally we're basically just using it as a nil check.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
@@ -181,15 +181,7 @@ static NSString *Branch_Key = nil;
 }
 
 + (NSString *)getAppKey {
-    NSString *ret = [[[NSBundle mainBundle] infoDictionary] objectForKey:KEY_APP_KEY];
-    if (!ret || ret.length == 0) {
-        // for backward compatibility
-        ret = [BNCPreferenceHelper readStringFromDefaults:KEY_APP_KEY];
-        if (!ret) {
-            ret = NO_STRING_VALUE;
-        }
-    }
-    return ret;
+    return [[[NSBundle mainBundle] infoDictionary] objectForKey:KEY_APP_KEY];
 }
 
 + (void)setAppKey:(NSString *)appKey {
@@ -217,10 +209,6 @@ static NSString *Branch_Key = nil;
         }
     }
     
-    if (!key || key.length == 0) {
-        key = NO_STRING_VALUE;
-    }
-    
     [BNCPreferenceHelper setBranchKey:key];
     
     return key;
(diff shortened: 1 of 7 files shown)


Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
```diff
@@ -181,15 +181,7 @@ + (NSInteger)getRetryCount {
 }
 
 + (NSString *)getAppKey {
-    NSString *ret = [[[NSBundle mainBundle] infoDictionary] objectForKey:KEY_APP_KEY];
-    if (!ret || ret.length == 0) {
-        // for backward compatibility
-        ret = [BNCPreferenceHelper readStringFromDefaults:KEY_APP_KEY];
```
qinweigong: Do we want to keep this backup for now?

shortstuffsushi: I don't think so -- if you recall, we decided that we needed to stop pulling these items from defaults as it could cause unexpected behavior when a key is (perhaps accidentally) removed, and the app keeps working.

For instance, say I used key "asdf" at some point, which was then stored into defaults (I don't think anything stores into defaults for key anymore, but say in a past version it did). If I then strip out that line of code and release, the app will incidentally continue to work for users who already had the app, but won't work for new users.

I'm open to discussion on this, but I'm of the opinion that we shouldn't be pulling from there.

qinweigong: `If I then strip out that line of code and release, the app will incidentally continue to work for users who already had the app, but won't work for new users.` I am confused...you are already stripping it out right?

shortstuffsushi: I'm talking in a hypothetical situation.

Say an early adopter used a version of Branch that saved to defaults with key "asdf."
In a later version, the removed the key "asdf," but didn't add a new key.
When that app is released, new users would have no key, old users would incidentally still get the "asdf" key.

That's the scenario I meant -- it was the same w/ the `setRetryInterval` methods and such that were removed a few versions back.

qinweigong: Yes, I agree. So what you are expecting is our SDK should not work with either old users or new users in this case (no key is specified).
What this line was meant to do was to "save" the old user when the developer forgets to put in the key (either branch key or app key), i.e. at least those old users could still use the SDK correctly. Guess this would be a rare case, but just want to have some backup.

shortstuffsushi: I was leaning toward it not working for either, so that this problem would be rooted out potentially before going live -- either the tester would realize because it'd blow up for them, or the beta testers would see it. Also, when was this feature present in Branch, any idea? It seems we moved away from saving into defaults a while ago. Will try to find reference.

shortstuffsushi: It appears that that would have been part of the `0.4.3` tag, for what that's worth.

qinweigong: Way before the introduction of Branch key, when we used to store app key in defaults, but moved it to plist. It was at that time this "backward compatibility" line of code was added, I think.
But, I agree on your point. We should disable for both old and new users in case the Branch key is missing.

qinweigong:
:+1:

PR #113, opened 2015-05-27
Split out requests

shortstuffsushi:

The idea here is to further remove logic from the `Branch.m` file, with the goal eventually being that it does pretty much nothing itself.

The majority of the diff here is breaking each actual request into it's own file. For every deletion, you'll notice a corresponding request creation.

I need to clean up a few more things before this can be merged (the BranchServerInterface will be entirely removed), at this point I'm just looking for feedback on the idea.


Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
```diff
@@ -75,45 +76,24 @@ + (BNCPreferenceHelper *)getInstance {
...
-}
-
-+ (void)disconnectRemoteDebug {
-    BNCPreferenceHelper *preferenceHelper = [BNCPreferenceHelper getInstance];
-    if (preferenceHelper.isConnectedToRemoteDebug) {
-        preferenceHelper.isConnectedToRemoteDebug = NO;
-        
-        [serverInterface disconnectFromDebugWithKey:Branch_Key callback:NULL];
-    }
+    [BNCPreferenceHelper getInstance].isConnectedToRemoteDebug = NO;
```
qinweigong: Is this difference between clearDebug and setDebug:NO on purpose? or should setDebug:NO do this line too? If so, could refactor?

Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
```diff
@@ -191,33 +154,32 @@ + (void)setAppKey:(NSString *)appKey {
...
-    NSString *key = nil;
+    BNCPreferenceHelper *instance = [BNCPreferenceHelper getInstance];
+    NSString *key = instance.branchKey;
+    
+    if (key && isLive == instance.isUsingLiveKey) {
+        return key;
+    }
     
-    id ret = [[[NSBundle mainBundle] infoDictionary] objectForKey:KEY_BRANCH_KEY];
+    id ret = [[[NSBundle mainBundle] infoDictionary] objectForKey:@"branch_key"];
```
qinweigong: why not use the constant?

Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -172,7 +172,7 @@ - (NSDictionary *)prepareParamDict:(NSDictionary *)params key:(NSString *)key re
     fullParamDict[@"retryNumber"] = @(retryNumber);
     
     if ([key hasPrefix:@"key_"]) {
-        fullParamDict[KEY_BRANCH_KEY] = key;
+        fullParamDict[@"branch_key"] = key;
```
qinweigong: and here

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -7,7 +7,7 @@
 //
 
 #import "Branch.h"
-#import "BranchServerInterface.h"
+#import "BNCServerInterface.h"
```
qinweigong: Do we still need to import it here now that it's already imported in the header file?

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -18,6 +18,26 @@
...
+#import "BranchShortUrlRequest.h"
+#import "BranchShortUrlSyncRequest.h"
+#import "BranchCloseRequest.h"
+#import "BranchGetAppListRequest.h"
+#import "BranchUpdateAppListRequest.h"
+#import "BranchOpenRequest.h"
+#import "BranchInstallRequest.h"
+#import "BranchConnectDebugRequest.h"
+#import "BranchDisconnectDebugRequest.h"
+#import "BranchLogRequest.h"
```
qinweigong: Like this OOP design :-)

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1004,35 +714,16 @@ - (void)applyReferralCode:(NSString *)code andCallback:(callbackWithParams)callb
...
-        if (callback) {
-            callback(response.data, error);
-        }
-    };
 
-    [self.requestQueue enqueue:req];
+#pragma mark - Logging
+- (void)log:(NSString *)log {
+    BranchLogRequest *request = [[BranchLogRequest alloc] initWithLog:log];
+    [self.requestQueue enqueue:request];
```
qinweigong: Correct me if I have mis-understanding here. The initial design had two queues, one to handle "real" requests, and the other to handle log/debug related requests. The purpose is to avoid having impact on the main queue as much as possible. Seems now they are mixed together?

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1132,12 +791,13 @@ - (NSString *)generateShortUrl:(NSArray *)tags andAlias:(NSString *)alias andTyp
         shortURL = [self.linkCache objectForKey:linkData];
     }
     else {
-        req.postData = linkData.data;
+        // TODO still need to figure out sync here
```
qinweigong: // TODO :-)

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1503,111 +1043,49 @@ - (void)initUserSessionAndCallCallback:(BOOL)callCallback {
 }
 
 - (void)initializeSession {
-    if (![self hasBranchKey] && ![self hasAppKey]) {
+    if (!self.branchKey) {
```
qinweigong: So we are dropping app key from this version on?

aaustin:
Seems to be an issue with the Share Sheet code and this PR - crashes the TestBed app 100% of time

Branch-SDK/Branch-SDK/BranchOpenRequest.m
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    params[@"identity_id"] = [BNCPreferenceHelper getIdentityID];
+    params[@"ad_tracking_enabled"] = @([BNCSystemObserver adTrackingSafe]);
+    params[@"is_referrable"] = @([BNCPreferenceHelper getIsReferrable]);
+    params[@"debug"] = @([BNCPreferenceHelper isDebug]);
+
+    [self safeSetValue:[BNCSystemObserver getAppVersion] forKey:@"app_version" onDict:params];
+    [self safeSetValue:[BNCSystemObserver getOS] forKey:@"os" onDict:params];
+    [self safeSetValue:[BNCSystemObserver getOSVersion] forKey:@"os_versionN" onDict:params];
```
aaustin: mistyped key...

Branch-SDK/Branch-SDK/BNCServerRequest.m
```diff
@@ -1,61 +1,29 @@
...
-        self.postData = [coder decodeObjectForKey:DATA];
-    }
-    return self;
+- (void)makeRequest:(BNCServerInterface *)serverInterface key:(NSString *)key callback:(BNCServerCallback)callback {
+    [NSException raise:@"Improper BNCServerRequest" format:@"BNCServerRequest subclasses must implement makeRequest:key:callback:"];
 }
 
-- (id)initWithTag:(NSString *)tag {
-    return [self initWithTag:tag andData:nil];
+- (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
```
qinweigong: Now that we are going more OOP, semantically, feel it shouldn't be server request's job to process response. Maybe this should be sort of a "manager", which manages both server requests and responses.

Branch-SDK/Branch-SDK/Branch.m
aaustin: We need to support the app key still. No forced migration yet

qinweigong:
generally, like the hierarchical OOP design to handle server requests...just the number of source files dramatically increases :-)

Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
shortstuffsushi: `clearDebug` is basically shorthand for calling `setDebug:NO` `setRemoteDebug:NO`. Could be removed, just something that already existed that I didn't change.

Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
shortstuffsushi: I'm not sure. It felt weird that the BNCPreferences defined that constant, and that they were just kinda strewn about all over. Maybe I'll make a constants file for that and some of the other items.

Branch-SDK/Branch-SDK/Branch.m
shortstuffsushi: Nope. Will be removed.

Branch-SDK/Branch-SDK/Branch.m
shortstuffsushi: Perhaps I was the one misunderstanding. I didn't know why we had two queues, so I just merged them. I didn't like that the PreferenceHelper had it's own queue, it felt strange for a preference helper. I could make a separate one here, though.

Branch-SDK/Branch-SDK/Branch.m
shortstuffsushi: Hmm... this was actually done, apparently I just forgot to remove my note.

shortstuffsushi:
The number of requests does get large here, but it makes each individual request testable. You can check that it hits the right end point with the right params, and that it handles responses appropriately. It also makes dropping in a new request a lot easier (and lower barrier of entry, imo -- easier pattern to follow). It also removes a ton of logic from the Branch file itself. It would be nice to remove as much as possible from there.

Branch-SDK/Branch-SDK/BNCServerRequest.m
shortstuffsushi: The way I see it, the request knows how to construct itself and send to the server, and how to process the response. That way, as a consumer all I do is this:

```
MyRequest *req = [[MyRequest alloc] initWithCallback:^(NSDictionary *params, NSError *error) {
    if (err) {
        NSLog(@"Request Failed");
    }
    else {
        NSLog("Request completed with params: %@", params);
    }
}];

[req send];// (or in our case, enqueue)
```

shortstuffsushi:
@aaustin I'm not able to reproduce a crash with share sheet, can you tell me where you're seeing it?

aaustin:
Oh weird - now it seems alright on that crash...

aaustin:
Wait - reproduced it with more info
screen shot 2015-06-06 at 9 43 21 pm

aaustin:
Just press it repeatedly - now it's showing this
screen shot 2015-06-06 at 9 44 27 pm

shortstuffsushi:
@aaustin I've been able to reproduce this on master now, so it seems to be an unrelated issue. I'd like to take that out of this PR and just make a separate fix (though it's not clear why it's failing yet :disappointed:).

Branch-SDK/Branch-SDK/Branch.m
shortstuffsushi: App Key is still supported, look at the getInstance and getTestInstance methods -- both do fallback on the getAppKey method. The `self.branchKey` property is only filled in once, but can have either a Branch Key or an App Key.

aaustin:
This is weird, but it seems that the first install doesn't return the data. See my logs below. There should be keys/values in the init callback, right?

2015-06-08 12:02:20.055 Branch-TestBed[3519:2371876] [BNCServerInterface.m:197] returned = Status: 200; Data: {
    "browser_fingerprint_id" = "<null>";
    data = "{\"+is_first_session\":true,\"+clicked_branch_link\":false}";
    "device_fingerprint_id" = 136532620564535860;
    "identity_id" = 136532620614836909;
    link = "https://bnc.lt/i/5Z1v5k-08h";
    "session_id" = 136532620782609078;
}
2015-06-08 12:02:20.056 Branch-TestBed[3519:2371876] finished init with params = {
}

aaustin:
Confirmed with an actual link click as well:

2015-06-08 12:05:04.325 Branch-TestBed[3534:2373171] [BNCServerInterface.m:197] returned = Status: 200; Data: {
    "browser_fingerprint_id" = "<null>";
    data = "{\"$og_image_url\":\"https://s3-us-west-1.amazonaws.com/branchhost/mosaic_og.png\",\"$og_title\":\"Kindred\",\"key1\":\"test_object\",\"key2\":\"here is another object!!\",\"~channel\":\"facebook\",\"~feature\":\"invite\",\"~tags\":[\"tag1\",\"tag2\"],\"~stage\":\"2\",\"~creation_source\":\"iOS SDK\",\"+is_first_session\":true,\"+clicked_branch_link\":true}";
    "device_fingerprint_id" = 136533309478967339;
    "identity_id" = 136533309579600044;
    link = "https://bnc.lt/i/5Z237NU00g";
    "referring_identity" = test1235;
    "session_id" = 136533309692846256;
}
2015-06-08 12:05:04.326 Branch-TestBed[3534:2373171] finished init with params = {
}

aaustin:
Let's do this :+1:

PR #114, opened 2015-05-27
Error Code Update

shortstuffsushi:

Separating out failure codes for a bad user request and a Branch server problem.
Fixing check to continue processing in Branch.


qinweigong:
:+1:

PR #115, opened 2015-05-28
Expose Identity State

shortstuffsushi:
Making it possible for developers to understand whether Branch has a developer identity set for this User.
Previously, it wasn't possible for developers to know if Branch already had a developer identity set for a User, which persists through installs.


Sarkar:
this is a good one! :+1:

qinweigong:
:+1:

PR #116, opened 2015-05-28
Create framework

shortstuffsushi:

This is a pretty simple PR that just adds a couple scripts for creating the framework and uploading it to S3. The next step would be adding it to CircleCI to automate uploads.


Sarkar:
hellllll yeah! :100:

PR #117, opened 2015-05-29
Update Update State

shortstuffsushi:
@derrickstaten 

Change update state check to one day to account for "long downloads" which may cause the doc dir and bundle to have > 1 min difference in timestamp.


derrickstaten:
:+1:

PR #118, opened 2015-06-01
Dont add source in encoding

shortstuffsushi:

Source shouldn't be added to every dictionary that is encoded, it should only be added where desired. Specifically, it should be added to url generation, where the creation source is present.

### DIFF
--- a/Branch-SDK/Branch-SDK/Branch.m
+++ b/Branch-SDK/Branch-SDK/Branch.m
@@ -1185,7 +1185,7 @@ static int BNCDebugTriggerFingersSimulator = 2;
     
     NSData *jsonData = [BNCEncodingUtils encodeDictionaryToJsonData:params];
     NSString *base64EncodedParams = [BNCEncodingUtils base64EncodeData:jsonData];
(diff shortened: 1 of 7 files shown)


Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1204,13 +1204,7 @@ - (BNCLinkData *)prepareLinkDataFor:(NSArray *)tags andAlias:(NSString *)alias a
...
     [post setupMatchDuration:duration];
     [post setupIgnoreUAString:ignoreUAString];
-    
-    NSString *args = @"{\"source\":\"ios\"}";
-    if (params) {
-        args = [BNCEncodingUtils encodeDictionaryToJsonString:params];
-    }
-    
-    [post setupParams:args];
+    [post setupParams:[BNCEncodingUtils encodeDictionaryToJsonString:params]];
```
aaustin: Can params be nil?

shortstuffsushi: Yep -- and that's fine. It'll become an empty object on the post.

aaustin:
:+1:

PR #123, opened 2015-06-05
Readme.md Update

LisaStucky:
No need to review this now; it's just a work in progress.


shortstuffsushi:
The way the individual methods are broken up feels really cluttered, with it being something like:

###Methods
###Parameters
###Methods
###Parameters

It's not clear as a user what method I'm actually looking at just by skimming.

shortstuffsushi:
More specifically, this is what I'm referring to:

screen shot 2015-06-11 at 2 18 55 pm

Instead of having textual headers, why not just use the method names? The textual headers could be a short description below.

Branch-TestBed/Branch-TestBed.xcodeproj/project.pbxproj
```diff
@@ -595,7 +595,7 @@
 						CreatedOnToolsVersion = 6.3.2;
 					};
 					6700165F1940F51400A9E103 = {
-						DevelopmentTeam = 6F3Q27CVFQ;
+						DevelopmentTeam = T5TQ36Q2SQ;
```
shortstuffsushi: Note to self: revert this before merging.

README.md
```diff
@@ -217,43 +175,72 @@ func application(application: UIApplication, openURL url: NSURL, sourceApplicati
...
+####Parameters
+
+None
 
-If you ever want to access the original session params (the parameters passed in for the first install event only), you can use this line. This is useful if you only want to reward users who newly installed the app from a referral link or something.
+####Returns
+
+**NSDictionary *** When initSession returns a parameter set in the deep link callback, we store it in NSUserDefaults for the duration of the session in case you want to retrieve it later. Careful, once the app is minimized and the session ends, this will be cleared.
+
+###Retrieve Install (Install Only) Parameters
```
shortstuffsushi: These are updated in other instances, fwiw. When you setIdentity, for instance, if that identity already existed on the backend, the install params for that identity will be set within the app.

README.md
```diff
@@ -870,18 +928,39 @@ Branch.getInstance().getReferralCodeWithPrefix("BRANCH",
...
+
+### Validate Referral Code
 
 Validate if a referral code exists in Branch system and is still valid.
 A code is vaild if:
 
-1. It hasn't expired.
-1. If its calculation type is uniqe, it hasn't been applied by current user.
+* It hasn't expired.
+* If its calculation type is uniqe, it hasn't been applied by current user.
```
shortstuffsushi: Typo: unique.

Also, not sure if this is important, but it may be useful to mention it's also valid if it's type unlimited, as long as it hasn't expired.

README.md
```diff
@@ -675,23 +729,27 @@ The response will return an array that has been parsed from the following JSON:
...
+1. _0_ - A reward that was added automatically by the user completing an action or referral.
+1. _1_ - A reward that was added manually.
+2. _2_ - A redemption of credits that occurred through our API or SDKs.
+3. _3_ - This is a very unique case where we will subtract credits automatically when we detect fraud.
 
-1. _0_ - A reward that was added automatically by the user completing an action or referral
-1. _1_ - A reward that was added manually
-2. _2_ - A redemption of credits that occurred through our API or SDKs
-3. _3_ - This is a very unique case where we will subtract credits automatically when we detect fraud
+### Get Referral Code
```
shortstuffsushi: We are working on renaming all of these methods to Promo Code, could you please update all references accordingly?

Referral URLs are staying the same, just Referral Code changes.

README.md
```diff
@@ -597,10 +640,16 @@ Branch().loadRewardsWithCallback { (changed: Bool, error: NSError!) -> Void in
...
 ```
 
-### Redeem all or some of the reward balance (store state)
+####Parameters
+
+**callback**: The callback that is called once the request has completed.
+
+### Redeem All or Some of the Reward Balance (Store State)
 
 We will store how many of the rewards have been deployed so that you don't have to track it on your end. In order to save that you gave the credits to the user, you can call redeem. Redemptions will reduce the balance of outstanding credits permanently.
```
shortstuffsushi: This description is really confusing to me. Yes, we do keep track of the number of credits you have, but this method specifically is about *spending* those credits. It should probably be re-worded to say as much.

PR #124, opened 2015-06-09
Share Sheet Fix

shortstuffsushi:
@aaustin 

Something in the base64 encoding was causing problems when creating the activity item provider. This just changes that impl and now it works.


aaustin:
:+1: god help us all

PR #130, opened 2015-06-17
Dissociative Callbacks

shortstuffsushi:
@derrickstaten 

When an Open/Install request is persisted to disk, its callback is lost.
This means that even when it is actually executed, the Branch file is never informed of the update, and never marks itself as initialized.
Now, we always re-add the callback to make sure that link is still in place.

### DIFF
--- a/Branch-SDK/Branch-SDK/Branch.m
+++ b/Branch-SDK/Branch-SDK/Branch.m
@@ -1071,9 +1071,7 @@ static int BNCDebugTriggerFingersSimulator = 2;
     // Make sure a callback is associated with this request. This callback can
     // be cleared if the app is terminated while an Open/Install is pending.
     else {
-        [self.requestQueue moveInstallOrOpenToFront:self.networkCount];
-        
-        BranchOpenRequest *req = (BranchOpenRequest *)[self.requestQueue peek];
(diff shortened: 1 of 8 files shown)


Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1055,20 +1052,29 @@ - (void)initializeSession {
...
-        }];
+        BranchOpenRequest *req = [[clazz alloc] initWithCallback:initSessionCallback];
 
         [self insertRequestAtFront:req];
     }
+    // If there is already one in the queue, make sure it's in the front.
+    // Make sure a callback is associated with this request. This callback can
+    // be cleared if the app is terminated while an Open/Install is pending.
     else {
         [self.requestQueue moveInstallOrOpenToFront:self.networkCount];
```
qinweigong: it'd be nice for moveInstallOrOpenToFront to return the req after the move directly, such that you don't need a peek after this to get the req.

qinweigong:
:+1:

Branch-SDK/Branch-SDK/Branch.m
shortstuffsushi: I think you're right. Peeking after the request is wrong here, since it could be in the first or second spot.

PR #131, opened 2015-06-22
Auto deep link

shortstuffsushi:
@aaustin


aaustin:
This looks great. Can you think if about the case where they want to re-use an existing view controller and want to different the view configuration logic from a Branch launched version and the version launched by their own app. Basically, if there's a way we could call that configuration delegate in the scenario where the view controller is launched elsewhere. Think about this for a little bit - otherwise, we should document that.

shortstuffsushi:
I'm not entirely sure what you mean -- the basic idea would be something like this:

Most people implement their view layout logic in `viewDidLoad` or `viewWill/DidAppear`. In that scenario, they've already passed in data that controller needs from the previous view / context (typically in `prepareForSegue`). Here's a simple example:

```
@implementation FirstViewController

- (void)prepareForSegeue:(UISegue *)segue {
    if ([segue.identifier isEqualToString:@"TheSegue"]) {
        SecondViewController *controller = segue.destinationController;
        controller.someContextSensitiveProp = @{ @"some": @"stuff" };
    }
}

@end

@interface SecondController

@property (strong, nonatomic) NSDictionary *someContextSensitiveProp;
@property (weak, nonatomic) IBOutlet UILabel *label;

@end

@implementation SecondViewController

- (void)viewDidLoad {
    [super viewDidLoad];

    self.label.text = self.someContextSensitiveProp[@"foo"];
}

@end

```

Now, to alter this to allow Branch's auto-deeplinking to work, all they'd have to do is conform to the protocol and get `someContextSensitiveProp` from the config data, as such:

```
@interface SecondController <BranchDeepLinkingController>

@property (strong, nonatomic) NSDictionary *someContextSensitiveProp;
@property (weak, nonatomic) IBOutlet UILabel *label;

@end

@implementation SecondViewController

- (void)viewDidLoad {
    [super viewDidLoad];

    self.label.text = self.someContextSensitiveProp[@"foo"];
}

- (void)configureWithData:(NSDictionary *)data {
    self.someContextSensitiveProp = data[@"someContextSensitiveProp"];
}

@end
```

We call the configure method before the view lifecycle, so the rest of the code functions exactly the same way -- just the properties are set in a different location. Does that make sense?

aaustin:
I got it. That's great. We just need to add some explicit indication that this delegate method will be guaranteed to execute before any lifecycle call. It's obvious to me how to use it after you show the demo, but I doubt it would be obvious to any junior dev approaching it. This is kind of a like a best practices recommendation.

Can you update the readme and make a documentation PR for these two pages below? It'd be great to show how to subclass and talk about how to use it. The example you have above, while verbose in current form, could be good if we can cut it down to a paragraph or so.
https://dev.branch.io/recipes/content_sharing/ios/#routing-to-content-within-your-ios-app
https://dev.branch.io/recipes/personalized_welcome/ios/#the-personal-touch-a-welcome-screen

PR #133, opened 2015-06-22
Removing a whole bunch of false-safety nets.

shortstuffsushi:
@qinweigong

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCServerRequestQueue.m
+++ b/Branch-SDK/Branch-SDK/BNCServerRequestQueue.m
@@ -194,23 +194,17 @@ NSUInteger const BATCH_WRITE_TIMEOUT = 3;
 }
 
 - (void)persistToDisk {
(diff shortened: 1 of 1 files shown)


qinweigong:
I remember I tried this "work on a copy" approach in the past, but there seemed to be some synchronization problem still.
You might want to have some more tests (maybe more load tests) for this.

qinweigong:
:+1: if tests good.

shortstuffsushi:
We may need to use better synchronization around the entirety of this class. Currently there are no guarantees at all :disappointed: The `@synchronized` block that was in there doesn't actually do anything, except preventing multiple calls to *that* method from going through at once -- the rest of the class can continue to modify that queue (which is almost certainly what's happening here). By taking a copy, we're promising ourselves that, at minimum, the requests we are planning to save to disk right now will not be cleared out from under us.

I think this should be revisited in the future, though.

qinweigong:
Definitely.

PR #134, opened 2015-06-23
Non singleton preference helper

shortstuffsushi:
@qinweigong 

A couple things here:

1. Moving away from a singleton for the PreferenceHelper (though this isn't really done, since they're all grabbing the global reference).
2. Making properties on the instance, so that they don't have to be reloaded from NSUserDefaults every time.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
@@ -15,6 +15,8 @@ static const NSInteger DEFAULT_RETRY_INTERVAL = 0;
 static const NSInteger DEFAULT_RETRY_COUNT = 1;
 static const NSInteger APP_READ_INTERVAL = 520000;
 
+NSString * const BRANCH_PREFS_FILE = @"BNCPreferences";
+
 NSString * const BRANCH_PREFS_KEY_APP_KEY = @"bnc_app_key";
 NSString * const BRANCH_PREFS_KEY_APP_VERSION = @"bnc_app_version";
 NSString * const BRANCH_PREFS_KEY_LAST_RUN_BRANCH_KEY = @"bnc_last_run_branch_key";
@@ -40,6 +42,7 @@ NSString * const BRANCH_PREFS_KEY_UNIQUE_BASE = @"bnc_unique_base_";
 
 @interface BNCPreferenceHelper ()
 
+@property (strong, nonatomic) NSMutableDictionary *persistenceDict;
 @property (strong, nonatomic) NSMutableDictionary *countsDictionary;
 @property (strong, nonatomic) NSMutableDictionary *creditsDictionary;
 @property (assign, nonatomic) BOOL isUsingLiveKey;
@@ -67,7 +70,7 @@ NSString * const BRANCH_PREFS_KEY_UNIQUE_BASE = @"bnc_unique_base_";
         
         _isDebug = NO;
         _isConnectedToRemoteDebug = NO;
-        _isReferrable = [BNCPreferenceHelper readBoolFromDefaults:KEY_IS_REFERRABLE];
+        _isReferrable = [self readBoolFromDefaults:BRANCH_PREFS_KEY_IS_REFERRABLE];
     }
     
     return self;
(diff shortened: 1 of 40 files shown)


Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
```diff
@@ -76,218 +90,239 @@ + (BNCPreferenceHelper *)getInstance {
...
+    if (![_userUrl isEqualToString:userUrl]) {
+        _userUrl = userUrl;
+        [self writeObjectToDefaults:BRANCH_PREFS_KEY_USER_URL value:userUrl];
+    }
 }
 
-+ (void)setIsReferrable {
-    [BNCPreferenceHelper writeBoolToDefaults:KEY_IS_REFERRABLE value:YES];
+- (BOOL)getIsReferrable {
+    return _isReferrable;
```
qinweigong: should this be read from defaults too?

Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
```diff
@@ -298,126 +333,142 @@ + (BOOL)getNeedAppListCheck {
...
-+ (NSDictionary *)getCreditsDictionary {
-    NSDictionary *dict = (NSDictionary *)[BNCPreferenceHelper readObjectFromDefaults:KEY_CREDITS];
-    if (!dict)
-        dict = [[NSDictionary alloc] init];
-    return dict;
-}
-
-+ (void)setCreditsDictionary:(NSDictionary *)credits {
-    [BNCPreferenceHelper writeObjectToDefaults:KEY_CREDITS value:credits];
+- (NSDictionary *)getCreditsDictionary {
```
qinweigong: shall we change the return type to NSMutableDictionary * here?

Branch-SDK/Branch-SDK/BNCServerInterface.h
```diff
@@ -18,6 +19,8 @@ static NSString *REQ_TAG_DEBUG_DISCONNECT = @"t_debug_disconnect";
 
 @interface BNCServerInterface : NSObject
 
+@property (strong, nonatomic) BNCPreferenceHelper *preferenceHelper;
```
qinweigong: Not a big deal, just feel a bit inconsistent: in some cases, e.g. here, pref helper is a property, and used like [self.preferenceHelper ...]; in some other cases, it's used with the singleton syntax, e.g. [[BNCPeferenceHelper preferenceHelper] ...].

Branch-SDK/Branch-SDK/BNCServerRequestQueue.m
```diff
@@ -47,7 +45,7 @@ - (void)enqueue:(BNCServerRequest *)request {
 - (void)insert:(BNCServerRequest *)request at:(unsigned int)index {
     @synchronized(self.queue) {
         if (index > self.queue.count) {
-            [BNCPreferenceHelper log:FILE_NAME line:LINE_NUM message:@"Invalid queue operation: index out of bound!"];
+            [[BNCPreferenceHelper preferenceHelper] log:FILE_NAME line:LINE_NUM message:@"Invalid queue operation: index out of bound!"];
```
qinweigong: this is the singleton syntax i mentioned above. again, not a big deal, but better have some consistency.

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -942,7 +939,7 @@ - (void)callClose {
     if (self.isInitialized) {
         self.isInitialized = NO;
 
-        if ([self hasSession] && ![self.requestQueue containsClose]) {
+        if (![self.requestQueue containsClose]) {
```
qinweigong: i noticed the [self hasSession] check was added recently, just wanna make sure you are not removing it here by accident.

Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
shortstuffsushi: It is... but because it's a boolean, it can't be "lazy loaded" like all the objects in here. We instead have to load it [at initialization time](https://github.com/BranchMetrics/Branch-iOS-SDK/pull/134/files#diff-4db16133408e16908e5c2f5bfa6038e5R73) :(

Branch-SDK/Branch-SDK/BNCServerInterface.h
shortstuffsushi: Yes, I want to move toward everything have an *instance* of the PreferenceHelper rather than grabbing the global one. This makes testing easier (read: DI), but it's too much of a hassle to try to get all into this PR, so I just did it in the one, easy spot.

Branch-SDK/Branch-SDK/Branch.m
shortstuffsushi: Hmm, this was an accidental / "partial" remove. This was really just a band aid to figure out *why* session id was being lost (the root of which is, I believe, because defaults were being cleared). I believe this PR will fix that so this check shouldn't be necessary, but I will revert this for now until I've had more time to prove that.

qinweigong:
:+1:

PR #135, opened 2015-06-23
Fix update state check

shortstuffsushi:
@derrickstaten


derrickstaten:
@shortstuffsushi I'm not sure what the logic is but it doesn't seem to pick the right state in some cases.

- isReferrable not set as part of initSession
- on a background and re-foreground on first launch. It just keeps sending up is_referrable: 1. 
- once I kill the app and launch a second time, is_referrable is 0 as expected.

So far all other cases I tested seem normal.

derrickstaten:
:+1: really nasty diff haha. I trust you here. The behavior was as expected

PR #137, opened 2015-06-25
Test everything

shortstuffsushi:
@qinweigong 

This adds tests around all of the request classes. It also adds back constants for pretty much everything within the requests.


Branch-SDK/Branch-SDK/BranchCreditHistoryRequest.m
```diff
@@ -63,11 +64,11 @@ - (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
     }
     
     for (NSMutableDictionary *transaction in response.data) {
-        if (transaction[@"referrer"] == [NSNull null]) {
-            [transaction removeObjectForKey:@"referrer"];
+        if ([transaction[BRANCH_RESPONSE_KEY_REFERRER] isEqual:[NSNull null]]) {
```
qinweigong: Agreed. isEqual is safer than ==, regardless the fact that [NSNull null] is singleton.

Branch-SDK/Branch-SDK/BranchShortUrlRequest.m
```diff
@@ -62,7 +63,7 @@ - (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
         if (self.callback) {
             NSString *failedUrl = nil;
             NSString *userUrl = [BNCPreferenceHelper preferenceHelper].userUrl;
-            if (!userUrl) {
+            if (userUrl) {
```
qinweigong: whoa, so it used to pass nil to the createLongUrlForUserUrl function without any problem?

qinweigong:
I assume most of the changes are 1) replacing strings with constants, 2) moving tests around, so didn't verify each of them one by one.
Looks good to me, and the if(!userUrl) -> if(userUrl) seems a legit fix.
:+1:

Branch-SDK/Branch-SDK/BranchShortUrlRequest.m
shortstuffsushi: Yeah, this has been broken for some time.

PR #139, opened 2015-06-27
On successful logout, the link cache should be cleared.

### LINKED ISSUE #138: Clear BNCLinkCache on logout
Running into a problem where a user logs out and another user then logs in. We go to generate a referral URL for the new user and it returns the referral URL from the BNCLinkCache for the previous user. Is this by design? Should the link cache be cleared when a user logs out?

shortstuffsushi:
@qinweigong

Fixes #138


qinweigong:
:+1:

PR #140, opened 2015-06-28
Delegate for activity item

shortstuffsushi:
@qinweigong @francisjervis

This pull request adds the ability for devs to change the items used to create the short url based on the channel the user chooses.

Additionally, this allows the dev to override the "human readable channel" that Branch automatically uses (for instance, you could use "fb" instead of "facebook" or something like that).

It also re-organizes the Branch interface so the ordering is more logical.

```
- (id)foo;
- (id)foo:bar;
- (id)foo:bar:baz;
```

Instead of

```
- (id)foo;
- (id)foo:baz:bar;
- (id)foo:bar;
```

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
@@ -443,7 +443,7 @@ NSString * const BRANCH_PREFS_KEY_UNIQUE_BASE = @"bnc_unique_base_";
 
 - (void)persistPrefsToDisk {
     dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-        if (![NSKeyedArchiver archiveRootObject:self.persistenceDict toFile:[self prefsFile]]) {
+        if (![NSKeyedArchiver archiveRootObject:[self.persistenceDict copy] toFile:[self prefsFile]]) {
             NSLog(@"[Branch Warning] Failed to persist preferences to disk");
         }
     });
--- a/Branch-SDK/Branch-SDK/Branch.h
+++ b/Branch-SDK/Branch-SDK/Branch.h
@@ -163,13 +163,14 @@ typedef NS_ENUM(NSUInteger, BranchPromoCodeUsageType) {
 
 /**
  Create a BranchActivityItemProvider which subclasses the `UIActivityItemProvider` This can be used for simple sharing via a `UIActivityViewController`.
- 
+
  Internally, this will create a short Branch Url that will be attached to the shared content.
- 
+
  @param params A dictionary to use while building up the Branch link.
  @param feature The feature the generated link will be associated with.
- */
-+ (BranchActivityItemProvider *)getBranchActivityItemWithParams:(NSDictionary *)params andFeature:(NSString *)feature;
+*/
++ (BranchActivityItemProvider *)getBranchActivityItemWithParams:(NSDictionary *)params feature:(NSString *)feature;
++ (BranchActivityItemProvider *)getBranchActivityItemWithParams:(NSDictionary *)params andFeature:(NSString *)feature  __attribute__((deprecated(("Use methods without 'and'"))));
 
 /**
  Create a BranchActivityItemProvider which subclasses the `UIActivityItemProvider` This can be used for simple sharing via a `UIActivityViewController`.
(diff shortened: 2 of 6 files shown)


qinweigong:
:+1: need some documentation in readme?

francisjervis:
Since a major reason for implementing this is Twitter integration, docs should either specify how to conform to Twitter object protocol (to enable image display in timeline, for example) or (better) $og_image tags (etc) should be automatically re-labelled media_url (etc) prior to posting if the channel is Twitter.
Reference:
https://dev.twitter.com/overview/api/entities-in-twitter-objects

shortstuffsushi:
Hey Francis,

I'm not sure that the SDK should be "smart" enough to try to change your content for you (it sounds useful, but supporting all the platforms would be a lot of effort), but it could certainly be added to documentation.

We [already list OG Tags here](https://dev.branch.io/recipes/dynamic_link_creation/#parameters-1), perhaps we could also mention the link you've provided to the Twitter cards as well. Again, this sort of becomes a mess if we try to have all the constants for all the social media platform, but perhaps just providing links to common platform link docs would be useful.

francisjervis:
> On Jun 29, 2015, at 11:24 AM, Graham Mueller <notifications@github.com> wrote:
> 
> Hey Francis,
> 
> I'm not sure that the SDK should be "smart" enough to try to change your content for you (it sounds useful, but supporting all the platforms would be a lot of effort), but it could certainly be added to documentation.
> 
> 
Hi Graham,

Posting content with an App Card is at least as cool as the FB invites feature - essentially turns every Twitter share into an ad with an install button on it!!  App Cards don’t link to content across install, so definitely an added value proposition using Branch for this too ;) 

There’s something else going on here though, have a look at the link below in the validator and then look at the data with it on your end (tl;dr: Twitter thinks it’s a “summary with large image" card, whereas params have “product” tag). Basically looks the same with or without all the “twitter” tags added. Adding $ before tags doesn’t fix this (and tried replacing : with _ as in the OG tag format). 

Does Branch return all the data with a link as meta tags + content, or only certain items? There are 14 data items on the link in question, so it seems not everything is being passed on. Twitter validator reports 11 meta tags (but not their contents, except for one being the wrong card tag!)

http://tfr.link/l/6-ZJtzk04o


> We already list OG Tags here <https://dev.branch.io/recipes/dynamic_link_creation/#parameters-1>, perhaps we could also mention the link you've provided to the Twitter cards as well. Again, this sort of becomes a mess if we try to have all the constants for all the social media platform, but perhaps just providing links to common platform link docs would be useful.
> 
> 

Not sure if you just updated the docs with the Twitter card link today (just noticed it now!) - hopefully won’t be hard to figure out how to pass the tags on correctly.

Also, the thing I showed Mike here in NY: the redirect for Twitter is a bit off, which I suspect is related (see pics). If it’s not tags, this is may be how your server detects the browser within the Twitter iOS app? Desktop from Twitter is fine, redirect from FB to browser is fine, etc.

F

shortstuffsushi:
I didn't mean to put Twitter Cards down or anything, more that we can't just have logic in place for every possible network (we don't do anything automatic for Facebook, we just tell you the list of constants you need).

As far as the actual card functionality, we do provide some of that in our service for you. If you look at the Text-Me-The-App feature provided by Branch (which is basically the default page we send back to scrapers), we insert a whole bunch of tags for Facebook and Twitter already. Check out [this example on Facebook](https://developers.facebook.com/tools/debug/og/echo?q=http%3A%2F%2Ftfr.link%2Fl%2F6-ZJtzk04o) (using your link). Note the og:tags and twitter:tags at the top.

I'm not sure if the validator is currently broken or if it's just my crappy internet, but it keeps telling me that it can't load your link -- `ERROR: Fetching the page failed because the fetcher cannot resolve the address.` If we're doing something wrong here, I can try to look into it more, but I need their service to come back up first.

I did actually just update the docs to include that card reference. I know it looks a bit less important on there than the large Facebook description, but it's just our primary use case. The largest percentage of our users who want social sharing are Facebook first, so we put the emphasis on that.

Your last issue with the redirects, is that the same as the above, or separate? What is "off" about it?

francisjervis:
Totally get why you don’t want this to get out of control ;)

> On Jun 30, 2015, at 11:37 AM, Graham Mueller <notifications@github.com> wrote:
> 
> I didn't mean to put Twitter Cards down or anything, more that we can't just have logic in place for every possible network (we don't do anything automatic for Facebook, we just tell you the list of constants you need).
> 
> As far as the actual card functionality, we do provide some of that in our service for you. If you look at the Text-Me-The-App feature provided by Branch (which is basically the default page we send back to scrapers), we insert a whole bunch of tags for Facebook and Twitter already. Check out this example on Facebook <https://developers.facebook.com/tools/debug/og/echo?q=http%3A%2F%2Ftfr.link%2Fl%2F6-ZJtzk04o> (using your link). Note the og:tags and twitter:tags at the top.
> 
> 
I thought so!! Thanks for the pointer, hadn’t looked at the raw output for my link. I think I’m right about the server inserting a card: tag, compare this output and shot of the branch dashboard. Other Twitter tags aren’t coming through at all (creator, site, etc). This is for
http://tfr.link/l/6-ZJtzk04o <http://tfr.link/l/6-ZJtzk04o> 
re-scraped just now.

https://developers.facebook.com/tools/debug/og/echo?q=http%3A%2F%2Ftfr.link%2Fl%2F6-ZJtzk04o <https://developers.facebook.com/tools/debug/og/echo?q=http://tfr.link/l/6-ZJtzk04o>

francisjervis:
More readable pic, bloody Mail.app ;)
F

shortstuffsushi:
Just FYI, if you're responding via the Mail.app, and adding attachments, they aren't coming through.

PR #144, opened 2015-07-13
Fixing bad key used for object removal.

shortstuffsushi:
@aaustin 

Fixed this once before, but must have been lost in a merge. This is what actually caused the bug in the Invite SDK. I'll probably go through and revert those checks in that SDK, as they will be redundant.


aaustin:
:+1:

PR #145, opened 2015-07-13
Identity Type Mixup

shortstuffsushi:
@aaustin 
Because the JSON parser is naive to what types it's getting back, it seems there are cases where they identity can become a number.
To avoid this, we are now checking on both the set/get sides to ensure the type we expect comes through.


aaustin:
:+1:

PR #146, opened 2015-07-16
Remove App List

shortstuffsushi:
@aaustin @qinweigong 

Removing AppList methods and requests, as this sort of thing is (officially) not allowed on iOS 9.
Updating tests to remove all app-list disabling.


aaustin:
Ahh. Deleting code feels so good. :+1:

PR #147, opened 2015-07-17
Updating Install Param Set Logic

shortstuffsushi:
@aaustin @derrickstaten 

Updating scenarios around when installParams are set, following these rules:

* If isReferrable is false, don't set, period.
* Otherwise, if isReferrable and
** Install: set to whatever we get.
** Open and installParams set: don't set.
** Open and not installParams and isReferrable: set if not null.

Adding tests around scenarios to validate them.
Updating a couple of other tests to compile for Xcode 7.
Removing some project cruft that has been building up from Pods, I think.


Branch-SDK/Branch-SDK/BranchOpenRequest.m
```diff
@@ -98,16 +98,18 @@ - (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
...
+    // Scenarios:
+    // If isReferrable is false, don't set, period.
+    // Otherwise, if isReferrable and
+    // * Install: set to whatever we get.
+    // * Open and installParams set: don't set.
+    // * Open and not installParams and isReferrable: set if not null.
     if (preferenceHelper.isReferrable) {
-        // If present, set it.
-        if (sessionData) {
+        BOOL storedParamsAreEmptyAndRequestValueIsNonNull = !preferenceHelper.installParams.length && sessionData.length;
```
aaustin: we don't want to set if +clicked_branch_link = 0

Branch-SDK/Branch-SDK/BranchOpenRequest.m
```diff
@@ -98,16 +98,16 @@ - (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
...
-        else if (self.allowInstallParamsToBeCleared) {
-            preferenceHelper.installParams = nil;
-        }
+    // Scenarios:
+    // If isReferrable is false, don't set, period.
+    // Otherwise, if isReferrable, is from a link click, and
+    // * Install: set to whatever we get.
+    // * Open and installParams set: don't set.
+    // * Open and not installParams and isReferrable: set if not null.
+    BOOL dataIsFromALinkClick = [sessionData rangeOfString:@"\"+clicked_branch_link\":0"].location == NSNotFound;
```
aaustin: Can we be guaranteed that there won't be spaces or anything between ":0? Why not just parse the string and look for the specific key?

aaustin:
:+1:

PR #148, opened 2015-07-20
iOS 9 Updates

shortstuffsushi:
@aaustin 

This isn't quite done, as it will require some backend work as well (faking a link click like Facebook Ads or something), but you can test everything *except* that with this.


Branch.podspec
```diff
@@ -23,5 +23,5 @@ Use the Branch SDK (branch.io) to create and power the links that point back to
   s.requires_arc = true
 
   s.source_files = "Branch-SDK/Branch-SDK/*.{h,m}"
-  s.frameworks = 'AdSupport', 'CoreTelephony'
+  s.frameworks = 'AdSupport', 'CoreTelephony', 'CoreSpotlight', 'MobileCoreServices'
```
aaustin: What happens if someone drops the framework in and leaves off these frameworks - will there be build errors?

Branch-SDK/Branch-SDK/ContentDiscovery/BNCContentDiscoveryManager.m
```diff
@@ -0,0 +1,137 @@
...
+            self.currentUserActivity.eligibleForSearch = YES;
+            self.currentUserActivity.eligibleForPublicIndexing = publiclyIndexable;
+            self.currentUserActivity.contentAttributeSet = attributes;
+            self.currentUserActivity.userInfo = userInfo; // As of iOS 9 Beta 3, this gets lost and never makes it through to application:continueActivity:restorationHandler:
+            self.currentUserActivity.requiredUserInfoKeys = [NSSet setWithArray:userInfo.allKeys]; // This, however, seems to force the userInfo to come through.
+            self.currentUserActivity.keywords = keywords;
+            [self.currentUserActivity becomeCurrent];
+            
+            // Index via the CoreSpotlight strategy
+            CSSearchableItem *item = [[CSSearchableItem alloc] initWithUniqueIdentifier:identifier domainIdentifier:spotlightIdentifier attributeSet:attributes];
```
aaustin: does this do both local and cloud search?

Branch-SDK/Branch-SDK/Core/Branch.m
```diff
@@ -354,6 +357,16 @@ - (BOOL)handleDeepLink:(NSURL *)url {
     return handled;
 }
 
+- (BOOL)continueUserActivity:(NSUserActivity *)userActivity {
+    NSString *linkIdentifier = [self.contentDiscoveryManager spotlightLinkIdentifierFromActivity:userActivity];
```
aaustin: don't we need to set this in the preference helper so that it's sent back on /v1/install or /v1/open?

shortstuffsushi: Yes, this is pending the backend change. I thought I had put a TODO in here specifically stating as much, but apparently I forgot.

Branch-SDK/Branch-SDK/ContentDiscovery/BNCContentDiscoveryManager.m
shortstuffsushi: Yes, the CoreSpotlight stuff is local, NSUserActivity is (potentially) public.

Branch.podspec
shortstuffsushi: I'm not sure yet what to do with this. When I run against non-beta Xcode, this doesn't seem to have any effect. I would need to review this before merging, though.

Branch-SDK/Branch-SDK/Core/Branch.m
aaustin: Cool - looks good. Let me know when it's ready for merge.

Branch-SDK/Branch-SDK/ContentDiscovery/BNCContentDiscoveryManager.m
```diff
@@ -0,0 +1,138 @@
...
+- (void)indexContentWithTitle:(NSString *)title description:(NSString *)description publiclyIndexable:(BOOL)publiclyIndexable type:(NSString *)type thumbnailUrl:(NSURL *)thumbnailUrl keywords:(NSSet *)keywords userInfo:(NSDictionary *)userInfo callback:(callbackWithUrl)callback {
+    if ([BNCSystemObserver getOSVersion].integerValue < 9) {
+        if (callback) {
+            callback(nil, [NSError errorWithDomain:BNCErrorDomain code:BNCVersionError userInfo:@{ NSLocalizedDescriptionKey: @"Cannot use CoreSpotlight indexing service prior to iOS 9" }]);
+        }
+        return;
+    }
+
+    #if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 90000
+        [[Branch getInstance] getSpotlightUrlWithParams:userInfo callback:^(NSDictionary *data, NSError *urlError) {
```
derrickstaten: Is there a reason we don't pass up title, description, etc? Perhaps as "+title", "+description", etc. Or perhaps I'm overlooking something.

derrickstaten:
- [ ] with spotlight, download NSURLs starting with `http` synchronously and then have Apple index with that thumbnail present

shortstuffsushi:
@derrickstaten you can explain this?
> with spotlight, download NSURLs starting with http synchronously and then have Apple index with that thumbnail present
That seems like a bad idea. I mean, indexing an image is great, but sync download is less so.

Branch-SDK/Branch-SDK/Core/Branch.m
```diff
@@ -346,6 +351,26 @@ - (BOOL)handleDeepLink:(NSURL *)url {
     return handled;
 }
 
+- (BOOL)continueUserActivity:(NSUserActivity *)userActivity {
+    if ([userActivity.activityType isEqualToString:NSUserActivityTypeBrowsingWeb]) {
+        self.preferenceHelper.universalLinkUrl = [userActivity.webpageURL absoluteString];
```
aaustin: Seems like we're setting any page URL to a Branch universal link URL. It's likely a company will use our ULinks in addition to theirs. Is this going to break something?

Branch-SDK/Branch-SDK/Core/Branch.m
```diff
@@ -346,6 +351,26 @@ - (BOOL)handleDeepLink:(NSURL *)url {
...
+
+        [self initUserSessionAndCallCallback:YES];
+        
+        return YES;
+    }
+    
+    NSString *spotlightIdentifier = [self.contentDiscoveryManager spotlightIdentifierFromActivity:userActivity];
+    
+    if (spotlightIdentifier) {
+        self.preferenceHelper.spotlightIdentifier = spotlightIdentifier;
```
aaustin: same thing here.

aaustin: nevermind.. that above method strips it

aaustin:
I think the server will handle it. Let's merge and do end to end testing :+1:

derrickstaten:
:+1:

derrickstaten:
I think we should close without merging, leave the branch, and point people to the Branch. Don't delete the Branch! 

Then when iOS 9 is live we can merge the branch.

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -346,6 +351,26 @@ - (BOOL)handleDeepLink:(NSURL *)url {
     return handled;
 }
 
+- (BOOL)continueUserActivity:(NSUserActivity *)userActivity {
+    if ([userActivity.activityType isEqualToString:NSUserActivityTypeBrowsingWeb]) {
+        self.preferenceHelper.universalLinkUrl = [userActivity.webpageURL absoluteString];
```
derrickstaten: so this looks like we will try to handle any universal link!! even non-Branch ones. With spotlight, at least we look for the branch prefix.

derrickstaten: crudely handled now by returning NO if `bnc.lt` isn't present in the `webpageURL`

shortstuffsushi: This won't work for custom subdomains (which I think you're already probably aware of). I think the backend may always generate these urls with just bnc.lt, though, so it might work.

derrickstaten: yeah, you're right. we still need to figure out what we're going to do here, we've been discussing alternatives today. We'll probably address next week.

derrickstaten:
here goes :+1:

PR #149, opened 2015-07-22
Update README.md

pgherveou:
fixed swift example
I think thats the correct syntax


shortstuffsushi:
Both signatures are valid, though the old way is deprecated. This appears to be correct from local testing. Thanks for the addition.

PR #150, opened 2015-07-24
Use Time Intervals

shortstuffsushi:
@derrickstaten 

Since there is a desire to use non-whole second timeouts, just changed to using NSTimeIntervals.
TimeIntervals are typedef'd as doubles, so they allow non-whole numbers.
Since this is the type the NSURLRequest is expecting anyway, it's a pretty easy change to make.


qinweigong:
:+1:

PR #151, opened 2015-07-28
Urgent: Add back in update app list

aaustin:
Crashes are being reported when the update app list server request was
persisted pre-removal, then being retrieved post-removal. We need to
leave this in for a while so that we can retrieve the queue from disk.

@qinweigong

### DIFF
--- a/Branch-SDK/Branch-SDK/BranchUpdateAppListRequest.h
+++ b/Branch-SDK/Branch-SDK/BranchUpdateAppListRequest.h
@@ -1,0 +1,15 @@
+//
+//  BranchUpdateAppListRequest.h
+//  Branch-TestBed
+//
+//  Created by Graham Mueller on 5/26/15.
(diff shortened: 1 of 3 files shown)


qinweigong:
Shall we add back the .m file too?

shortstuffsushi:
You would need to add back the .m file as @qinweigong mentioned, and also add the GetList request (you've only added the UpdateList request). It might be better to try to find a way to gracefully degrade in the case that a class is missing, so this works if we miss one in the future.

qinweigong:
Minor thing: wanna move the two files to inside the BranchRequests folder? Otherwise, looks good to me :+1:

aaustin:
There's no folder in my local version - is there one in yours?

qinweigong:
Ah, it's not a folder in Mac file system, but a folder in Xcode project navigator.

PR #152, opened 2015-07-28
Urgent: just notified that get applist is persisted too

aaustin:
@qinweigong Sorry but can you help me merge this one too? I should have
read his code more closely..


qinweigong:
:+1:

PR #153, opened 2015-08-03
Request queue error handling

shortstuffsushi:
@aaustin @qinweigong 

Adding some handling to the queue to handle request retrieve/persist failures.


Branch-SDK/Branch-SDK/BranchGetAppListRequest.h
```diff
@@ -1,16 +0,0 @@
-//
```
qinweigong: we can get rid of BranchGetAppListRequest and BranchUpdateAppListRequest now? Just want to make sure this is not an accidental removal.

shortstuffsushi: The majority of the idea w/ this PR was to add error handling, which in turn allows us to safely remove requests w/o fear of pending requests in the queue biting us. 

However, this made me realize that I'm wrapping the *entire* block in a try -- it might be better to have a try around *each* so that one request being discarded doesn't result in the whole block of requests being thrown out.

qinweigong: Ah, i was talking about the removal of the two classes (.h & .m files) that we removed, caused crashes, and i added back last week. Just want to make sure it's safe to finally remove them now?

shortstuffsushi: Right, the problem was that they were failing to be parsed by the `retrieve` function, so now there's a try/catch wrapped around that, which ensures that if the request is of an unknown type, it doesn't get added to the queue (and doesn't cause the app to crash). Try it out locally -- run on master, get the request enqueued, then switch to this branch -- should work :smiley:

qinweigong: Oh, i see! And having the "side effect" of catch all other wild exceptions :-)

qinweigong:
:+1:

PR #154, opened 2015-08-03
Conditionally defining constants for iOS 6.

shortstuffsushi:
@qinweigong @aaustin 
Not sure how this was working in the past, but apparently simply putting the constants inside of a version check isn't enough -- they cause a crash because they aren't defined at all for iOS 6.
This also made it apparent that we're doing an == check on strings, updated accordingly.

### DIFF
--- a/Branch-SDK/Branch-SDK/BranchActivityItemProvider.m
+++ b/Branch-SDK/Branch-SDK/BranchActivityItemProvider.m
@@ -10,6 +10,16 @@
 #import "Branch.h"
 #import "BNCSystemObserver.h"
 
+// Define constants so that iOS 6 won't crash. Don't care about their values since you can't
+// possible access them, just necessary for the sake of not accessing missing symbols.
+#ifndef UIActivityTypeAddToReadingList
+#define UIActivityTypeAddToReadingList @""
+#define UIActivityTypePostToFlickr @""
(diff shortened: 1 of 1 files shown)


Branch-SDK/Branch-SDK/BranchActivityItemProvider.m
```diff
@@ -10,6 +10,16 @@
 #import "Branch.h"
 #import "BNCSystemObserver.h"
 
+// Define constants so that iOS 6 won't crash. Don't care about their values since you can't
+// possible access them, just necessary for the sake of not accessing missing symbols.
+#ifndef UIActivityTypeAddToReadingList
+#define UIActivityTypeAddToReadingList @""
```
qinweigong: do we want to do #ifndef for each of them individually?

qinweigong:
:+1:

Branch-SDK/Branch-SDK/BranchActivityItemProvider.m
shortstuffsushi: I'm using the absence of `UIActivityTypeAddToReadingList` as a marker. It's lazy, but I don't think we're going to see a scenario where just one of those is missing, since those 5 are all added with iOS 7.

PR #156, opened 2015-08-04
Remove Debug Requests from Persistence

shortstuffsushi:
@qinweigong @Sarkar 
Persist the log when for log requests, so they can be restored (potentially causing nil in dictionary).


Sarkar:
:+1:

qinweigong:
looks good :+1:

PR #158, opened 2015-08-05
fix an encoding bug when json object contains non-ascii characters

### LINKED ISSUE #157: Wrong string length used when converting JSON dict to NSData
https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK/blob/master/Branch-SDK/Branch-SDK/BNCEncodingUtils.m#L200

In method `+ (NSData *)encodeDictionaryToJsonData:(NSDictionary *)dictionary`, you used `jsonString.length` instead of length of bytes. As a result, if the dictionary contains non-ascii characters, e.g. Asian characters (which happens mostly in `carrier` field), `jsonString.length` < length of bytes, so the result `NSData` is shorter than jsonString, which result in server returning:
```
Status: 400; Data: {
    error =     {
        code = 400;
        message = "Invalid JSON";
    };
}
```

Solution is quite strait forward:
```
+ (NSData *)encodeDictionaryToJsonData:(NSDictionary *)dictionary {
    NSString *jsonString = [BNCEncodingUtils encodeDictionaryToJsonString:dictionary];
    NSUInteger length = [jsonString lengthOfBytesUsingEncoding:NSUTF8StringEncoding];
    return [NSData dataWithBytes:[jsonString UTF8String] length:length];
}
```

Another solution is use system class `NSJSONSerialization` instead of custom converter:
```
[NSJSONSerialization dataWithJSONObject:dictionary options:0 error:&error];
```

allenhsu:
fix #157

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCEncodingUtils.m
+++ b/Branch-SDK/Branch-SDK/BNCEncodingUtils.m
@@ -199,7 +199,7 @@ static const short _base64DecodingTable[256] = {
 
 + (NSData *)encodeDictionaryToJsonData:(NSDictionary *)dictionary {
     NSString *jsonString = [BNCEncodingUtils encodeDictionaryToJsonString:dictionary];
-    return [NSData dataWithBytes:[jsonString UTF8String] length:jsonString.length];
+    return [NSData dataWithBytes:jsonString.UTF8String length:strlen(jsonString.UTF8String)];
 }
 
 + (NSString *)encodeDictionaryToJsonString:(NSDictionary *)dictionary {


shortstuffsushi:
Hi @allenhsu, thanks for this! Do you have any good example strings? I'd like to add tests around this scenario.

shortstuffsushi:
I actually added a branch that exhibits the issue, I'm not sure if it's possible, but could you target that branch with your pull request?

allenhsu:
OK, I'll open another pr.
For test case:
```
@{@"carrier": @"\U4e2d\U56fd\U8054\U901a"}
```

Where `@"\U4e2d\U56fd\U8054\U901a"` is UTF8 for `@"中国联通"`

shortstuffsushi:
Thanks! Feel free to add to / update the existing test case I added.

PR #159, opened 2015-08-06
Encoded data length issues

allenhsu:
Use `strlen()` instead of `length` of `NSString`

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCEncodingUtils.m
+++ b/Branch-SDK/Branch-SDK/BNCEncodingUtils.m
@@ -199,7 +199,7 @@ static const short _base64DecodingTable[256] = {
 
 + (NSData *)encodeDictionaryToJsonData:(NSDictionary *)dictionary {
     NSString *jsonString = [BNCEncodingUtils encodeDictionaryToJsonString:dictionary];
-    return [NSData dataWithBytes:[jsonString UTF8String] length:jsonString.length];
+    return [NSData dataWithBytes:jsonString.UTF8String length:strlen(jsonString.UTF8String)];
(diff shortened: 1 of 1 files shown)


Branch-SDK/Branch-SDK/BNCEncodingUtils.m
```diff
@@ -199,7 +199,7 @@ + (NSString *)sanitizedStringFromString:(NSString *)dirtyString {
 
 + (NSData *)encodeDictionaryToJsonData:(NSDictionary *)dictionary {
     NSString *jsonString = [BNCEncodingUtils encodeDictionaryToJsonString:dictionary];
-    return [NSData dataWithBytes:[jsonString UTF8String] length:jsonString.length];
+    return [NSData dataWithBytes:jsonString.UTF8String length:strlen(jsonString.UTF8String)];
```
shortstuffsushi: For style preference, could you use the other method you described in #157? Specifically, I prefer not using dot notation for method calls, and here you make the same method call (`UTF8String`) twice.

allenhsu: OK, I'll do that. But in fact, `UTF8String` is a `property` of `NSString` pointing to a structure inside the string object, and there's no overhead calling twice.

allenhsu:
@shortstuffsushi updated

Branch-SDK/Branch-SDK/BNCEncodingUtils.m
shortstuffsushi: So it is, my mistake!

PR #160, opened 2015-08-08
Change isReferrable default

aaustin:
There has been too much confusion around setting isReferrable only on
fresh installs.

@shortstuffsushi

### DIFF
--- a/Branch-SDK/Branch-SDK/Branch.m
+++ b/Branch-SDK/Branch-SDK/Branch.m
@@ -248,18 +248,15 @@ static int BNCDebugTriggerFingersSimulator = 2;
 #pragma mark - InitSession Permutation methods
 
 - (void)initSession {
-    BOOL isReferrable = [[BNCSystemObserver getUpdateState] isEqualToNumber:@0] && !self.preferenceHelper.sessionID;
-    [self initSessionWithLaunchOptions:nil isReferrable:isReferrable explicitlyRequestedReferrable:NO automaticallyDisplayController:NO registerDeepLinkHandler:nil];
+    [self initSessionWithLaunchOptions:nil isReferrable:YES explicitlyRequestedReferrable:NO automaticallyDisplayController:NO registerDeepLinkHandler:nil];
 }
 
 - (void)initSessionAndAutomaticallyDisplayDeepLinkController:(BOOL)automaticallyDisplayController {
-    BOOL isReferrable = [[BNCSystemObserver getUpdateState] isEqualToNumber:@0] && !self.preferenceHelper.sessionID;
-    [self initSessionWithLaunchOptions:nil isReferrable:isReferrable explicitlyRequestedReferrable:NO automaticallyDisplayController:automaticallyDisplayController registerDeepLinkHandler:nil];
+    [self initSessionWithLaunchOptions:nil isReferrable:YES explicitlyRequestedReferrable:NO automaticallyDisplayController:automaticallyDisplayController registerDeepLinkHandler:nil];
 }
 
 - (void)initSessionWithLaunchOptions:(NSDictionary *)options {
-    BOOL isReferrable = [[BNCSystemObserver getUpdateState] isEqualToNumber:@0] && !self.preferenceHelper.sessionID;
-    [self initSessionWithLaunchOptions:options isReferrable:isReferrable explicitlyRequestedReferrable:NO automaticallyDisplayController:NO registerDeepLinkHandler:nil];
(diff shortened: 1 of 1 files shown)


shortstuffsushi:
I don't really understand this change, could you give a little more context? If we're going to do this on both iOS and Android, why don't we just start ignoring it server side again?

aaustin:
Well, it's still an argument that's optional for the developer? How is ignoring it equivalent to this?

shortstuffsushi:
Were we ignoring developer requested referable in the past? I suppose we probably were, since that was a blanket "ignore anything below xxx version." So, I guess in that sense, it wouldn't be the same if we moved back to ignoring on the server. 

However, I still don't understand the motivation for this. We've had [a](https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK/commit/cf447d3eb5dc349a10a3212aa86c8924f38a630b) [number](https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK/commit/e1477321ff4e172ebe35888dd73c454bb202f592) [of](https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK/commit/15524b7b6ac7107dc3bc91037e2c3fa54ad694fb) [updates](https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK/commit/31245fc3f663ac40f85c8508b5c3daf7366ee49d) [around](https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK/commit/681d2362aa062ff459d0a4c4f403ebe9719f5caa) this behavior, but I was under the impression we were fairly confident about it at this point. We even have [tests around it](https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK/blob/6de5536c1cbbd13bfce11baa7b7d4ee981dc2e5a/Branch-TestBed/Branch-SDK-Tests/BNCSystemObserverTests.m#L26-L118). Can you explain more about what's confusing people here, and why we want to remove this behavior as a result?

aaustin:
We made a product decision early in the SDK days to only allow Branch users to be eligible for referral if they were genuine fresh installs. This is what felt like the right thing based on when we were working on Kindred. Unfortunately, this is what we refer to as 'magic' - undocumented and unexpected functionality that happens without notifying the user. It causes things like getFirstReferring params to be empty when a user certainly clicked on a link. It has continued to generate a significant number of support requests, where people don't understand why referrals aren't being counted for existing users or params are empty. We then have to explain the design decision. 

Since we've been ignoring it for older versions and continue to have to explain it's functionality to new integrators, it's safe to assume that this default magic was a poor product decision that should be reversed.

aaustin:
I didn't know there were tests - does CircleCI ignore tests right now or something?

shortstuffsushi:
I'm not sure that this is the right solution to the problem you're describing, but if you'd rather do it this way, we can merge it in. I just think it's likely that now you're going to start hearing from the people relying on this behavior suddenly seeing unexpected spikes in installs. We'll see, though.

The tests are not ignored, they're run against all branches, but they aren't integration tests, so this particular change isn't captured. The tests are around the `getUpdateState` behavior (it's linked above), which could probably be removed if we're moving to this behavior.

aaustin:
Ah - this doesn't change the getUpdateState at all. That will still be used determine whether it's an install | open. This change just reflects isReferrable.

Do you have another idea? I want to change the product functionality and I don't know how else this could be done.

shortstuffsushi:
We don't use update state to determine whether we send install or open, we're basing that purely on presence of an `identity_id`.

Aside from `isReferrable`, the update state is passed into the open/install request bodies, and that's the extent of its use, I think. The backend seems to use that value for determining `+is_first_session`, so if that's the functionality that our users are confused by, this won't fix that (as a side note).

The open/install response handling also will still only set the `firstReferringParams` if there is no `identity_id` (and no referring params already). Will this, or that backend `+is_first_session` still result in confused users?

aaustin:
The update state is used to determine if a 'install' or 'open' event should be saved in the database on that call the initSession. We definitely need it. That is not the functionality that users are confused by....

The functionality users are confused by is that we only pass is referrable true for fresh installs on the client side. Please see 4 comments up for explanation...... I want to reverse that product decision. Since we've been ignoring it for older versions and continue to have to explain it's functionality to new integrators, it's safe to assume that this change will not impact anyone.

shortstuffsushi:
OK. I'm not entirely confident in this change, but I'll take your word for it :+1:

PR #161, opened 2015-08-14
Codeless Install

shortstuffsushi:
@aaustin @qinweigong @Sarkar 

Tested this w/ Cocoapods and w/ the Framework file. This makes it so all that you have to do is include / load the Branch symbols, and it will automatically complete the rest of the process -- no initSession required. You do still need to add your branch_key, however.


Sarkar:
:+1: 

@aaustin for context, this allows someone to drag and drop our framework into their XCode project, hit run, and automatically send us a request ("ping").

qinweigong:
Interesting!

shortstuffsushi:
@wlindemann Please consider updating the Onboarding or Docs with this update. The process is basically identical to the [Apptimize](http://apptimize.com/docs/installation/) now, since they inspired this addition. The README is updated with the steps, but feel free to reach out if you need any additional info.

PR #163, opened 2015-08-31
Short URI Scheme Fix

shortstuffsushi:
@derrickstaten 

Fix scenario where uri scheme is < 3 characters length causing crash.
Add test around this scenario.


derrickstaten:
:+1:

PR #164, opened 2015-09-08
handle 400 errors

derrickstaten:
@aaustin


aaustin:
hmm - why are there 400 errors?

aaustin:
:+1:

PR #167, opened 2015-09-11
stop using deprecated methods

derrickstaten:
@aaustin @Sarkar this is my initial pass. I haven't tested this extensively. I'm going to do so now. Please look for any potential issues.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCEncodingUtils.m
+++ b/Branch-SDK/Branch-SDK/BNCEncodingUtils.m
@@ -330,7 +330,7 @@ static const short _base64DecodingTable[256] = {
 }
 
 + (NSString *)urlEncodedString:(NSString *)string {
-    return (NSString *)CFBridgingRelease(CFURLCreateStringByAddingPercentEscapes(NULL, (CFStringRef)string, NULL, CFSTR("!*'\"();:@&=+$,/?%#[]% "), CFStringConvertNSStringEncodingToEncoding(NSUTF8StringEncoding)));
+    return [string stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet characterSetWithCharactersInString:@"!*'\"();:@&=+$,/?%#[]% "]];
 }
 
 + (NSString *)encodeDictionaryToQueryString:(NSDictionary *)dictionary {
@@ -413,7 +413,7 @@ static const short _base64DecodingTable[256] = {
         NSArray *kv = [pair componentsSeparatedByString:@"="];
         if (kv.count > 1) { // If this key has a value (so, not foo&bar=...)
             NSString *key = kv[0];
-            NSString *val = [kv[1] stringByReplacingPercentEscapesUsingEncoding:NSUTF8StringEncoding];
+            NSString *val = [kv[1] stringByRemovingPercentEncoding]; // uses the default UTF-8 encoding
             
             // Don't add empty items
             if (val.length) {
--- a/Branch-SDK/Branch-SDK/BNCEncodingUtils.m
+++ b/Branch-SDK/Branch-SDK/BNCEncodingUtils.m
@@ -330,7 +330,7 @@ static const short _base64DecodingTable[256] = {
 }
 
 + (NSString *)urlEncodedString:(NSString *)string {
-    return [string stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet characterSetWithCharactersInString:@"!*'\"();:@&=+$,/?%#[]% "]];
+    return [string stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLQueryAllowedCharacterSet]];
 }
 
 + (NSString *)encodeDictionaryToQueryString:(NSDictionary *)dictionary {
--- a/Branch-SDK/Branch-SDK/BNCEncodingUtils.m
+++ b/Branch-SDK/Branch-SDK/BNCEncodingUtils.m
@@ -330,7 +330,9 @@ static const short _base64DecodingTable[256] = {
 }
 
 + (NSString *)urlEncodedString:(NSString *)string {
-    return [string stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLQueryAllowedCharacterSet]];
+    NSMutableCharacterSet *charSet = [[NSCharacterSet URLQueryAllowedCharacterSet] mutableCopy];
+    [charSet removeCharactersInString:@"!*'\"();:@&=+$,/?%#[]% "];
+    return [string stringByAddingPercentEncodingWithAllowedCharacters:charSet];
 }
 
 + (NSString *)encodeDictionaryToQueryString:(NSDictionary *)dictionary {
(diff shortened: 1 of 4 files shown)


derrickstaten:
got a crash when simulating connectivity problems, so this is definitely not ready to go: 

```
2015-09-11 11:04:34.113 Branch-TestBed[55894:13849085] This application is modifying the autolayout engine from a background thread, which can lead to engine corruption and weird crashes.  This will cause an exception in a future release.
 Stack:(
	0   CoreFoundation                      0x00000001083639b5 __exceptionPreprocess + 165
	1   libobjc.A.dylib                     0x0000000107ddbdeb objc_exception_throw + 48
	2   CoreFoundation                      0x00000001083638ed +[NSException raise:format:] + 205
	3   Foundation                          0x0000000107b513f5 _AssertAutolayoutOnMainThreadOnly + 79
	4   Foundation                          0x00000001079b223e -[NSISEngine withBehaviors:performModifications:] + 31
	5   UIKit                               0x0000000106de1a47 -[UIView(AdditionalLayoutSupport) _withAutomaticEngineOptimizationDisabledIfEngineExists:] + 58
	6   UIKit                               0x0000000106de258a -[UIView(AdditionalLayoutSupport) updateConstraintsIfNeeded] + 254
	7   UIKit                               0x0000000106de329f -[UIView(AdditionalLayoutSupport) _updateConstraintsAtEngineLevelIfNeeded] + 272
	8   UIKit                               0x0000000106628c69 -[UIView(Hierarchy) _updateConstraintsAsNecessaryAndApplyLayoutFromEngine] + 159
	9   UIKit                               0x0000000106637a3b -[UIView(CALayerDelegate) layoutSublayersOfLayer:] + 710
	10  QuartzCore                          0x000000010c76436a -[CALayer layoutSublayers] + 146
	11  QuartzCore                          0x000000010c758bd0 _ZN2CA5Layer16layout_if_neededEPNS_11TransactionE + 366
	12  QuartzCore                          0x000000010c758a4e _ZN2CA5Layer28layout_and_display_if_neededEPNS_11TransactionE + 24
	13  QuartzCore                          0x000000010c74d1d5 _ZN2CA7Context18commit_transactionEPNS_11TransactionE + 277
	14  QuartzCore                          0x000000010c77a9f0 _ZN2CA11Transaction6commitEv + 508
	15  QuartzCore                          0x000000010c77acd4 _ZN2CA11Transaction14release_threadEPv + 224
	16  libsystem_pthread.dylib             0x0000000109c3f739 _pthread_tsd_cleanup + 86
	17  libsystem_pthread.dylib             0x0000000109c3f460 _pthread_exit + 117
	18  libsystem_pthread.dylib             0x0000000109c3fb33 pthread_get_stackaddr_np + 0
	19  libsystem_pthread.dylib             0x0000000109c3d3dd start_wqthread + 13
)
```

shortstuffsushi:
Did you check that these work with older versions of iOS? `NSURLSession` specifically won't work with (I believe) 7.

You're also using semaphores wrong. The wait needs to be *before* the action (the request), and I believe your count should be greater than zero. Check out [Apple's docs](https://developer.apple.com/library/ios/documentation/General/Conceptual/ConcurrencyProgrammingGuide/OperationQueues/OperationQueues.html#//apple_ref/doc/uid/TP40008091-CH102-SW24)

shortstuffsushi:
Also, I'm fairly certain the data task isn't returning on the main thread. Before calling the callback, you need to dispatch to that thread.

derrickstaten:
yeah that last issue is likely the cause of the crash. You're talking about inside the first method with the task, correct? 

What do you recommend as far as backwards compatibility goes? Pre-processor macros like what we have with CoreSpotlight related functions?

derrickstaten:
I've never used semaphores before, this was a first attempt. I'll have to look at this further

derrickstaten:
from the examples I'm seeing online, this seems like a proper use of semaphores: http://stackoverflow.com/a/4326754/1914567.

shortstuffsushi:
You use Tasks in both the sync and async methods. The async one will definitely need to rejoin main thread, not sure about the sync one.

Not sure re: backwards compat. Preprocessor stuff just makes a mess of everything, I prefer to avoid it, but then you have to live with all the older methods that are eventually deprecated. It's really kind of a preference choice: preprocessor garbage everything or potentially deprecated methods. Since we're targeting the older SDK versions, though, you should see those in Xcode, I wouldn't think.

It seems I may be misunderstanding the intention of the semaphore, so your usage is correct. I missed that you were trying to fake a synchronous request with it, I was thinking more of a lock, like the queue uses.

derrickstaten:
Okay this is super helpful, thanks. Regarding deprecated methods, unfortunately I think the oldest OS I have with me is iOS 8. Hoping I can find older ones today, as I'm working remotely.

derrickstaten:
dispatching to main queue solved the issue inside the first async task. pushing.

derrickstaten:
def not backwards compatible. i may resort to preprocessor macros :/

derrickstaten:
@aaustin @Sarkar this is ready for review. I'm going to do a separate PR for BranchCSSearchableItemAttributeSet, which currently keeps the SDK from compiling against ios 8 and earlier :/

derrickstaten:
This needs extensive manual testing as it involves encoding URLs, decoding URLs, and different methods of making server requests. So far it performs the same for me, but I may be overlooking some scenarios.

derrickstaten:
encoding tests are failing. So this is definitely wrong.

aaustin:
Just tested and it seems to work well :+1:

derrickstaten:
@shortstuffsushi if you know how to solve for cocoapods, please let me know. Hoping to merge this in soon

shortstuffsushi:
@derrickstaten the Podspec still requires [those libs](https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK/blob/0d267c6970118e59942062d0c8646e03514b3f03/Branch.podspec#L25), which means it won't work for people importing that way. A simple fix would be to make those a subspec (like `s.subspec 'iOS9'` or w/e the syntax is, and just put the frameworks in there. It would make people manually require those subspecs for now, until 9 is out.

derrickstaten:
@shortstuffsushi I used the most recent cocoapod with build targets of ios 6 and ios 9 and didn't have any issues. What are the potential issues for partners with the current setup? I don't think we want to go the subspec route. I think instead we should notify partners of any potential issues in the documentation.

shortstuffsushi:
The Podspec has the libs required in it, so if people try to build with older versions of Xcode, I believe they'd see linker errors for those missing frameworks. Haven't tried to be sure, but simple test is a new Xcode project with the latest Branch (0.11.1) installed as a pod with Xcode 5 (or 6? the one before iOS 9).

derrickstaten:
@shortstuffsushi got it. We don't want to annoy partners.

shortstuffsushi:
I'm not sure what you're saying. Were you able to test that out on the previous Xcode to verify it works? I think that would be potentially more annoying for all those users who hadn't updated yet, than making new users opt in. But up to you guys now.

PR #174, opened 2015-09-14
fix universal links not returning deeplink data on cold start

derrickstaten:
@aaustin @Sarkar this fixes spotlight search / UL not working on cold start

However, I'm not sure if there are other side effects to doing this, like failing to initialize in certain scenarios. I want to test, and if you have scenarios you think I should test please let me know


derrickstaten:
This still needs work. Unfortunately it doesn't appear to work every time.

derrickstaten:
Okay, appears to work consistently. An OS-level bug prevented it from working previously. If I see this bug on newer versions of iOS 9 I'll file a bug report.

Will test a few more edge cases. @aaustin and @Sarkar this is urgent and ready for review.

aaustin:
:+1:

PR #176, opened 2015-09-15
make new initsession check safe for version prior to ios 8

derrickstaten:
@aaustin


aaustin:
Please remove ```If you are working in an older version of Xcode, you may need to use an older version of our SDK. Please try:

pod 'Branch', ‘0.10.9’``` from the documentation as well. It's not needed.

derrickstaten:
Odd, I tried all that before and I swear I got compile errors. But if it works, great. I'm gonna pull down quickly.

derrickstaten:
I'm not sure what errors I was seeing before on the testbed but I'm not seeing them on my test project :+1:

PR #177, opened 2015-09-15
no longer requires SafariServices

derrickstaten:
@aaustin @Sarkar partners will need to link against SafariServices if they want this functionality

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
@@ -24,11 +24,9 @@
 
 #else
 
-#import <SafariServices/SafariServices.h>
-
 NSInteger const ABOUT_30_DAYS_TIME_IN_SECONDS = 60 * 60 * 24 * 30;
 
-@interface BNCStrongMatchHelper () <SFSafariViewControllerDelegate>
+@interface BNCStrongMatchHelper ()
 
 @property (strong, nonatomic) UIWindow *secondWindow;
 @property (assign, nonatomic) BOOL requestInProgress;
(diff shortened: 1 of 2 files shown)


Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
```diff
@@ -97,27 +95,28 @@ - (void)presentSafariVCWithBranchKey:(NSString *)branchKey {
...
-
-- (void)safariViewController:(SFSafariViewController *)controller didCompleteInitialLoad:(BOOL)didLoadSuccessfully {
-    [self.secondWindow.rootViewController dismissViewControllerAnimated:NO completion:NULL];
-
-    if (didLoadSuccessfully) {
-        [BNCPreferenceHelper preferenceHelper].lastStrongMatchDate = [NSDate date];
+    Class SFSafariViewControllerClass = NSClassFromString(@"SFSafariViewController");
+    if (SFSafariViewControllerClass) {
+        id safController = [[SFSafariViewControllerClass alloc] initWithURL:[NSURL URLWithString:urlString]];
+
```
Sarkar: noticed we set the delegate earlier; why was it removed? should we keep it?

Sarkar: actually, since we don't conform to the protocol, probably don't need it. looks like we don't rely on any of their callbacks either.

derrickstaten: I couldn't keep it unfortunately. I had to remove line 31 above because we're not importing it explicitly:  https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK/pull/177/files#diff-02041277e39d3db3a75c1036ed54f1d7L31

What we lose is the ability to know for sure that the request went through. So if the request times out, the cookie<>IDFA pair will not be established via this mechanism. Thankfully there are other backend mechanisms, and this will try again in 30 days. So marginal loss.

Sarkar: ah, I see. that makes sense. I've tested on 8.4 and 9.0, and can confirm that it's only executed when it's needed or available ( iOS 9 >= ). :+1:

aaustin:
Won't test but code changes look good. :+1:

aaustin:
let's hold on merging for a bigger release

derrickstaten:
Agreed. No reason to release this separately

PR #183, opened 2015-09-18
remove codeless install

derrickstaten:
@aaustin please leave on branch until ready for release. Now that we've added handling for race conditions that presumably were not an issue before this, I need to go test that removing the `load` call doesn't re-break any functionality.

### DIFF
--- a/Branch-SDK/Branch-SDK/Branch.m
+++ b/Branch-SDK/Branch-SDK/Branch.m
@@ -907,11 +907,6 @@ static int BNCDebugTriggerFingersSimulator = 2;
 
 #pragma mark - Private methods
 
-+ (void)load {
-     // Force load / registration of notification listener, allowing for codeless install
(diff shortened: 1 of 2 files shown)


shortstuffsushi:
Seems like it'd be better to find a working solution than just drop this. Ideally, calling `getInstance` vs `getTestInstance` would actually get you different instances. This is something that probably should have been done all along.

derrickstaten:
I'm not sure what solution is workable. Since this occurs at load time, we need some way to know whether getInstance or getTestInstance is going to be invoked in `application:didFinishLaunchingWithOptions:`. One hacky, partial solution I came up with was to not fire this if `branch_key` is a dictionary and `test` is a key that's present. Then codeless installs would have to refrain from including a `test` key. 

For now, until we can decide on functionality, we need to remove this, as it's already generated confusion for 2 partners.

PR #185, opened 2015-09-18
remove all references to iOS 9 frameworks from podfile (Xcode 6 compa…

derrickstaten:
…tibility)
@aaustin after approval, hold on branch until release


aaustin:
:+1:

shortstuffsushi:
GM seeds of Xcode 7 are now available, probably don't need to remove these. In fact, could probably move them back into the primary Frameworks section.

derrickstaten:
Unfortunately some users are still on Xcode 6.4 which is why we're having to pull this.

PR #186, opened 2015-09-18
better view controller presentation/dismissal to not trigger warnings

derrickstaten:
@aaustin leave on branch until 0.11.5 release

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
@@ -106,13 +106,11 @@ NSInteger const ABOUT_30_DAYS_TIME_IN_SECONDS = 60 * 60 * 24 * 30;
         [self.secondWindow makeKeyAndVisible];
         [self.secondWindow setAlpha:0];
         
-        [windowRootController presentViewController:safController animated:NO completion:NULL];
-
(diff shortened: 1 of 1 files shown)


aaustin:
:+1:

shortstuffsushi:
Isn't this going to make the race conditions for 100% match almost certain to fail? If it waits 5 seconds before launching, `initSession` will have either finished or timed out.

derrickstaten:
That's not what is happening here. This is dealing with dismissal.

It used to wait 5 seconds to dismiss the invisible view controller, now it dismisses it as soon as the controller finishes presenting. 

However, that raises another issue -- in poor connectivity scenarios, does this prevent retries that would have otherwise happened if this invisible view controller were kept open for up to 5 seconds? I think that's something we can worry about later. The important change here is to stop presenting with `animated:NO` as that triggers a warning that end users see.

shortstuffsushi:
Ah, this is *removing* that timeout. My mistake. Previously, five seconds after presenting, it would dismiss. Now it's doing so immediately in the completion block -- won't that potentially kill requests lasting longer than the time it took to present? Why not just use the delegate properties to determine when to dismiss it? To your point about retries, I'm not sure that it retries, but this seems like it would cut things off for the main request early.

>The important change here is to stop presenting with `animated:NO` as that triggers a warning that end users see.

That's probably an unbalanced transition call, right? I wouldn't think that changing to `YES` there would fix that. In fact, I think that changing *to* animated is what usually causes that. Is there a reason you need the plain controller to present that Saf controller though? Why not just make it the root controller originally?

PR #188, opened 2015-09-21
invalidate NSURLSession to avoid memory leaks

derrickstaten:
@aaustin

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCServerInterface.m
+++ b/Branch-SDK/Branch-SDK/BNCServerInterface.m
@@ -121,6 +121,7 @@
         }
     }];
     [task resume];
+    [session finishTasksAndInvalidate];
 #else
             callback(serverResponse, error);
         }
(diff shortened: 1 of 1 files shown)


aaustin:
:+1:

Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -121,6 +121,7 @@ - (void)genericHTTPRequest:(NSURLRequest *)request retryNumber:(NSInteger)retryN
         }
     }];
     [task resume];
+    [session finishTasksAndInvalidate];
```
shortstuffsushi: Should this be done in the completion block?

derrickstaten: It doesn't appear to be necessary, per the [documentation](https://developer.apple.com/library/ios/documentation/Foundation/Reference/NSURLSession_class/#//apple_ref/occ/instm/NSURLSession/finishTasksAndInvalidate): `Once a session is invalidated, new tasks cannot be created in the session, but existing tasks continue until completion.`

shortstuffsushi: Ah, indeed. Should work then.

PR #190, opened 2015-09-22
Clear credits on logout

Sarkar:
If response dictionary has length (1 or more buckets for credits), update shared prefs with the values, otherwise, user has no reward history. Set count to 0 for default bucket. 

@aaustin @derrickstaten

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
@@ -412,6 +412,11 @@ NSString * const BRANCH_PREFS_KEY_UNIQUE_BASE = @"bnc_unique_base_";
     return [self.creditsDictionary[[BRANCH_PREFS_KEY_CREDIT_BASE stringByAppendingString:bucket]] integerValue];
 }
 
+- (void)clearUserCredits {
+    self.creditsDictionary = [[NSMutableDictionary alloc] init];
+    [self setCreditCount:0 forBucket:@"default"];
+}
+
 #pragma mark - Count Storage
 
 - (NSMutableDictionary *)countsDictionary {
--- a/Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
@@ -404,6 +404,23 @@ NSString * const BRANCH_PREFS_KEY_UNIQUE_BASE = @"bnc_unique_base_";
     [self writeObjectToDefaults:BRANCH_PREFS_KEY_CREDITS value:self.creditsDictionary];
 }
 
+- (void)removeCreditCountForBucket:(NSString *)bucket {
+    NSMutableDictionary *dictToWrite = self.creditsDictionary;
+    [dictToWrite removeObjectForKey:[BRANCH_PREFS_KEY_CREDIT_BASE stringByAppendingString:bucket]];
+
+    [self writeObjectToDefaults:BRANCH_PREFS_KEY_CREDITS value:self.creditsDictionary];
+}
+
+- (NSDictionary *)getCreditDictionary {
+    NSMutableDictionary *returnDictionary = [[NSMutableDictionary alloc] init];
+    for(NSString *key in self.creditsDictionary) {
+        NSString *cleanKey = [key stringByReplacingOccurrencesOfString:BRANCH_PREFS_KEY_CREDIT_BASE
+                                                                                     withString:@""];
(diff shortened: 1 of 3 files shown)


Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
```diff
@@ -412,6 +412,11 @@ - (NSInteger)getCreditCountForBucket:(NSString *)bucket {
     return [self.creditsDictionary[[BRANCH_PREFS_KEY_CREDIT_BASE stringByAppendingString:bucket]] integerValue];
 }
 
+- (void)clearUserCredits {
+    self.creditsDictionary = [[NSMutableDictionary alloc] init];
+    [self setCreditCount:0 forBucket:@"default"];
```
derrickstaten: why is this getting set?

shortstuffsushi:
What's the value of setting it to 0 vs just letting `counts[default]` return 0 for the nil / missing value?

derrickstaten:
both the previous solution and this solution fail to properly handle the case where

cached locally: 
"default": 1

on server:
"special": 1

--

the outcome after should be

cached locally:
"special": 1

unfortunately the actual result would be:
"default": 1,
"special": 1

shortstuffsushi:
So then you just need to start from a clean dictionary on each load, right? I'm actually curious why these are even kept locally, as they can be out of date at any point (the server has to re-verify when you try to actually use them). I had made a task at one point to just make the `getCounts` methods async, but never got to it.

Sarkar:
@derrickstaten - your use case is solved when `logout` is invoked - so we don't need accommodate that logic here. In the case of the same user no longer having "default" (or rather, that going to 0), then the result would be:

"default": 0
"special": 1

shortstuffsushi:
@Sarker what about when the user doesn't log out, but between sessions, their credit amounts change? Locally the value isn't cleared, so you'd still see one.

derrickstaten:
I think for a full solution, setting should be moved to BNCPreferenceHelper, which can then return hasChanged. 

As mentioned above, we have a scenario that isn't covered. With the current method, if a key isn't present in response.data but it is present locally on BNCPreferenceHelper, then we don't update locally -- we don't set that bucket to 0 (or remove the bucket entirely) on BNCPreferenceHelper.

Sarkar:
updated, please review @derrickstaten

Branch-SDK/Branch-SDK/Requests/BranchLoadRewardsRequest.m
```diff
@@ -39,19 +39,37 @@ - (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
...
+             [preferenceHelper setCreditCount:credits forBucket:key];
+        }
+        for(NSString *key in storedKeys) {
+            if(![response.data objectForKey:key]) {
+                [preferenceHelper removeCreditCountForBucket:key];
+                hasUpdated = YES;
+            }
+        }
+    } else {
+        if ([storedKeys count]) {
```
derrickstaten: it's safe to remove this if statement, right?

derrickstaten:
looks good :+1: 

It's also possible to locate the comparison logic in the BNCPreferenceHelper and have a method there return YES/NO for updated, which would involve less passing back and forth between preference helper and load rewards request. But I see no advantage to refactoring at this point. Good work!

Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
```diff
@@ -412,6 +429,11 @@ - (NSInteger)getCreditCountForBucket:(NSString *)bucket {
     return [self.creditsDictionary[[BRANCH_PREFS_KEY_CREDIT_BASE stringByAppendingString:bucket]] integerValue];
 }
 
+- (void)clearUserCredits {
+    self.creditsDictionary = [[NSMutableDictionary alloc] init];
+    [self writeObjectToDefaults:BRANCH_PREFS_KEY_CREDITS value:self.creditsDictionary];
```
derrickstaten: good

Branch-SDK/Branch-SDK/Requests/BranchLoadRewardsRequest.m
Sarkar: It's strictly for the case that user A had "default" : x, and suddenly doesn't have default anymore - usually a redeem takes care of this, but if a user didn't redeem inside the app, this could pop up

derrickstaten: useful for the hasUpdated = YES, got it

PR #192, opened 2015-09-22
fallback to /a/ link for BranchActivityItemProvider

derrickstaten:
@aaustin @Sarkar sorry for the nasty diff, I had to update a bunch of methods in order to pass one new param through

### DIFF
--- a/Branch-SDK/Branch-SDK/Branch.m
+++ b/Branch-SDK/Branch-SDK/Branch.m
@@ -993,9 +993,8 @@ static int BNCDebugTriggerFingersSimulator = 2;
         }
         else {
             if (forceLinkCreation) {
-                NSString *branchKey = [[BNCPreferenceHelper preferenceHelper] branchKey];
-                if (branchKey) {
-                    return [BranchShortUrlSyncRequest createLinkFromBranchKey:branchKey tags:tags alias:alias type:type matchDuration:duration channel:channel feature:feature stage:stage params:params];
+                if (self.branchKey) {
+                    return [BranchShortUrlSyncRequest createLinkFromBranchKey:self.branchKey tags:tags alias:alias type:type matchDuration:duration channel:channel feature:feature stage:stage params:params];
                 }
             }
             NSLog(@"Branch SDK Error: making request before init succeeded!");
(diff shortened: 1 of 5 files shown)


Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -988,6 +992,12 @@ - (NSString *)generateShortUrl:(NSArray *)tags andAlias:(NSString *)alias andTyp
             }
         }
         else {
+            if (forceLinkCreation) {
+                NSString *branchKey = [[BNCPreferenceHelper preferenceHelper] branchKey];
```
shortstuffsushi: This class has a `branchKey` property, shouldn't need to look up.

derrickstaten: oh, nice! good catch

derrickstaten:
sorry got distracted, self.branchKey is used now

Branch-SDK/Branch-SDK/Requests/BranchShortUrlSyncRequest.m
```diff
@@ -78,36 +79,44 @@ - (NSString *)processResponse:(BNCServerResponse *)response {
 }
 
 - (NSString *)createLongUrlForUserUrl:(NSString *)userUrl {
-    NSMutableString *longUrl = [[NSMutableString alloc] initWithFormat:@"%@?", userUrl];
-    
-    for (NSString *tag in self.tags) {
-        [longUrl appendFormat:@"tags=%@&", tag];
+    NSMutableString *baseUrl = [[NSMutableString alloc] initWithFormat:@"%@?", userUrl];
+    return [BranchShortUrlSyncRequest createLongUrlWithBaseUrl:baseUrl tags:self.tags alias:self.alias type:self.type matchDuration:self.matchDuration channel:self.channel feature:self.feature stage:self.stage params:self.params];
```
Sarkar: it looks like `createLongUrlForUserUrl` is being replaced by `createLongUrlWithBaseUrl`, so that all paths that traverse down this will just hit what you wrote, plus `createLinkFromBranchKey` going down `createLongUrlWithBaseUrl`, right?

Sarkar:
I've tested locally with 100% loss and forcing a fallback and received a link as expected with all the right data, looks good to me :+1:

Branch-SDK/Branch-SDK/Requests/BranchShortUrlSyncRequest.m
derrickstaten: that's correct I didn't want to have duplicate code

PR #195, opened 2015-09-25
make spotlight search safer for bad inputs

derrickstaten:
@Sarkar @aaustin

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCContentDiscoveryManager.m
+++ b/Branch-SDK/Branch-SDK/BNCContentDiscoveryManager.m
@@ -135,14 +135,17 @@
         }
         return;
     }
-    if (!title) {
+    if (!title || ![title isKindOfClass:[NSString class]]) {
         if (callback) {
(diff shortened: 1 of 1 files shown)


Branch-SDK/Branch-SDK/BNCContentDiscoveryManager.m
```diff
@@ -135,14 +135,17 @@ - (void)indexContentWithTitle:(NSString *)title description:(NSString *)descript
         }
         return;
     }
-    if (!title) {
+    if (!title || ![title isKindOfClass:[NSString class]]) {
```
shortstuffsushi: The method def here specifically calls for an `NSString`, type checking that seems unnecessary.

shortstuffsushi:
This whole thing looks excessive. What are the use cases for people passing in specifically bad typed values into a method with explicit types?

aaustin:
:+1:

derrickstaten:
I agree that this is excessive. I don't see any real value-add here. I'm going to close this for now.

PR #203, opened 2015-10-08
Added new license

AustinHay:
@dmitrig01


shortstuffsushi:
Hey @AustinHay, we [already have](https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK/blob/master/LICENSE) a license file, would you consider editing that instead?

AustinHay:
Something was messed up here. Just pushing a new updated branch w/ the changes. Delete both these branches when done.

PR #205, opened 2015-10-14
Disable the Enabe Modules option

Sarkar:
CLANG_ENABLE_MODULES set to NO. Eliminates our module cache warnings: https://forums.developer.apple.com/thread/17921

@derrickstaten @aaustin


derrickstaten:
:+1:

PR #206, opened 2015-10-17
BranchUniversalObject 

derrickstaten:
@aaustin @sojanpr @Sarkar first commit, still need:
- [x] share sheet
- [x] core spotlight indexing
- [x] more manual testing to find edge cases


Branch-SDK/Branch-SDK/Branch.h
```diff
@@ -1334,4 +1365,19 @@ typedef NS_ENUM(NSUInteger, BranchPromoCodeUsageType) {
...
+- (void)registerViewWithParams:(NSDictionary *)params andCallback:(callbackWithParams)callback;
+
+/**
+ Gets the global Branch instance, whether live or test, but only if a key has been specified.
+ 
+ @warning This is meant for use internally only and should not be used by apps. Invoking this method could result
+ in broken functionality.
+ */
++ (Branch *)getCurrentInstanceIfAny;
+
```
aaustin: Don't put this in the header if it's for internal use

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1060,6 +1073,24 @@ - (BNCLinkData *)prepareLinkDataFor:(NSArray *)tags andAlias:(NSString *)alias a
 }
 
 
+#pragma mark - BranchUniversalObject methods
+
++ (Branch *)getCurrentInstanceIfAny {
+    BNCPreferenceHelper *preferenceHelper = [BNCPreferenceHelper preferenceHelper];
+    return [Branch getInstanceInternal:preferenceHelper.lastRunBranchKey returnNilIfNoCurrentInstance:NO];
+}
+
```
aaustin: I think you mean yes as the argument here but I disagree that we shouldn't do anything if the simpleton isn't initialized. More on this below.

Branch-SDK/Branch-SDK/BranchUniversalObject.m
```diff
@@ -0,0 +1,137 @@
...
+- (instancetype)initWithTitle:(NSString *)title {
+    if (self = [super init]) {
+        self.title = title;
+    }
+    return self;
+}
+
+- (void)registerView {
+    Branch *branch = [Branch getCurrentInstanceIfAny];
+    if (!branch) {
```
aaustin: All of our other methods will automatically init branch if its not done so by the developer. I think we should take that same course here and create the singleton automatically for the developer. You already have the logic to call in it automatically anyways.

aaustin:
Other than those comments, all looks good

Branch-SDK/Branch-SDK/BranchUniversalObject.h
```diff
@@ -0,0 +1,41 @@
...
+@property (nonatomic) ContentIndexMode contentIndexMode;
+@property (nonatomic, strong) NSArray *keywords;
+@property (nonatomic) NSUInteger expirationInMilliSec;
+
+- (instancetype)initWithCanonicalIdentifier:(NSString *)canonicalIdentifier;
+- (instancetype)initWithTitle:(NSString *)title;
+- (void)registerView;
+- (void)registerViewWithCallback:(callbackWithParams)callback;
+- (NSString *)getShortUrlWithLinkProperties:(BranchLinkProperties *)linkProperties;
+- (void)getShortUrlWithLinkProperties:(BranchLinkProperties *)linkProperties andCallback:(callbackWithUrl)callback;
```
sojanpr: Minor, Are we using `getShortUrl()` method name for both sync and Async calls . On Android we are following convention `public String getShortLink()` for sync call and `public void generateShortLink(callback)` for Async

Branch-SDK/Branch-SDK/BranchUniversalObject.m
```diff
@@ -0,0 +1,137 @@
...
+    [branch registerViewWithParams:[self getParamsForServerRequest] andCallback:callback];
+}
+
+- (NSString *)getShortUrlWithLinkProperties:(BranchLinkProperties *)linkProperties {
+    Branch *branch = [Branch getCurrentInstanceIfAny];
+    if (!branch) {
+        NSLog(@"[Branch Warning] getInstance has not yet been invoked, so could not generate a URL.");
+        return nil;
+    }
+    if (!self.canonicalIdentifier && !self.title) {
```
sojanpr: not sure whether we need this check since BOU can be initialized only  with title or Canonical ID

Branch-SDK/Branch-SDK/BranchLinkProperties.m
```diff
@@ -0,0 +1,29 @@
...
+
+- (NSDictionary *)controlParams {
+    if (!_controlParams) {
+        _controlParams = [[NSDictionary alloc] init];
+    }
+    return _controlParams;
+}
+
+- (void)addControlParam:(NSString *)controlParam withValue:(NSString *)value {
+    if (!controlParam || ! value) {
```
Sarkar: what are you checking with `!controlParam` and `!value`? if you're going to check null ness and emptiness, I recommend you compare agains `[NSNull null]` and then check length of string.

Branch-SDK/Branch-SDK/BranchUniversalObject.m
```diff
@@ -0,0 +1,137 @@
...
+    return self;
+}
+
+- (void)registerView {
+    Branch *branch = [Branch getCurrentInstanceIfAny];
+    if (!branch) {
+        NSLog(@"[Branch Warning] getInstance has not yet been invoked, so could not register view.");
+        return;
+    }
+    if (!self.canonicalIdentifier && !self.title) {
```
Sarkar: same as above, ideally, with strings, should probably make sure they aren't empty quotes or nil.

Branch-SDK/Branch-SDK/Branch.h
derrickstaten: How would you reference it from `BranchUniversalObject` then? At the very least, it's unlikely a dev would find it since I named it `getCurrentInstance...`, not `getInstance...`

Branch-SDK/Branch-SDK/BranchLinkProperties.m
derrickstaten: The backend will still have to do validation. This is the standard validation we do elsewhere in the SDK. Even checking null ness, emptiness, length... all these things could be circumvented by specifying @"-" as `controlParam` or `value`. I'm going to keep this the same as validation elsewhere in the SDK for now

Branch-SDK/Branch-SDK/BranchUniversalObject.h
derrickstaten: in iOS SDK, we use `getShortUrl` for both, and the return type (as well as the presence or lack of a callback parameter) indicates whether it's synchronous or asynchronous. perhaps not the best design, but this is parallel to the existing SDK

Branch-SDK/Branch-SDK/BranchUniversalObject.m
derrickstaten: the difference is in that in all other cases, the developer has specified `getInstance` or `getTestInstance` -- with BranchUniversalObject we have to decide whether to use the test or live key. So `getCurrentInstanceIfAny` make sure it's invoked. This is different from checking whether the SDK is initialized. As long as the synchronous call to `getInstance` or `getTestInstance` has taken place, we'll know whether to use the test or live key. Maybe we should walk through this in person.

Also, a side note, the content indexing wrapper currently has this problem but blindly ignored it (I didn't think through this issue at the time). It always uses `getInstance` even if the dev specified `getTestInstance` in `application:didFinishLaunchingWithOptions:`

Branch-SDK/Branch-SDK/BranchUniversalObject.m
derrickstaten: same response as above. we always still need server-side validation, and the dev can get dirty data by passing in @"-" and @"-". so this is supposed to cover the general case but not be bulletproof on its own

Branch-SDK/Branch-SDK/BranchUniversalObject.m
derrickstaten: someone could fairly easily set both to nil after initialization. This is a nice check to help devs catch issues early on.

Branch-SDK/Branch-SDK/BranchUniversalObject.m
derrickstaten: This faulty logic can also be found in the share sheet. @aaustin I could emulate the existing faulty logic and simply call `[Branch getInstance]`, remove the new method, and remove the checks for whether it's been invoked or not

Branch-SDK/Branch-SDK/BranchUniversalObject.h
```diff
@@ -0,0 +1,42 @@
...
+typedef NS_ENUM(NSInteger, ContentIndexMode) {
+    ContentIndexModePublic,
+    ContentIndexModePrivate
+};
+
+@property (nonatomic, strong) NSString *canonicalIdentifier;
+@property (nonatomic, strong) NSString *title;
+@property (nonatomic, strong) NSString *contentDescription;
+@property (nonatomic, strong) NSString *imageUrl;
+@property (nonatomic, strong) NSDictionary *metatdata;
```
aaustin: typo

Branch-SDK/Branch-SDK/BranchUniversalObject.h
```diff
@@ -0,0 +1,42 @@
...
+@property (nonatomic, strong) NSString *type;
+@property (nonatomic) ContentIndexMode contentIndexMode;
+@property (nonatomic, strong) NSArray *keywords;
+@property (nonatomic) NSUInteger expirationInMilliSec;
```
aaustin: I think it makes more sense for this to be a NSDate and convert that to milliseconds before submission. I'm assuming this is how it is on Android too, so I'll make a note to change it.

aaustin: Ah. The android version has a setter which converts date into epoch milliseconds.

aaustin:
This looks great after typo fix. :+1:

Branch-SDK/Branch-SDK/BranchUniversalObject.h
derrickstaten: good catch

derrickstaten:
Sorry for the nasty diff. I decided to include more standard spacing here in Branch.h and Branch.m. If you hate it, I can spend ~10-15 min removing

derrickstaten:
Now the testbed app has BranchUniversalObject example

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1245,10 +1304,15 @@ - (void)registerInstallOrOpen:(Class)clazz {
...
+    
     NSDictionary *latestReferringParams = [self getLatestReferringParams];
-    if (self.shouldCallSessionInitCallback && self.sessionInitWithParamsCallback) {
-        self.sessionInitWithParamsCallback(latestReferringParams, nil);
+    if (self.shouldCallSessionInitCallback) {
+        if (self.sessionInitWithParamsCallback) {
+            self.sessionInitWithParamsCallback(latestReferringParams, nil);
+        }
+        else if (self.sessionInitWithBranchUniversalObjectCallback) {
+            self.sessionInitWithBranchUniversalObjectCallback([self getLatestReferringBranchUniversalObject], [self getLatestReferringBranchLinkProperties], nil);
```
aaustin: Can we test this with poor connections and try to simulate 500s?

Branch-SDK/Branch-SDK/BranchUniversalObject.h
```diff
@@ -0,0 +1,47 @@
...
+- (instancetype)initWithCanonicalIdentifier:(NSString *)canonicalIdentifier;
+- (instancetype)initWithTitle:(NSString *)title;
+- (void)addMetadataKey:(NSString *)key value:(NSString *)value;
+- (void)registerView;
+- (void)registerViewWithCallback:(callbackWithParams)callback;
+- (NSString *)getShortUrlWithLinkProperties:(BranchLinkProperties *)linkProperties;
+- (void)getShortUrlWithLinkProperties:(BranchLinkProperties *)linkProperties andCallback:(callbackWithUrl)callback;
+- (UIActivityItemProvider *)getBranchActivityItemWithLinkProperties:(BranchLinkProperties *)linkProperties;
+- (void)showShareSheetWithShareText:(NSString *)shareText;
+- (void)showShareSheetWithLinkProperties:(BranchLinkProperties *)linkProperties andShareText:(NSString *)shareText fromViewController:(UIViewController *)viewController;
```
aaustin: Let's make the share sheet callback a requirement for both of these methods. Every time I've used it, I always put some completion code in the callback.

aaustin:
after you add that callback :+1:

PR #217, opened 2015-11-06
Added logoutWithCallback to Branch object

danwalkowski:
@aaustin @derrickstaten


derrickstaten:
instead of having duplicate logic in `logout` and `logoutWithCallback`, can we combine them so that `logout` simply invokes `logoutWithCallback:nil`?

derrickstaten:
Also, before you do that, can we clean up the diff by cutting this branch off master instead of `deprecate_promo_codes`?

PR #218, opened 2015-11-06
After investigation, this turned out to be a one-liner.    

danwalkowski:
To be clear, this was to fix sessionID not getting updated on SetIdentityRequest

The information was there, and if it gets changed, it is persisted, the sessionID was simply missing from the set.
@aaustin @derrickstaten


derrickstaten:
:+1: nice

PR #229, opened 2015-11-16
allow dashboard debug mode on iOS 9

derrickstaten:
@danwalkowski @rubinsingh

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCSystemObserver.m
+++ b/Branch-SDK/Branch-SDK/BNCSystemObserver.m
@@ -146,10 +146,14 @@
 
 + (BOOL)isSimulator {
     UIDevice *currentDevice = [UIDevice currentDevice];
+    NSString *device;
     if ([BNCSystemObserver getOSVersion].integerValue >= 9) {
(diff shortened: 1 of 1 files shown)


Branch-SDK/Branch-SDK/BNCSystemObserver.m
```diff
@@ -146,6 +146,9 @@ + (NSString *)getModel {
 
 + (BOOL)isSimulator {
     UIDevice *currentDevice = [UIDevice currentDevice];
+    if ([BNCSystemObserver getOSVersion].integerValue >= 9) {
```
shortstuffsushi: The check between the two is the same; you DRY it up a bit with something like

```
NSString *device = [BNCSystemObserver getOSVersion].integerValue >= 9 ? currentDevice.name : currentDevice.model;

return [device rangeOfString:@"Simulator"].location != NSNotFound;

derrickstaten: thanks! decided not to use a ternary since it's barely used in the SDK

danwalkowski:
:+1:   var just changes names

PR #230, opened 2015-11-16
fix race condition by moving the copy operation into the queued task

derrickstaten:
@danwalkowski @aaustin the preference helper probably needs a rewrite at some point, as it's possible for us to persist state that's different from the preference helper's final state. I can demonstrate the race condition in person, but ultimately it's related to changing several properties on pref helper in rapid succession. It seems the queueing mechanism seen below can result in saving old preferences to disk. E.g. if 5 properties are changed in a row, and the 4th save to disk somehow occurs after the 5th save (aka they finish in opposite order), then the final state persisted will exclude the change of the 5th property. 

Note that I have to kill the app to see this issue.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
@@ -492,8 +492,8 @@ NSString * const BRANCH_PREFS_KEY_UNIQUE_BASE = @"bnc_unique_base_";
 }
 
 - (void)persistPrefsToDisk {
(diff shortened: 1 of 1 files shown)


derrickstaten:
One partner saw an error on line 500 of preference helper. So we should test this more thoroughly. Again, persisting the preference helper data probably needs a rewrite.

Error they saw:
```
dodv3(9354,0x700000d24000) malloc: *** error for object 0x7fcdcad21c50: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
```

derrickstaten:
@danwalkowski please take a look when you can

PR #231, opened 2015-11-17
make sure that we don't set a nil value for session id

derrickstaten:
@danwalkowski @aaustin

### DIFF
--- a/Branch-SDK/Branch-SDK/Requests/BranchSetIdentityRequest.m
+++ b/Branch-SDK/Branch-SDK/Requests/BranchSetIdentityRequest.m
@@ -57,7 +57,9 @@
     preferenceHelper.identityID = response.data[BRANCH_RESPONSE_KEY_BRANCH_IDENTITY];
     preferenceHelper.userUrl = response.data[BRANCH_RESPONSE_KEY_USER_URL];
     preferenceHelper.userIdentity = self.userId;
-    preferenceHelper.sessionID = response.data[BRANCH_RESPONSE_KEY_SESSION_ID];
+    if (response.data[BRANCH_RESPONSE_KEY_SESSION_ID]) {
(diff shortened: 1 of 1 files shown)


derrickstaten:
for reference, https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK/pull/218

aaustin:
:+1:

Branch-SDK/Branch-SDK/Requests/BranchSetIdentityRequest.m
```diff
@@ -57,7 +57,9 @@ - (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
     preferenceHelper.identityID = response.data[BRANCH_RESPONSE_KEY_BRANCH_IDENTITY];
     preferenceHelper.userUrl = response.data[BRANCH_RESPONSE_KEY_USER_URL];
     preferenceHelper.userIdentity = self.userId;
-    preferenceHelper.sessionID = response.data[BRANCH_RESPONSE_KEY_SESSION_ID];
+    if (response.data[BRANCH_RESPONSE_KEY_SESSION_ID]) {
```
shortstuffsushi: Why would this be possible? This seems to indicate a problem elsewhere. Kinda similar to that type checking thing, it doesn't seem like the best idea to just wrap everything in if statements. I know it's just the one now, but you could really do the same for any of these.

derrickstaten: We only sent down a session_id if it had changed. So this was a unique situation. (Last night to remedy this in realtime, we started sending down session_id 100% of the time.)

shortstuffsushi: :+1: to that. That seems like the more consistent solution.

danwalkowski:
I also :+1: this pr, and the 'always return a sessionID' behavior

PR #232, opened 2015-11-17
Readd initsession check

derrickstaten:
@danwalkowski @aaustin note that this purposefully includes one other commit (submitted in a different PR) so that I can ask a partner to test this.


aaustin:
Isn't this a lot of what you had before?

derrickstaten:
Yes, it's exactly what we had before. We should put this back in, as it prevents race conditions when initSession is in progress but not yet finished

danwalkowski:
:+1:   this pr was blamed for the problems caused by the other pr that sometimes produced a nil sessionID

danwalkowski:
:+1:

PR #234, opened 2015-11-17
Fix setting default value for userIdentity

zacwest:


danwalkowski:
:+1: certainly _looks_ like the right thing to do :)

danwalkowski:
:+1:

PR #237, opened 2015-11-20
after initsession, close the app if now in bg

derrickstaten:
@danwalkowski @aaustin we'll need to test this extensively. I can demonstrate the race condition in person, using a VPN through HK to simulate slow Internet. This is a rare condition that will almost never occur.


derrickstaten:
Also, this is a proof of concept. We should refactor the close-related methods so they're a bit less redundant.

aaustin:
this seems harmless due to the check in callClose :+1:

Issue #193, opened 2015-09-23
Test failed in iOS9


Hibbert-pku:
Hi,
    iOS version of branchIO in circleCI is iOS8,so when I use xcode7&iOS9 to run the test, the BNCServerInterfaceTests was failed. After I delete the iOS9 code in BNCServerInterface.m between line 77 and 128, the tests were passed.

Sarkar:
Hibbert,

Thanks for telling us about this. We're certainly aware of this ourselves, and we're going to update tests very shortly. Please hang tight!

cc @aaustin @derrickstaten

derrickstaten:
Tests are passing on master again. Resolving.

PR #238, opened 2015-11-21
Add support for setting email subject when using BranchUniversalObject.

mikeantonelli:
This Pull Request will allow consumers to set the potential email subject line if Mail is selected from the share sheet presented for a BranchUniversalObject.

### DIFF
--- a/Branch-SDK/Branch-SDK/BranchUniversalObject.h
+++ b/Branch-SDK/Branch-SDK/BranchUniversalObject.h
@@ -29,7 +29,6 @@ typedef NS_ENUM(NSInteger, ContentIndexMode) {
 @property (nonatomic) ContentIndexMode contentIndexMode;
 @property (nonatomic, strong) NSArray *keywords;
 @property (nonatomic, strong) NSDate *expirationDate;
-@property (nonatomic, strong) NSString *emailSubject;
 
 - (instancetype)initWithCanonicalIdentifier:(NSString *)canonicalIdentifier;
 - (instancetype)initWithTitle:(NSString *)title;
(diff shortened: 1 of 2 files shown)


mikeantonelli:
I checked out master and ran tests, it looks like the two tests that are failing also currently fail in master - might be related to https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK/issues/193?

danwalkowski:
Thanks Mike,
I will review it.

derrickstaten:
@mikeantonelli thanks for submitting this! We're trying to keep this class pretty lightweight and not let it get bloated with a lot of methods (we're learning from other parts of our code base). We may instead refactor this to use a special key like `$email_subject` that can be set as a control parameter on the LinkProperties object. How does that sound?

mikeantonelli:
I originally added a property on `BranchUniversalObject` and spoke with @aaustin over email about extending the existing `showShareSheet` API. It's a pretty simple refactor to make it a control param on `BranchLinkProperties`, sound good @aaustin?

derrickstaten:
@mikeantonelli can you check out this branch? https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK/tree/email-subject

Specifically, https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK/blob/email-subject/Branch-SDK/Branch-SDK/BranchUniversalObject.m#L143-L150

mikeantonelli:
Yep, exactly what I was thinking @derrickstaten - have you guys thought about using `FOUNDATION_EXPORT` instead of `extern`?

derrickstaten:
@mikeantonelli convention more than anything is the reason we use `extern`. Do you think there are any tangible benefits to `FOUNDATION_EXPORT`? 

Also, I'm opening a PR with the code in that branch. Can you make sure it works as expected for you?

derrickstaten:
I'm closing this for now, since we refactored in https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK/pull/245. Thanks for all the help here @mikeantonelli! We'll likely get this in as part of a release tomorrow.

PR #239, opened 2015-11-23
Client timeout changes

danwalkowski:
- changed network timeout from 5 seconds to 5.5 seconds
- changed retry count from 1 to 3


derrickstaten:
3 seems high, but if that's what @aaustin decided, then :+1:

danwalkowski:
It is indeed what he asked for.

PR #243, opened 2015-11-30
Pass content expiration date through to spotlight index

danwalkowski:
Concerted effort to pass through the expiration date from the partner API to Spotlight.  This required far more changes than it ought to have.

In the end, although it -appears- that everything is correctly configured, the items do not get flushed from Spotlight, despite the correct date appearing in the mdItemExpirationDate field of the CSSearchItem.
I'm checking this in on a floating branch for further investigation.


danwalkowski:
btw: the changes include an update to the Testbed App, which allow setting the expiration date.  
This dovetails into the changes requested for the full BUO makeover of the Testbed App.

derrickstaten:
:+1:

PR #244, opened 2015-12-01
warn if easy deeplinking view controller...

derrickstaten:
...does not implement configure ControlWithData:


derrickstaten:
@danwalkowski

aaustin:
Nice :+1:

PR #247, opened 2015-12-02
100 matching prototype

derrickstaten:
@danwalkowski @aaustin @Sarkar 100% matching is back

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
@@ -19,6 +19,7 @@
 
 + (BNCStrongMatchHelper *)strongMatchHelper { return nil; }
 - (void)createStrongMatchWithBranchKey:(NSString *)branchKey { }
+- (BOOL)shouldDelayInstallRequest { return NO; }
 
 @end
 
@@ -30,6 +31,7 @@ NSInteger const ABOUT_30_DAYS_TIME_IN_SECONDS = 60 * 60 * 24 * 30;
 
 @property (strong, nonatomic) UIWindow *secondWindow;
 @property (assign, nonatomic) BOOL requestInProgress;
+@property (assign, nonatomic) BOOL shouldDelayInstallRequest;
 
 @end
 
@@ -60,6 +62,7 @@ NSInteger const ABOUT_30_DAYS_TIME_IN_SECONDS = 60 * 60 * 24 * 30;
         return;
     }
     
+    self.shouldDelayInstallRequest = YES;
     [self presentSafariVCWithBranchKey:branchKey];
 }
 
@@ -70,6 +73,8 @@ NSInteger const ABOUT_30_DAYS_TIME_IN_SECONDS = 60 * 60 * 24 * 30;
     BOOL isRealHardwareId;
     NSString *hardwareId = [BNCSystemObserver getUniqueHardwareId:&isRealHardwareId andIsDebug:preferenceHelper.isDebug];
     if (!hardwareId || !isRealHardwareId) {
+        NSLog(@"[Branch Warning] Cannot use cookie-based matching while setDebug is enabled");
+        self.shouldDelayInstallRequest = NO;
         self.requestInProgress = NO;
         return;
     }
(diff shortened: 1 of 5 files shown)


Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
```diff
@@ -99,24 +104,31 @@ - (void)presentSafariVCWithBranchKey:(NSString *)branchKey {
...
+                self.requestInProgress = NO;
+            }];
+        });
     }
     else {
         self.requestInProgress = NO;
     }
 }
 
+- (BOOL)shouldDelayInstallRequest {
```
shortstuffsushi: This method shouldn't be necessary -- the property generates it for you.

Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
```diff
@@ -99,24 +104,31 @@ - (void)presentSafariVCWithBranchKey:(NSString *)branchKey {
     if (SFSafariViewControllerClass) {
         id safController = [[SFSafariViewControllerClass alloc] initWithURL:[NSURL URLWithString:urlString]];
         
+        self.secondWindow = [[UIWindow alloc] initWithFrame:[[[[UIApplication sharedApplication] delegate] window] bounds]];
```
shortstuffsushi: Is the reordering here significant?

Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
```diff
@@ -99,24 +104,31 @@ - (void)presentSafariVCWithBranchKey:(NSString *)branchKey {
...
         UIViewController *windowRootController = [[UIViewController alloc] init];
-        
-        self.secondWindow = [[UIWindow alloc] initWithFrame:CGRectZero];
         self.secondWindow.rootViewController = windowRootController;
-        [self.secondWindow makeKeyAndVisible];
+        [self.secondWindow setHidden:NO];
```
shortstuffsushi: Is setting this hidden, but not showing it important? Also, does this block UI input?

Branch-SDK/Branch-SDK/Requests/BranchInstallRequest.m
```diff
@@ -48,7 +49,14 @@ - (void)makeRequest:(BNCServerInterface *)serverInterface key:(NSString *)key ca
     params[BRANCH_REQUEST_KEY_IS_REFERRABLE] = @(preferenceHelper.isReferrable);
     params[BRANCH_REQUEST_KEY_DEBUG] = @(preferenceHelper.isDebug);
     
-    [serverInterface postRequest:params url:[preferenceHelper getAPIURL:BRANCH_REQUEST_ENDPOINT_INSTALL] key:key callback:callback];
+    if ([[BNCStrongMatchHelper strongMatchHelper] shouldDelayInstallRequest]) {
+        dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(500 * NSEC_PER_MSEC)), dispatch_get_main_queue(), ^{
+            [serverInterface postRequest:params url:[preferenceHelper getAPIURL:BRANCH_REQUEST_ENDPOINT_INSTALL] key:key callback:callback];
```
shortstuffsushi: Relying on time makes me somewhat uncomfortable. Might it be better to at least wait the duration of th request timeout, so the request will have certainly completely or been killed?

derrickstaten: this is a lighter weight request than the v1/open or v1/install request. so if it is kicked off first it will almost always finish first. I don't want to delay initializing the session longer than that

derrickstaten:
@aaustin new commit added

Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
derrickstaten: it doesn't block input, and this is the "magic" we can use to avoid ever setting it as the key window

Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
derrickstaten: no

Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
derrickstaten: I may have just been following convention in the rest of the class

Branch-SDK/Branch-SDK/Requests/BranchInstallRequest.m
shortstuffsushi: > it will almost always finish first

Read: be prepared for race conditions.

Naturally, your choice to do so, but this may lead to issues. Fortunately, this request is fairly insignificant -- if it fails, life goes on, and the only loss is potentially 100% match (which falls to probably what, 95+?).

If this request was required, this would not work, but for the role is fills, I think it'll be ok.

derrickstaten: agreed

MikeSilvis:
What is the status of this? Would love to get this into our upcoming release.

aaustin:
Approved :+1:

PR #248, opened 2015-12-03
Add canonical URL

aaustin:
@derrickstaten


derrickstaten:
Let's also add a check here to allow an object with only a canonical URL: https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK/blob/canonical-url/Branch-SDK/Branch-SDK/BranchUniversalObject.m#L72

derrickstaten:
added check :+1:

PR #253, opened 2015-12-10
Proposed small fix to get the team identifier working in iOS 9.2

danwalkowski:
We need to have a discussion about this; it seems that Apple broke access to poking around in the kSecClass:  classes, (iOS 92. latest Xcode) which we were using to get the Team ID.

I found another way, but it requires the partner to add a key to their plist.

### DIFF
--- a/Branch-TestBed/Branch-TestBed/Branch-TestBed-Info.plist
+++ b/Branch-TestBed/Branch-TestBed/Branch-TestBed-Info.plist
@@ -45,6 +45,8 @@
 	<array>
 		<string>UIInterfaceOrientationPortrait</string>
 	</array>
+	<key>AppIdentifierPrefix</key>
+	<string>$(AppIdentifierPrefix</string>
 	<key>branch_key</key>
 	<dict>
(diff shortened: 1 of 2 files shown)


danwalkowski:
additional info: the plist key is filled in at build time

Branch-TestBed/Branch-TestBed/Branch-TestBed-Info.plist
```diff
@@ -45,6 +45,8 @@
 	<array>
 		<string>UIInterfaceOrientationPortrait</string>
 	</array>
+	<key>AppIdentifierPrefix</key>
+	<string>$(AppIdentifierPrefix</string>
```
derrickstaten: should this be `$(AppIdentifierPrefix)`? seems weird to have just an open parens

derrickstaten: also, is this the required plist entry?

derrickstaten:
:+1:

PR #259, opened 2015-12-16
* implemented externalIntentURI

danwalkowski:
which required more effort than I expected, as there were several likely looking spots in the code, which had to be traced and evaluated.

* Sojan provided helpful advice.


Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -384,6 +384,9 @@ - (BOOL)handleDeepLink:(NSURL *)url {
 }
 
 - (BOOL)continueUserActivity:(NSUserActivity *)userActivity {
+    //always save the incoming url in the preferenceHelper, as the externalIntentURI
+    self.preferenceHelper.externalIntentURI = [userActivity.webpageURL absoluteString];
```
shortstuffsushi: Note that this is saving (and later sending, in OpenRequest) *ALL* incoming urls, including those in no way related to Branch. Not that it's invalid, just pointing it out for your consideration.

derrickstaten: I don't think we want this -- we already send this up as `universal_link_url`. Just need the URIs

danwalkowski:
The Android version sends all the URLs, so I am doing the same.

danwalkowski:
thumb?

Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
```diff
@@ -32,6 +32,7 @@
 NSString * const BRANCH_PREFS_KEY_USER_URL = @"bnc_user_url";
 NSString * const BRANCH_PREFS_KEY_IS_REFERRABLE = @"bnc_is_referrable";
 NSString * const BRANCH_PREFS_KEY_BRANCH_UNIVERSAL_LINK_DOMAINS = @"branch_universal_link_domains";
+NSString * const BRANCH_REQUEST_KEY_EXTERNAL_INTENT_URI = @"external_intent_uri";
```
derrickstaten: this does not compile for me because this symbol is defined 2x (see also https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK/pull/259/files#diff-05f7881939e7f659c301d3248c960442R63)

derrickstaten:
Please remove the duplicate definition of the constant

Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
danwalkowski: weird! It works fine for me.  I'll figure out what is going on

derrickstaten:
saw some strange behavior that went away when I merged in master. :+1:

PR #260, opened 2015-12-17
Broken UIViewControllerBasedStatusBarAppearance and unbalance view transition calls

joshuafeldman:
Fixes an issue with unbalanced calls to view transitions. Also fixes an issue where view controller based status bar appearance is not restored.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
@@ -112,11 +112,23 @@ NSInteger const ABOUT_30_DAYS_TIME_IN_SECONDS = 60 * 60 * 24 * 30;
         
         // Must be on next run loop to avoid a warning
         dispatch_async(dispatch_get_main_queue(), ^{
-            [windowRootController presentViewController:safController animated:YES completion:^{
-                [self.secondWindow.rootViewController dismissViewControllerAnimated:YES completion:NULL];
-                [BNCPreferenceHelper preferenceHelper].lastStrongMatchDate = [NSDate date];
-                self.requestInProgress = NO;
-            }];
+            // Add the safari view controller using view controller containment
+            [windowRootController addChildViewController:safController];
+            [windowRootController.view addSubview:safController.view];
+            [safController didMoveToParentViewController:windowRootController];
+            
+            // Remove the safari view controller from view controller containment
+            [safController willMoveToParentViewController:nil];
+            [safController.view removeFromSuperview];
+            [safController removeFromParentViewController];
+            
+            // Remove the window and release it's strong reference. This is important to ensure that
+            // applications using view controller based status bar appearance are restored.
+            [self.secondWindow removeFromSuperview];
+            self.secondWindow = nil;
+            
+            [BNCPreferenceHelper preferenceHelper].lastStrongMatchDate = [NSDate date];
+            self.requestInProgress = NO;
         });
     }
     else {
--- a/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
@@ -102,7 +102,7 @@ NSInteger const ABOUT_30_DAYS_TIME_IN_SECONDS = 60 * 60 * 24 * 30;
     
     Class SFSafariViewControllerClass = NSClassFromString(@"SFSafariViewController");
     if (SFSafariViewControllerClass) {
-        id safController = [[SFSafariViewControllerClass alloc] initWithURL:[NSURL URLWithString:urlString]];
+        UIViewController * safController = [[SFSafariViewControllerClass alloc] initWithURL:[NSURL URLWithString:urlString]];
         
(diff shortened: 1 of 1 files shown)


Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
```diff
@@ -112,11 +112,23 @@ - (void)presentSafariVCWithBranchKey:(NSString *)branchKey {
...
-                [self.secondWindow.rootViewController dismissViewControllerAnimated:YES completion:NULL];
-                [BNCPreferenceHelper preferenceHelper].lastStrongMatchDate = [NSDate date];
-                self.requestInProgress = NO;
-            }];
+            // Add the safari view controller using view controller containment
+            [windowRootController addChildViewController:safController];
+            [windowRootController.view addSubview:safController.view];
+            [safController didMoveToParentViewController:windowRootController];
+            
+            // Remove the safari view controller from view controller containment
```
joshuafeldman: I am actually not sure if removing it is actually needed. We can probably actually remove this since it will get deallocated when the window is released.

derrickstaten: @joshuafeldman Can you test and see? Then I'll test on my machine. I need to make sure that this code is still making an outbound request to `bnc.lt/_strong_match`.

derrickstaten:
@joshuafeldman thanks for contributing here -- we really appreciate it. After you decide whether to remove the safari view controller on line 120 above, we'll test on our end to make sure it's still sending outbound requests.

Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
joshuafeldman: @derrickstaten how much time does that request need?

derrickstaten: in our testing, almost no time at all. I'll go ahead and test what you've written, and I'll let you know if anything needs adjusting. I'll probably change the URL to point to our staging servers, and tail those logs. Give me a few minutes.

joshuafeldman: Ok let me know. I am around. Another optional could be to use `SFSafariViewControllerDelegate` to only tear down the window once the initial load has completed.

derrickstaten: Unfortunately we can't do that because of the way we're using SFSafariViewController. If we required everyone to link their project to SFSafariServices / use Xcode 7+, then we could write clearer code instead of using `NSClassFromString`, but we want to keep that optional.

joshuafeldman:
My pleasure. The only small downside to this fix is there is a very slight blink of the status bar. I am not sure if you can indicate that this secondary window should not be used for status bar appearance. It might have something to do with the order. Even though your window is hidden it may have a higher index than the application's key window.

Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
derrickstaten: Bad news, this isn't long enough. If you wrap the removal in a dispatch after .3 seconds, it will work:
```
            dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(.3 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{

                // Remove the safari view controller from view controller containment
                [safController willMoveToParentViewController:nil];
                [safController.view removeFromSuperview];
                [safController removeFromParentViewController];
                
                // Remove the window and release it's strong reference. This is important to ensure that
                // applications using view controller based status bar appearance are restored.
                [self.secondWindow removeFromSuperview];
                self.secondWindow = nil;
                
                [BNCPreferenceHelper preferenceHelper].lastStrongMatchDate = [NSDate date];
                self.requestInProgress = NO;
});
```
I hate hacks like this.

joshuafeldman: I think the best way would be to wire up the delegate to ensure the initial load. I am sure we can write a macro that wraps the delegate code if the safari class does not exist. Which would avoid your concerns.

joshuafeldman: Also I wonder if I can figure out a way to have the second window lower than the key window. I wonder if that will fix the status bar issue.

derrickstaten: Re delegate: I spent a while on that. You're welcome to attempt that, but I was unable to find a workable solution.

Re second window lower: if you figure that out, great. We haven't yet seen these issues, so it's hard for us to know what will fix them. I really appreciate you helping with this.

derrickstaten:
A flicker is obviously less than ideal. We'll spend some time trying to recreate. Can you give us a high level overview of how you build your initial window / view controller stack?

Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
joshuafeldman: Actually the dispatch is totally fine if I correct the status bar stuff. Let me try another commit.

PR #269, opened 2016-01-04
wallet pass functionality

aaustin:
@danwalkowski @derrickstaten I added the wallet pass scan functionality for our test with UA. I need to test and confirm.


danwalkowski:
Cool, thanks!  I had similar code, but was waiting on a test server from Hubert.  I suspect it feel off his radar during the break.

derrickstaten:
@danwalkowski I'll let you test this

Issue #274, opened 2016-01-12
Is it possible to use library in extensions?


lolgear:
Is it possible to use library in App extensions?

aaustin:
@lolgear Unfortunately, there will be a number of build errors if so. However, if it's possible to modify the core library for extension use, we'd welcome the contribution. Realistically, we won't be able to dive in for a month or two on that project.

aaustin:
We're happy to make ourselves available for questions if you want to tackle it!

aaustin:
Just a note that the library is now fully compatible with extensions! Here's some docs on how to add it to iMessage https://dev.branch.io/getting-started/sdk-integration-guide/guide/ios_imessage/

PR #122, opened 2015-06-04
Promo Arigato

shortstuffsushi:
@derrickstaten @dmitrig01 @aaustin @qinweigong 

Rename referral codes to promo codes, to better indicate the purpose they serve.

PR #276, opened 2016-01-13
App extensions support

medvedNick:
Now you are not able to link the Branch for app extension with Cocoapods. While compiling Branch pod there arise multiple errors about "[UIApplication sharedApplication] method not available in app extensions" and a couple of other ones. There is a workaround using (probably) a bug from Apple - you can build Branch as a static library by yourself and manually link it with your project. This will magically turn off all the errors. Also, [UIApplication sharedApplication] will actually work and return non-nil object.

In this pull-request I've added a Keyboard extension as an example of iOS app extension. It has all Branch code as Compile source (see above why I can not simply link it with current libBranch library).

I've added some conditional compilation checks for all unavailable methods. For extensions, these parts of the Branch code do not make sense (like adding gesture recognizers for main window), so we can simply exclude them.

This should solve #274


aaustin:
Nice!!! Can we send you a sweatshirt? Just put the referred app in as 'Branch SDK' https://branch.io/branch-referral-program-claim-reward/

We'll get this reviewed in the next few days and look into resolving conflicts.

medvedNick:
A few additions:

1) I think I have found a much more simple way to link Branch with the app extension, and it does not require code changes. You can simply set 'Require Only App-Extension-Safe API' to 'NO' in 'Build Settings' of the project. This will turn off the link errors like "[UIApplication sharedApplication] method not available in app extensions".

2) However, if you do not want to turn off this setting, you can use this pull request. Just put these strings into the Podfile:

`post_install do |installer|
	installer.pods_project.targets.each do |target|
		target.build_configurations.each do |config|
			config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] ||= ['$(inherited)', 'BRANCH_EXTENSION=1']
		end
	end
end`

This will manually remove unsupported extension code for Branch

Zhukn1:
@medvedNick, really cool man!😀 🍻

laptobbe:
Hi, I have the same problem. All I want to do is generate a link in the app extension, not use all the deep link functionality. Maybe it's possible to create a separate subspec for link creation and everything else? 

This is what Bolts did [#122](https://github.com/BoltsFramework/Bolts-ObjC/issues/122). This might be cleaner then using #ifdef, what do you think? If you think it's a good idea I will try to create a PR.

aaustin:
Note for those following this: The library is now fully extensions compatible. If you're building an iMessage extension, check out the docs here https://dev.branch.io/getting-started/sdk-integration-guide/guide/ios_imessage/

Closing this PR.

PR #278, opened 2016-01-14
Moved share text before URL.

YasKuraishi:
Fixed an issue where share text was being inserted after share URL.


derrickstaten:
Great catch. Feel free to use this in your own version of the SDK, and we'll include this in the next release (most likely on Tuesday in two weeks).

:+1:

YasKuraishi:
cool, thanks @derrickstaten

francisjervis:
I noticed this too, thanks for the fix!

PR #280, opened 2016-01-20
first pass - whitelist bundle ID of slack to prevent premature link c…

Sarkar:
…lick

@derrickstaten @danwalkowski 

we simply whitelist the bundle ID for the slack app when sharing links to slack. without this whitelist, a shared link to slack generates a click automatically without anyone doing anything. upon backgrounded and re-opening the app, deep link data appears (again, without a click).

this fix prevents that

### DIFF
--- a/Branch-SDK/Branch-SDK/BranchActivityItemProvider.m
+++ b/Branch-SDK/Branch-SDK/BranchActivityItemProvider.m
@@ -70,7 +70,7 @@
     }
     
     // Because Facebook immediately scrapes URLs, we add an additional parameter to the existing list, telling the backend to ignore the first click
-    if ([channel isEqualToString:@"facebook"] || [channel isEqualToString:@"twitter"]) {
+    if ([channel isEqualToString:@"facebook"] || [channel isEqualToString:@"twitter"]  || [channel isEqualToString:@"com.tinyspeck.chatlyio.share"]) {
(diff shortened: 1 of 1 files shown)


derrickstaten:
:+1: good catch

justjoeyuk:
I feel as if it might actually be more appropriate to ignore initial link clicks for any generated URL. This is because there are a massive number of options that the users can potentially share from (Messenger, Slack, Flowdock, etc) which can scrape the URLs if needed and cause problems within the app. The side effect of doing this would only affect the user doing the sharing, right? If not, it might still be worth the sacrifice of one click not deep linking than the app behaving erratic.

A feature that should be a must is exposing a "ignoreScrape" or "ignoreInitialClick" to the public API for those of us who are generating Short URLs manually for Facebook and Twitter.

@derrickstaten thoughts?

derrickstaten:
@justjoeyuk good news -- we already filter out the vast majority of bot clicks. We look at request headers, particularly user agent strings, when handling clicks on Branch links. This is the default behavior and not something that developers need to specify when creating links.

The problem is unique to iOS's SLComposeViewController/UIActivityViewController. When this was first implemented in iOS 6, the developers decided it was an okay idea to scrape the links with the exact same user agent string as if the user had just clicked the link in Safari. Since then, almost all scrapers of this sort have been better implemented and either (1) used a different user agent string, or (2) scraped from a different machine. So we've only seen this issue with Facebook, Twitter, etc. on iOS, and only when using SLComposeViewController/UIActivityViewController.

Please let us know if you have other questions about this.

PR #288, opened 2016-02-12
launchOptions are required

derrickstaten:
@aaustin @ahmednawar we should not have any versions of initSession that don't pass in launchOptions. This will only cause confusion. Our SDK does not function normally without launchOptions.

If anyone is using one of these methods and upgrades SDKs, they will need to switch to a more complete version of initSession. While I in general am highly opposed to forced upgrades, anyone currently using one of these methods has a buggy implementation of Branch.


aaustin:
:+1:

PR #294, opened 2016-02-21
fix match parameter sent to API

ahmednawar:
append the right match URL parameter sent to the API


ahmednawar:
@derrickstaten @Sarkar @aaustin

aaustin:
:+1:

Sarkar:
nice find! :+1:

PR #296, opened 2016-02-21
account for different anchors for share sheet on iPad

ahmednawar:
@aaustin @derrickstaten @Sarkar


aaustin:
Nice :+1:

galip91:
I'm showing the share sheet from an UIButton, not an UIBarButtonItem. Is it possible to add anchoring to other views?

derrickstaten:
@galip91 right now we have the anchor explicitly specified as a UIBarButtonItem. You could try anchoring from your UIButton (perhaps to test you could cast your UIButton to be a UIBarButtonItem), and if it works, we can make the anchor a more generic UIView. How does that sound?

PR #303, opened 2016-02-29
Carthage integration

hovox:


hovox:
Hi, any response on this, we are using carthage so it would be nice to integrate branch sdk with carthage, instead of having source code in our repository.

ahmednawar:
@hovox Thanks for sending the PR! Actually Carthage has been on our mind for some time and we actually have something ready for testing. I would be very interested in getting your feedback before we release.

I've a forked repo (https://github.com/ahmednawar/iOS-Deferred-Deep-Linking-SDK) with our Framework attached as a binary on a public release. Here are the instructions (https://github.com/ahmednawar/iOS-Deferred-Deep-Linking-SDK/blob/master/carthage.md) 

Thanks a lot. Appreciate it!

hovox:
@ahmednawar That's worked for us. I think it's nice. When do you plan to make official release with carthage support ?

Thanks.

hovox:
Is your build script support bitcode ? I mean if I import that framework with bitcode enabled project does it will work ? For us it's not critical because we didn't enabled bitcode.

ahmednawar:
@hovox Thanks for the feedback! I'll push a new release with Carthage support in a day or two. Also, we have the -fembed-bitcode flag in our build script.

Thanks again.

ahmednawar:
@hovox The new SDK is live with Carthage support. Let me know if you face any problems

PR #326, opened 2016-03-28
check for nsurlsession availability

ahmednawar:

Background: Our current code rely on compile-time macros to check for the availability with NSURLSession which was introduced in iOS 7. We can't rely on compile-time checks because the binary ships with one code path(NSURLSession) which fails on iOS 6. 

PR: changes the logic to rely on run-time checks. The code might be a bit confusing because it defines the completionhandler blocks before the request functions. I added some inline comments but not sure whether it is clear.

Tested on iOS 6, 7, 8, and 9

@aaustin @derrickstaten @Sarkar


aaustin:
Looks clean to me - thanks Ahmed :+1:

PR #336, opened 2016-04-16
Marshal's fix for session params

aaustin:
@pbcorriganBranch

Marshal’s fix on Cordova for the core SDK. We seem to only read from
defaults when the sessionParams are non-null, which is wrong.

I can’t even think of an edge case where this would happen, which is
probably why we’ve never encountered it.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
@@ -363,7 +363,7 @@ NSString * const BRANCH_PREFS_KEY_BRANCH_VIEW_USAGE_CNT = @"bnc_branch_view_usag
 }
 
 - (NSString *)sessionParams {
-    if (_sessionParams) {
+    if (!_sessionParams) {
         _sessionParams = [self readStringFromDefaults:BRANCH_PREFS_KEY_SESSION_PARAMS];
     }
     


Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
```diff
@@ -363,7 +363,7 @@ - (void)setUniversalLinkUrl:(NSString *)universalLinkUrl {
 }
 
 - (NSString *)sessionParams {
-    if (_sessionParams) {
+    if (!_sessionParams) {
```
ahmednawar: @aaustin doesn't this mean we won't read from it unless it's false?

aaustin: That is what it means. Session params are set while the app is running and the Branch session is active, so the only time they'll be false is if they were never set in the initSession callback. This would be if the app crashed or something. In this case, we should read them from shared prefs. Make sense?

ahmednawar: But if the app crashes, won't we create a new session and the session params will be overwritten? Probably I am confused because I am not clear why we save session params on disk. I need to dig deeper in the code to understand the session management part.

aaustin: It's a good point. This is a pretty pointless fix.

PR #337, opened 2016-04-19
BranchViewHandler Changes

pbcorriganBranch:
- Now only show the web view after a successful load of the page
- Do not show the web view if any http status code other than 200 is
received

### DIFF
--- a/Branch-SDK/Branch-SDK/BranchViewHandler.m
+++ b/Branch-SDK/Branch-SDK/BranchViewHandler.m
@@ -13,6 +13,10 @@
 #import "BranchView.h"
 
 @interface BranchViewHandler() <UIWebViewDelegate>
+
+@property (nonatomic, strong) BranchView *pendingBranchView;
+@property (nonatomic, strong) UIWebView *pendingWebview;
+
 @end
 
 NSString * const BRANCH_VIEW_REDIRECT_SCHEME = @"branch-cta";
@@ -42,7 +46,6 @@ NSString *currentBranchViewID;
     if ([branchView isAvailable]){
         self.branchViewCallback = callback;
         [self showView:branchView];
-        [branchView updateUsageCount];
         return YES;
     } else {
         return NO;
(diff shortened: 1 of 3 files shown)


pbcorriganBranch:
@sojanpr @ahmednawar @derrickstaten

Branch-SDK/Branch-SDK/BranchViewHandler.m
```diff
@@ -105,6 +107,28 @@ - (BOOL)webView:(UIWebView *)webView shouldStartLoadWithRequest:(NSURLRequest *)
...
+            UIViewController *presentingViewController = [[[[UIApplication sharedApplication] windows] firstObject] rootViewController];
+            [presentingViewController presentViewController:holderView animated:YES completion:nil];
+            
+            [self.pendingBranchView updateUsageCount];
+            
+            if (self.branchViewCallback) {
+                [self.branchViewCallback branchViewVisible:self.pendingBranchView.branchViewAction withID:self.pendingBranchView.branchViewID];
+            }
+        }
+    }
```
derrickstaten: it's not entirely clear to me how this works end to end, but the lack of an error handler worries me a bit. is there a case where the developer would want to know if status != 200?

pbcorriganBranch: Hi Derrick,

The objective, as explained to me, is to avoid having a blank white view
appear on the screen should the request result in any error.
Now, only if the page load is successful, will it be displayed.
Maybe Sojan can provide more detail.  I don't yet have enough knowledge
about the system to suggest what we may want to do to address errors.

Phillip

sojanpr: In case there is an error while loading the url , we should ,be calling back  `branchViewError` with an error code. Lets add `branchViewError` to `BranchViewControllerDelegate` and then call it whenever there is an error with the url loading

sojanpr:
LGTM 👍

Branch-SDK/Branch-SDK/BranchViewHandler.m
```diff
@@ -105,6 +108,33 @@ - (BOOL)webView:(UIWebView *)webView shouldStartLoadWithRequest:(NSURLRequest *)
     return !isRedirectHandled;
 }
 
+- (void)webViewDidFinishLoad:(UIWebView *)webView {
+    NSCachedURLResponse *resp = [[NSURLCache sharedURLCache] cachedResponseForRequest:webView.request];
+    NSInteger statusCode = [(NSHTTPURLResponse*)resp.response statusCode];
+    NSLog(@"Status: %ld",(long)statusCode);
```
aaustin: let's remove status

aaustin:
Other than logs removal 👍

ahmednawar:
@sojanpr nsnull change looks good!

PR #340, opened 2016-04-26
Subd support 1

ahmednawar:
@dmitrig01 @derrickstaten @aaustin 

cc @scott-branchio

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
@@ -77,7 +77,6 @@ NSInteger const ABOUT_30_DAYS_TIME_IN_SECONDS = 60 * 60 * 24 * 30;
         appDomainLinkURL = BNC_LINK_URL;
     }
     NSMutableString *urlString = [[NSMutableString alloc] initWithFormat:@"%@/_strong_match?os=%@", appDomainLinkURL, [BNCSystemObserver getOS]];
-    NSLog(@"strong match %@", urlString);
     
     BNCPreferenceHelper *preferenceHelper = [BNCPreferenceHelper preferenceHelper];
     BOOL isRealHardwareId;
--- a/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
@@ -67,7 +67,17 @@ NSInteger const ABOUT_30_DAYS_TIME_IN_SECONDS = 60 * 60 * 24 * 30;
 }
 
 - (void)presentSafariVCWithBranchKey:(NSString *)branchKey {
-    NSMutableString *urlString = [[NSMutableString alloc] initWithFormat:@"%@/_strong_match?os=%@", BNC_LINK_URL, [BNCSystemObserver getOS]];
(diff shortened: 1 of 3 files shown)


Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
```diff
@@ -67,7 +67,16 @@ - (void)createStrongMatchWithBranchKey:(NSString *)branchKey {
 }
 
 - (void)presentSafariVCWithBranchKey:(NSString *)branchKey {
-    NSMutableString *urlString = [[NSMutableString alloc] initWithFormat:@"%@/_strong_match?os=%@", BNC_LINK_URL, [BNCSystemObserver getOS]];
+    
+    NSString *appDomainLinkURL;
+    id ret = [[[NSBundle mainBundle] infoDictionary] objectForKey:@"branch_app_domain"];
+    if (ret) {
+        if ([ret isKindOfClass:[NSString class]])
+            appDomainLinkURL = [NSString stringWithFormat:@"https://%@", ret, nil];
```
derrickstaten: don't need `nil` as far as i can tell

Branch-SDK/Branch-SDK/Requests/BranchShortUrlSyncRequest.m
```diff
@@ -84,7 +84,9 @@ - (NSString *)createLongUrlForUserUrl:(NSString *)userUrl {
 }
 
 + (NSString *)createLinkFromBranchKey:(NSString *)branchKey tags:(NSArray *)tags alias:(NSString *)alias type:(BranchLinkType)type matchDuration:(NSInteger)duration channel:(NSString *)channel feature:(NSString *)feature stage:(NSString *)stage params:(NSDictionary *)params {
-    NSMutableString *baseUrl = [[NSMutableString alloc] initWithFormat:@"%@/a/%@?", BNC_LINK_URL, branchKey];
+    BNCPreferenceHelper *preferenceHelper = [BNCPreferenceHelper preferenceHelper];
+    NSMutableString *baseUrl = [preferenceHelper.userUrl mutableCopy];
```
aaustin: This presumes the SDK has been initialized once and that userUrl has been filled. Why change?

ahmednawar: Because for non-bnc.lt accounts, this will create links in the form of bnc.lt/a/key_livexx/blah whereas the userUrl will have the correct domain. My change will fail if a session was never initialized and currently will return an empty link

aaustin: Downstream will survive with an empty link? That was my only concern.

ahmednawar: The 2 cases I test are: getShortUrl and showShareSheet in the testbed. 
case 1: it will return a null url and the error will return "The Internet connection appears to be offline." This is the current behavior since getShortUrl hits createLongUrlForUserUrl at some point. 

case 2: It will not attach a link to the result, so if there is a share text, it will only share text.

Do you have other cases in mind that I could test?

aaustin:
Looks good to me 👍

PR #346, opened 2016-05-11
Add CocoaPod podspec without IDFA

ziogaschr:
As a developer I would like to be able to install the SDK using CocoaPods and without IDFA.
In the docs you pin point correctly [how to achieve this](https://dev.branch.io/getting-started/sdk-integration-guide/guide/ios/#submitting-to-the-app-store).

I added a new podspec that any developer will be able to use for installing the SDK without IDFA. This will allow us to keep the SDK updated easily


ziogaschr:
This is not ready to be merged

PR #347, opened 2016-05-11
Add CocoaPods subpec that doesn't add AdSupport (IDFA)

ziogaschr:
As a developer I would like to be able to install the SDK using CocoaPods and without the AdSupport framework (IDFA).
In the docs you pin point correctly [how to achieve this](https://dev.branch.io/getting-started/sdk-integration-guide/guide/ios/#submitting-to-the-app-store).

I added a new subspec in the podspec that any developer will be able to use for installing the SDK without IDFA. This will allow us to keep the SDK updated easily.

For testing it you can include this line in a test project: `pod 'Branch/without-IDFA', git: 'https://github.com/ziogaschr/iOS-Deferred-Deep-Linking-SDK.git', :branch => 'cocoapod-without-idfa'`

p.s.: `idfa.source_files` at line 31 is needed as having only the `idfa. frameworks ` at line 33 is not sufficient.

### DIFF
--- a/Branch.podspec
+++ b/Branch.podspec
@@ -21,7 +21,15 @@ Use the Branch SDK (branch.io) to create and power the links that point back to 
   s.platform     = :ios, '6.0'
   s.requires_arc = true
 
-  s.source_files = "Branch-SDK/Branch-SDK/*.{h,m}", "Branch-SDK/Fabric/*.h", "Branch-SDK/Branch-SDK/Requests/*.{h,m}"
+  s.subspec 'Core' do |core|
+    core.source_files = "Branch-SDK/Branch-SDK/*.{h,m}", "Branch-SDK/Fabric/*.h", "Branch-SDK/Branch-SDK/Requests/*.{h,m}"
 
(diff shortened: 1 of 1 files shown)


ahmednawar:
@ziogaschr thanks for the PR! we are gonna take a look and test this on our end..Will let you know how it goes. Cheers!

Branch.podspec
```diff
@@ -21,7 +21,15 @@ Use the Branch SDK (branch.io) to create and power the links that point back to
   s.platform     = :ios, '6.0'
   s.requires_arc = true
 
-  s.source_files = "Branch-SDK/Branch-SDK/*.{h,m}", "Branch-SDK/Fabric/*.h", "Branch-SDK/Branch-SDK/Requests/*.{h,m}"
+  s.subspec 'Core' do |core|
```
aaustin: Won't this require us to change all integrated pod files to `pod 'Branch/Core'` or will `pod Branch` still work?

ziogaschr: It will continue working. I was thinking the same but I gave it a test

aaustin:
👍

PR #349, opened 2016-05-15
add sanity checks to strong_match request url

ahmednawar:
`NSURL` returns `nil` if the string is an invalid URL. If a partner uses a space in their app version, this would return `nil` for example. `SFSafariViewController` crashes if given a `nil` for a URL. This change encodes the request and ensure it fails gracefully if the URL is `nil` for any reason.

2 things worth checking that I am not sure how to check for: ensuring that an encoded request would work fine on the backend (the logs are fine) and implications on the request queue on the client when returning early. I set `self.requestInProgress = NO` but not sure if this is enough.

@aaustin @derrickstaten @scott-branchio


derrickstaten:
:+1:

PR #350, opened 2016-05-18
add metadata to every request

derrickstaten:
@ahmednawar @aaustin

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCServerInterface.m
+++ b/Branch-SDK/Branch-SDK/BNCServerInterface.m
@@ -214,6 +214,11 @@ void (^NSURLConnectionCompletionHandler) (NSURLResponse *response, NSData *respo
     fullParamDict[@"sdk"] = [NSString stringWithFormat:@"ios%@", SDK_VERSION];
     fullParamDict[@"retryNumber"] = @(retryNumber);
     
+    NSMutableDictionary *metadata = [[NSMutableDictionary alloc] init];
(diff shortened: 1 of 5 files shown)


Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -214,6 +214,11 @@ - (NSDictionary *)prepareParamDict:(NSDictionary *)params key:(NSString *)key re
     fullParamDict[@"sdk"] = [NSString stringWithFormat:@"ios%@", SDK_VERSION];
     fullParamDict[@"retryNumber"] = @(retryNumber);
     
+    NSMutableDictionary *metadata = [[NSMutableDictionary alloc] init];
+    [metadata addEntriesFromDictionary:self.preferenceHelper.requestMetadataDictionary];
+    [metadata addEntriesFromDictionary:fullParamDict[BRANCH_REQUEST_KEY_STATE]];
```
ahmednawar: @derrickstaten just to make sure, this line guarantees that users can't override default request dictionary, right?

aaustin:
looks fine 👍

Branch-SDK/Branch-SDK/BNCServerInterface.m
derrickstaten: if the user has specified values for the same key both in the preferenceHelper.requestMetadataDictionary and the individual event (`withState:`), the latter should overwrite the former

PR #352, opened 2016-05-20
Check Facebook App Links client side

aaustin:
This PR bring App Link checking client side. I only inserted it from
cold-start to prevent too much regular boot up latency, and I figure
that’s the use case.

@ahmednawar @derrickstaten  @sojanpr


ahmednawar:
@aaustin LGTM

side question: this implies that users interested in analytics for facebook install ads need to update their sdk, right?

aaustin:
Only if folks want to try the client side version, which people report is supposedly more accurate. I'll wait for Derrick to take a gander too.

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -461,6 +464,33 @@ - (void)handlePushNotification:(NSDictionary *) userInfo {
...
+}
+
+- (BOOL)checkFacebookAppLinks {
+    if (self.FBSDKAppLinkUtility) {
+        SEL fetchDeferredAppLink = NSSelectorFromString(@"fetchDeferredAppLink:");
+        
+        if ([self.FBSDKAppLinkUtility methodForSelector:fetchDeferredAppLink]) {
+            void (^__nullable completionBlock)(NSURL *appLink, NSError *error) = ^void(NSURL *__nullable appLink, NSError *__nullable error) {
+                self.preferenceHelper.shouldWaitForInit = NO;
+                [self handleDeepLink:appLink];
```
sojanpr: Can this appLink returned be an `NSNull` Object?

sojanpr:
LGTM except for the minor qn

Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
```diff
@@ -419,6 +421,16 @@ - (void)setUserUrl:(NSString *)userUrl {
     }
 }
 
+- (BOOL)checkedFacebookAppLinks {
+    _checkedFacebookAppLinks = [self readBoolFromDefaults:BRANCH_PREFS_KEY_CHECKED_FACEBOOK_APP_LINKS];
```
derrickstaten: is there a reason to save this to defaults?

Branch-SDK/Branch-SDK/Branch.h
```diff
@@ -382,10 +382,17 @@ typedef NS_ENUM(NSUInteger, BranchPromoCodeUsageType) {
  
  @param debugParams dictionary of keystrings/valuestrings that will be added to response 
  */
--(void) setDeepLinkDebugMode:(NSDictionary *)debugParams;
+-(void)setDeepLinkDebugMode:(NSDictionary *)debugParams;
```
derrickstaten: thank you!

Branch-SDK/Branch-SDK/BNCPreferenceHelper.h
```diff
@@ -31,8 +31,9 @@
 @property (assign, nonatomic) BOOL explicitlyRequestedReferrable;
 @property (assign, nonatomic) BOOL isReferrable;
 @property (assign, nonatomic) BOOL isDebug;
-@property (assign, nonatomic) BOOL isContinuingUserActivity;
+@property (assign, nonatomic) BOOL shouldWaitForInit;
```
derrickstaten: I disagree with this name change. It's not actually about waiting for init. With this variable name, the following code reads a bit silly:

https://github.com/BranchMetrics/ios-branch-deep-linking/blob/db641cc19c0c33dd63ede59c57ddfdac7562efc2/Branch-SDK/Branch-SDK/Branch.m#L344-L350

I'm not sure what a good name is, which is why I gave it such a horrible name to begin with

aaustin: How about `initWillBeCalledSoon`? In your referenced code snippet, there's no reason for that boolean there. That boolean is in the wrong place. Sorry, I didn't review that code. Just changed the variable name. That should look like this:
```
if (self.accountForFacebookSDK) {
    id activity = [[options objectForKey:UIApplicationLaunchOptionsUserActivityDictionaryKey] objectForKey:@"UIApplicationLaunchOptionsUserActivityKey"];
    if (activity && [activity isKindOfClass:[NSUserActivity class]]) {
        return [self continueUserActivity:activity];
    }
}
self.preferenceHelper.shouldWaitForInit = YES;```

Comment (aaustin) [inline Branch-SDK/Branch-SDK/BNCPreferenceHelper.h]: Updating the PR with this change.

File: Branch-SDK/Branch-SDK/BNCPreferenceHelper.m (hunk 2 above)
Comment (aaustin) [inline Branch-SDK/Branch-SDK/BNCPreferenceHelper.m]: Yes because the variable is not passed through from whence it was checked.

File: Branch-SDK/Branch-SDK/Branch.m (hunk 1 above)
Comment (aaustin) [inline Branch-SDK/Branch-SDK/Branch.m]: added a check inside handleDeepLink

**Comment by aaustin** [comment]
@derrickstaten Please re-review. Want to get this in QA tomorrow ideally cc @scott-branchio

File: Branch-SDK/Branch-SDK/BNCPreferenceHelper.h (hunk 4 above)
Comment (derrickstaten) [inline Branch-SDK/Branch-SDK/BNCPreferenceHelper.h]: Are you sure it's in the wrong place? I'm not sure this change is right. I'll review when I'm in front of a computer, likely tomorrow. Either way, this is my fault for a poorly named variable

Comment (derrickstaten) [inline Branch-SDK/Branch-SDK/BNCPreferenceHelper.h]: Okay, we can keep the rename, though I'm wary of repurposing the flag

File: Branch-SDK/Branch-SDK/Branch.m [hunk 5]
```diff
@@ -335,16 +335,20 @@ - (void)initSessionWithLaunchOptions:(NSDictionary *)options isReferrable:(BOOL)
...
         else if ([options objectForKey:UIApplicationLaunchOptionsUserActivityDictionaryKey]) {
-            self.preferenceHelper.isContinuingUserActivity = YES;
             if (self.accountForFacebookSDK) {
                 id activity = [[options objectForKey:UIApplicationLaunchOptionsUserActivityDictionaryKey] objectForKey:@"UIApplicationLaunchOptionsUserActivityKey"];
                 if (activity && [activity isKindOfClass:[NSUserActivity class]]) {
                     [self continueUserActivity:activity];
+                    return;
                 }
             }
+            self.preferenceHelper.shouldWaitForInit = YES;
```
derrickstaten: let's not move this line of code. I worked through a very large matrix of tests to make sure that I didn't break our init code. This would definitely break things.

Branch-SDK/Branch-SDK/BNCPreferenceHelper.h
aaustin: Yes. Search where the boolean is used. It's just meant to prevent init from being called before openUrl or continueUserActivity actually receive the URI/Universal Link path.

Branch-SDK/Branch-SDK/Branch.m
aaustin: Follow the code path if you're calling continueUserActivity directly. It gets set to NO at the end of the function and init is called.

derrickstaten: correct, and that was the intended functionality. instead you've added a return early and then moved this flag down. six of one, half dozen of another

derrickstaten:
I have not manually tested this with many scenarios, but it appears that it should work :+1:

aaustin:
closing and re-PRing into staging.

PR #353, opened 2016-05-22
Fix retry code

aaustin:
@ahmednawar @derrickstaten When the retry code executes due to a 500
error code, we fall through and callback the block. This should not
happen and would lead to callbacks being executed multiple times. I
believe this to be the root cause of the crash when the credit service
went down last night.


derrickstaten:
:+1:

ahmednawar:
👍

aaustin:
oops - making PR into staging. My bad.

Issue #332, opened 2016-04-06
Conflict with Fabric


dskatz22:
The Branch Pod includes Fabric.h which conflicts with the Fabric framework,  causing the compiler to error on Fabric.with([Crashlytics()]).  Any suggestions on how to resolve

ahmednawar:
@dskatz22 Thanks for bringing this to our attention. I'll look into this and update you with a solution.

dskatz22:
fwiw downgraded to 0.11.18, and is working

andiradulescu:
Happens to me also.

​*Compile Error: definition of 'Fabric' must be imported from module 'Branch.Fabric' before it is required*​
```[Nomad/AppDelegate.m:38:6]
    [Fabric with:@[[Crashlytics class]]];
     ^
I'm not entirely sure, but using @import Fabric; instead of #import <Fabric/Fabric.h> compiled fine.

**Comment by hamdullahshah** [comment]
import the fabric after the branch.
e.g 
`
#import <Branch/Branch.h> 
#import <Fabric/Fabric.h>
`

**Comment by ahmednawar** [comment]
@dskatz22  @andiradulescu I am unable to reproduce this on my end..I created a vanilla project with Pods and Fabric but didn't get any errors. Do you have any steps to reproduce this?

**Comment by andiradulescu** [comment]
@ahmednawar it seems that @hamdullahshah is right. In order to get the error, you need to import the headers like this:

`#import <Fabric/Fabric.h>`
`#import <Branch/Branch.h>`

**Comment by ahmednawar** [comment]
This is very weird..this is the beginning of my AppDelegate and it still compiles and runs..I also changed the order but didn't make any difference 

```
#import "AppDelegate.h"
#import <Fabric/Fabric.h>
#import <Branch/Branch.h>
#import <Crashlytics/Crashlytics.h>

@implementation AppDelegate


- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
    // Override point for customization after application launch.
    [Fabric with:@[[Crashlytics class]]];
    [[Branch getInstance] initSessionWithLaunchOptions:launchOptions andRegisterDeepLinkHandler:^(NSDictionary *params, NSError *error) {
        NSLog(@"params %@", params);
    }];

    return YES;
}
```

**Comment by andiradulescu** [comment]
I tried reproducing it again, and it needs to be exactly like this (mind the Crashlytics one):

```
[...]

PR #360, opened 2016-05-25
Fixes issue #332 by explicitly specifying path to header files

vinceis1337:
https://github.com/BranchMetrics/ios-branch-deep-linking/issues/332


aaustin:
Nice ! 👍 Hard for us to verify that this works but compiles smoothly. We'll verify that the cocoapod is unaffected as well prior to release.

PR #365, opened 2016-06-06
Add-Branch-TestBed-Swift

dwestgate:
- New Swift version of the TestBed
- Modeled after the Android TestBed app
- Identical Objective-C version available for push based on testing
- No Fabric integration


aaustin:
nice 👍

PR #371, opened 2016-06-15
Fix for Facebook calling openUrl from within the app

aaustin:
As reported in https://github.com/BranchMetrics/ios-branch-deep-linking/issues/370, Facebook is now calling openUrl any time from within the app to handle their new auth flow. This incorrectly will force Branch to call the callback block with the same current session parameters.

This change will prevent the callback block from executing if the session is already initialized. @ahmednawar @derrickstaten


derrickstaten:
👍

PR #372, opened 2016-06-15
Retry in poor network condition

aaustin:
In response to this issue https://github.com/BranchMetrics/ios-branch-deep-linking/issues/369, we should be trying when the phone is experiencing poor network conditions that make HTTP unreliable. While experimenting, I found that all of these cases return negative HTTP error codes.

@derrickstaten @ahmednawar


ahmednawar:
👍

PR #373, opened 2016-06-16
expose branch view header files in framework

ahmednawar:
make Branch view header files public in the framework since our umbrella header for building modules imports it


ahmednawar:
@aaustin @derrickstaten

aaustin:
👍

PR #374, opened 2016-06-17
expand third party universal links

derrickstaten:
@ahmednawar @aaustin 

Please review my code carefully to see if I've overlooked anything. Test instructions will be provided on the #ios Slack channel.

cc @wlindemann

### DIFF
--- a/Branch-SDK/Branch-SDK/Branch.h
+++ b/Branch-SDK/Branch-SDK/Branch.h
@@ -436,6 +436,8 @@ typedef NS_ENUM(NSUInteger, BranchPromoCodeUsageType) {
  */
 - (void)setRequestMetadataKey:(NSString *)key value:(NSObject *)value;
 
+- (void)setThirdPartyUniversalLinksRegex:(NSRegularExpression *)regex;
+
 #pragma mark - Session Item methods
 
 ///--------------------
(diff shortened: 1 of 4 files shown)


Branch-SDK/Branch-SDK/Branch.h
```diff
@@ -436,6 +436,8 @@ typedef NS_ENUM(NSUInteger, BranchPromoCodeUsageType) {
  */
 - (void)setRequestMetadataKey:(NSString *)key value:(NSObject *)value;
 
+- (void)setThirdPartyUniversalLinksRegex:(NSRegularExpression *)regex;
```
aaustin: Given that ESPs are a dime-a-dozen, I think we should make this an app config and return it in /v1/open.

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1347,6 +1355,31 @@ - (void)processNextQueueItem {
     }
 }
 
+- (void)getBranchUniversalLinkUrlForThirdPartyUrl {
```
aaustin: Can we move this and it's delegate to a separate class as well.

derrickstaten: Sounds good. I should probably add code to use the webview's user agent string as well

PR #375, opened 2016-06-17
Updates modulemaps to describe framework dependencies

corydolphin:


carthage-files/module.modulemap
```diff
@@ -1,6 +1,10 @@
 framework module Branch {
-  umbrella header "BranchSDK.h"
+    umbrella header "BranchSDK.h"
+    header "Branch.h"
```
ahmednawar: @corydolphin just curious. Why is the Branch header needed for Carthage?

PR #376, opened 2016-06-18
kill old umbrella header

ahmednawar:
Killed old umbrella header and moved all needed header files to `Branch.h`. This introduced a circular dependency between `Branch.h` and `BranchUniversalObject.h` because of the callback typedefs. I solved this by moving them to a new header file. Let me know if you think of a better approach.

@aaustin


aaustin:
This one looked scary, so I wanted enough time to test but all seems great 👍 Thanks for being careful.

PR #378, opened 2016-06-21
Send user agent

aaustin:
@derrickstaten as discussed


ahmednawar:
@aaustin was this branched from staging or master?

derrickstaten:
Before we send this on every request, can we time how much it slows down requests? If it adds < 10ms, I'm okay with it. If it adds 200ms, then we have to reconsider.

ahmednawar:
Can we cache it on `PreferenceHelper`?

aaustin:
good point - let me find a better way.

aaustin:
this was off of master @ahmednawar

ahmednawar:
Ok, I should merge `master` into `staging` after every release then push to Github. The release script updates `master` and pushes to Github directly. The 2 branches will out of sync for files like `BNCConfig` and `ChangeLog`

PR #379, opened 2016-06-22
Fix for error handling of 400 errors.

Zhukn1:
One of the partners reported that there were cases that the object returned for key `error` was not a string but another object which led to a crash.

This fix will fallback to the default error string in such case.

@ahmednawar, please review.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCServerInterface.m
+++ b/Branch-SDK/Branch-SDK/BNCServerInterface.m
@@ -118,10 +118,10 @@ void (^NSURLConnectionCompletionHandler) (NSURLResponse *response, NSData *respo
                 error = [NSError errorWithDomain:BNCErrorDomain code:BNCDuplicateResourceError userInfo:@{ NSLocalizedDescriptionKey: @"A resource with this identifier already exists" }];
             }
             else if (status >= 400) {
-                NSString *errorString = @"The request was invalid."
+                NSString *errorString = @"The request was invalid.";
                 
                 if ([serverResponse.data objectForKey:@"error"] && [[serverResponse.data objectForKey:@"error"] isKindOfClass:[NSString class]]) {
-                    errorString = [serverResponse.data objectForKey:@"error"]
+                    errorString = [serverResponse.data objectForKey:@"error"];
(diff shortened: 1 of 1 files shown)


ahmednawar:
@Zhukn1 how can I reproduce this?

Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -118,7 +118,11 @@ - (void)genericHTTPRequest:(NSURLRequest *)request retryNumber:(NSInteger)retryN
                 error = [NSError errorWithDomain:BNCErrorDomain code:BNCDuplicateResourceError userInfo:@{ NSLocalizedDescriptionKey: @"A resource with this identifier already exists" }];
             }
             else if (status >= 400) {
-                NSString *errorString = [serverResponse.data objectForKey:@"error"] ?: @"The request was invalid.";
+                NSString *errorString = @"The request was invalid."
```
ahmednawar: can you add a `;` here?

Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -118,7 +118,11 @@ - (void)genericHTTPRequest:(NSURLRequest *)request retryNumber:(NSInteger)retryN
...
+                
+                if ([serverResponse.data objectForKey:@"error"] && [[serverResponse.data objectForKey:@"error"] isKindOfClass:[NSString class]]) {
+                    errorString = [serverResponse.data objectForKey:@"error"]
```
ahmednawar: same here

Zhukn1:
@ahmednawar thanks for finding that 👍. That's extensive Swift time kicking in.

ahmednawar:
@Zhukn1 👍 

Would be great if we can know which requests had objects instead of strings in the response in case we need to fix something in the backend.

Zhukn1:
@ahmednawar I tried to figure that out but this was the lowest I could get. No other info from the partner, unfortunately. 

But as this line was in the crash reports I think we should apply the patch anyway.

PR #382, opened 2016-07-03
Update test bed

dwestgate:
Updated the Objective-C TestBed to have the same look & feel as Swift and Android versions

Fixed typo in the TestBed-Swift entitlements file


aaustin:
Nice! Looks good to me 👍

PR #389, opened 2016-07-13
Switch showShareSheet...andCallback to showShareSheet...completion

rt2zz:
I believe this is the recommended format per: I believe this is now the recommended format per https://github.com/BranchMetrics/ios-branch-deep-linking/blob/master/Branch-SDK/Branch-SDK/BranchUniversalObject.h#L54


aaustin:
thanks ! 👍

Issue #392, opened 2016-07-15
Building the Carthage project fails since the last commit


lvandal:
Building the Carthage project fails since the last commit:

> error: unable to read module map contents from '/Users/ahmed/Documents/git_repos/iOS-Deferred-Deep-Linking-SDK/carthage-files/module.modulemap': Error Domain=NSCocoaErrorDomain Code=260 "The file “module.modulemap” couldn’t be opened because there is no such file." UserInfo={NSFilePath=/Users/ahmed/Documents/git_repos/iOS-Deferred-Deep-Linking-SDK/carthage-files/module.modulemap, NSUnderlyingError=0x7fca15e0d820 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}
error: unable to read module map contents from '/Users/ahmed/Documents/git_repos/iOS-Deferred-Deep-Linking-SDK/carthage-files/module.modulemap': Error Domain=NSCocoaErrorDomain Code=260 "The file “module.modulemap” couldn’t be opened because there is no such file." UserInfo={NSFilePath=/Users/ahmed/Documents/git_repos/iOS-Deferred-Deep-Linking-SDK/carthage-files/module.modulemap, NSUnderlyingError=0x7fca15de7cd0 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}
A shell task (/usr/bin/xcrun xcodebuild -project /Users/lucvandal/Repositories/Screens 4/screens-4-ios/Vendor/Carthage/Checkouts/iOS-Deferred-Deep-Linking-SDK/carthage-files/BranchSDK.xcodeproj -scheme Branch iOS SDK Carthage -configuration Release -sdk iphoneos ONLY_ACTIVE_ARCH=NO BITCODE_GENERATION_MODE=bitcode CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES clean build) failed with exit code 65:
** CLEAN FAILED **

ahmednawar:
@lvandal Thanks for reporting! Will push a fix today

ahmednawar:
@lvandal I've updated the Carthage project. Could you please test and let me know how it goes? Cheers!

siemensikkema:
@ahmednawar checking out the latest release (0.12.4) still triggers this problem. I'll try using a later commit but could you please push a new release?

siemensikkema:
[...]

PR #395, opened 2016-07-25
Fix path to module.modulemap

siemensikkema:
This is the actual fix for #392


ahmednawar:
@siemensikkema Thanks a lot for the fix. Didn't notice that I was adding an absolute local path :( Will push this with the next release.

PR #398, opened 2016-07-31
Update test bed

dwestgate:
Updated both TestBed apps so that they use new app.link domains and no longer show Promo code functionality


ahmednawar:
closing this because david has another PR with a new testbed #423

PR #401, opened 2016-08-10
Add missing frameworks to Podspec

### LINKED ISSUE #400: Outdated pod?
Hi,

I noticed that the version of the sdk in cocoapods is outdated, for example, i could not find the method "applyPromoCode".

ziogaschr:
Hello,
first of all I think that Podspec is missing 2 frameworks that you advice to be used [here](https://dev.branch.io/getting-started/sdk-integration-guide/advanced/ios/).
Let me know if you've done this by intention.

Also I am not sure, but this might fix #400. We are not using the feature @rafaellage wants to use, so please check.

The reason we created this PR is because of the following flow:

1. User click on a bnc.lt link, while App is not installed in phone
2. App Store opens and user installs the App
3. User opens the App but no data are being received from Branch.io

Note: We are installing App using `pod 'Branch/without-IDFA'` which means without `AdSupport`. Which is causing trouble to your flow, although on the other hand you say that App will matched 100% sure with use of `SafariServices`.

### DIFF
--- a/Branch.podspec
+++ b/Branch.podspec
@@ -22,12 +22,12 @@ Use the Branch SDK (branch.io) to create and power the links that point back to 
   s.subspec 'Core' do |core|
     core.source_files = "Branch-SDK/Branch-SDK/*.{h,m}", "Branch-SDK/Branch-SDK/Requests/*.{h,m}", "Branch-SDK/Fabric/*.h"
     core.private_header_files = "Branch-SDK/Fabric/*.h"
-    core.frameworks = 'AdSupport', 'CoreTelephony', 'MobileCoreServices'
+    core.frameworks = 'AdSupport', 'CoreTelephony', 'CoreSpotlight', 'MobileCoreServices', 'SafariServices'
   end
 
   s.subspec 'without-IDFA' do |idfa|
     idfa.source_files = "Branch-SDK/Branch-SDK/*.{h,m}", "Branch-SDK/Branch-SDK/Requests/*.{h,m}", "Branch-SDK/Fabric/*.h"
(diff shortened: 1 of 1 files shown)


aaustin:
Hey @ziogaschr - Thanks so much for the modification. We really appreciate it. However, this won't solve your deferred deep linking issue. There must be something else wrong. Can you send some details to support@branch.io? The team can help you resolve them there.

Now, in relation to this pull request, we wanted to make `CoreSpotlight` and `SafariServices` optional for our users in case folks were concerned about framework bloat as they are definitely not required.

Here's what you get with each framework:
- `AdSupport.framework`: This let's Branch read the IDFA to do attribution. We've got a pool of hundreds of millions of deterministically matched cookie-IDFA pairs, that we will use to do deferred deep linking. If we can read the IDFA, we will use this pool to match a new user to a click using this pool.
- `SafariServices.framework`: Allow us to do deferred deep linking _from Safari browser only_ with 100% accuracy as we can match the device ID with the Branch cookie in Safari using the SafariViewController. 
- `CoreSpotlight.framework`: This is necessary if you use the SDK methods `listOnSpotlight` that will list your app content in the Spotlight search portal. Using Branch for this allows you to track how much traffic comes from this channel, and leverage your existing Branch deep link routing.

Ultimately, none of these are required for the basic service. It's more about how the developer wants to optimize their Branch integration, knowing that they're adding Apple's heavy frameworks to their build.

ziogaschr:
Thanks a lot @aaustin for your detailed explanation. All the above make sense.

Actually I had a bug in our code and the above packages are not needed anymore for our case.

What might help developers who want to use these features will be to add a note under CocoaPods installation at your README file, noting that they have to manually add these frameworks.

PR #405, opened 2016-08-13
better instrumentation

ahmednawar:
1. send device information on retries
2. move the startTime variable to the genericHTTPRequest method. This will better computer the brtt on retries as well

@aaustin cc @brian-springer @chanian


aaustin:
Looks good to me ! 👍

brian-springer:
👍

PR #406, opened 2016-08-13
add checksumming to the release procedure

ahmednawar:
having some integrity
@aaustin cc @abelfletcher @sojanpr


aaustin:
nice 👍

PR #411, opened 2016-08-23
Nullability and nonnull support for Swift

aaustin:
This PR adds the proper null/nonnull flags as requested in https://github.com/BranchMetrics/ios-branch-deep-linking/issues/368 @ahmednawar 

This definitely exposed a bit of messiness :(

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCContentDiscoveryManager.m
+++ b/Branch-SDK/Branch-SDK/BNCContentDiscoveryManager.m
@@ -6,6 +6,7 @@
 //  Copyright © 2015 Branch Metrics. All rights reserved.
 //
 
+#import "BNCPreferenceHelper.h"
 #import "BNCContentDiscoveryManager.h"
 #import "BNCSystemObserver.h"
 #import "BNCError.h"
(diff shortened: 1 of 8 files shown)


Branch-SDK/Branch-SDK/BNCContentDiscoveryManager.m
```diff
@@ -309,7 +310,7 @@ - (void)indexContentWithTitle:(NSString *)title
     if ([BNCSystemObserver getOSVersion].integerValue < 9) {
         NSError *error = [NSError errorWithDomain:BNCErrorDomain code:BNCVersionError userInfo:@{ NSLocalizedDescriptionKey: @"Cannot use CoreSpotlight indexing service prior to iOS 9" }];
         if (callback) {
-            callback(nil, error);
+            callback([BNCPreferenceHelper preferenceHelper].userUrl, error);
```
ahmednawar: @aaustin: just curious, why return the `userUrl` here?

aaustin: I wanted to maintain this callback as nonnull as is the primary use case for all getShortUrl calls, but there's not a handy way to get a long link. This is the next best thing. I didn't invest in the long link messiness because I'm sure this code path is rarely executed, if ever.

ahmednawar: make sense

ahmednawar:
may the optionals be happy now :+1

PR #417, opened 2016-08-31
iOS 10 iMessage support for deep linking and BUO

aaustin:
This makes the library compatible for extensions and iOS 10. Branch links already support iOS 10 iMessage App Store routing, so we're good to go there! @ahmednawar


ahmednawar:
👍  exciting stuff.

I am thinking of not merging until QA tests separately to minimize confusion.

PR #426, opened 2016-09-07
AIS-60: add nullability type annotations to BUO

ahmednawar:
@aaustin


aaustin:
Just so that it automatically gets updated, here is the related issue https://github.com/BranchMetrics/ios-branch-deep-linking/issues/427

aaustin:
👍

PR #428, opened 2016-09-08
Block other requests until init finished

aaustin:
Before this change, it was possible for other requests to be called before init was called. I believe this would cause some types of requests to fail and possibly crash., especially on some of the other platforms like Cordova, where the lifecycle management is quite wonky.


aaustin:
@ahmednawar

ahmednawar:
👍

aaustin:
Should resolve https://github.com/BranchMetrics/ios-branch-deep-linking/issues/424

PR #431, opened 2016-09-09
AIS-40: Brtt encoding

ahmednawar:
This handles the encoding errors I was getting for brtt. The instrumentation dictionary is mutable and this wasn't one of the classes handled in the encoding method.

@aaustin


aaustin:
Ahh nice find 👍

PR #433, opened 2016-09-09
AIS-76: iOS 7 fix

ahmednawar:
replace containsString with rangeOfString for iOS 7 support
@derrickstaten @aaustin


aaustin:
Ack thanks for the fix Ahmed :+1:

aaustin:
Fixes https://github.com/BranchMetrics/ios-branch-deep-linking/issues/432

PR #435, opened 2016-09-09
New onboarding option

derrickstaten:
@aaustin @ahmednawar

### DIFF
--- a/Branch-TestBed/Branch-TestBed/AppDelegate.m
+++ b/Branch-TestBed/Branch-TestBed/AppDelegate.m
@@ -62,7 +62,7 @@
             NSLog(@"Branch TestBed: Initialization failed\n%@", error.localizedDescription);
         }
     }];
-        
+
     return YES;
 }
 
--- a/Branch-TestBed/Branch-TestBed/AppDelegate.m
+++ b/Branch-TestBed/Branch-TestBed/AppDelegate.m
@@ -10,6 +10,11 @@
 #import "LogOutputViewController.h"
 #import "NavigationController.h"
 #import "ViewController.h"
+#import <SafariServices/SafariServices.h>
+
+@interface AppDelegate() <SFSafariViewControllerDelegate>
+@property (nonatomic, strong) SFSafariViewController *onboardingVC;
+@end
 
 @implementation AppDelegate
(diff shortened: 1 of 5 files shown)


Branch-TestBed/Branch-TestBed/AppDelegate.m
```diff
@@ -52,10 +62,45 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
+
+        self.onboardingVC = [[SFSafariViewController alloc] initWithURL:urlForOnboarding];
+        self.onboardingVC.delegate = self;
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [[[[UIApplication sharedApplication].delegate window] rootViewController] presentViewController:self.onboardingVC animated:YES completion:NULL];
+        });
+    }
+}
+
+- (void)safariViewControllerDidFinish:(SFSafariViewController *)controller {
```
aaustin: How about we implement the example to trigger init after the `updatedUrlForOnboarding` is finished loading and it's redirecting to the original URL?

aaustin:
Looks awesome! Let's slightly tweak the example so that it intercepts the redirect and triggers init then.

Branch-TestBed/Branch-TestBed/AppDelegate.m
derrickstaten: I thought about this yesterday when coding this up and realized that would be problematic. initSession would finish and we'd fire the callback with the SFSafariViewController onscreen. If deep link data comes back, then potentially we'd be trying to push a view controller while the SFSafariViewController is still on screen.

The downside of this example is that init is delayed until after the VC is dismissed when optimally this would occur while the VC is still onscreen. However, for the sake of simplicity, I think we should keep it like this. If partners want, they can implement a more complex setup that proceeds with init while the VC is still onscreen, and add more logic to the initSession callback that delays pushing any new VCs until after SFSafariViewController is dismissed. Given that everyone's view controller hierarchy is unique, I'd rather keep the example here as agnostic as possible to anyone's particular setup. Is that okay?

aaustin:
Sounds good :+1:

PR #437, opened 2016-09-10
Support separate keys for Fabric config

aaustin:
This is in response to https://github.com/BranchMetrics/ios-branch-deep-linking/issues/429. It supports the dictionary based key storage method in the Fabric config.

@derrickstaten


derrickstaten:
👍

thorntonbm:
It's probably a good idea to get the docs updated here to reflect this: https://www.fabric.io/kits/ios/branch/install

PR #438, opened 2016-09-10
Pass completion error through

aaustin:
This update will:
1. support passing the completion error through the share sheet callback block per the request in https://github.com/BranchMetrics/ios-branch-deep-linking/issues/436
2. fix the nullability label of the activityType as referenced in https://github.com/BranchMetrics/ios-branch-deep-linking/issues/436

@derrickstaten


derrickstaten:
looks good to me. even more completion handler options.. 👍

PR #441, opened 2016-09-12
change onboarding domain

aaustin:
@derrickstaten I'll make sure we comm this to the right folks. Don't merge in until I get a chance to do so.


derrickstaten:
hmm, this is also used for dynamically-created links. we'll need to make sure those work with this domain as well (aka does app.link/a/key_live... work?)

dmitrig01:
no, it does not work. they need to plug in their own subdomain for that...

aaustin:
this is no longer needed per the redirect we added on this domain.

PR #443, opened 2016-09-13
Clear timer on activation

aaustin:
This resolves a potential lifecycle edge case where applicationDidBecomeActive handler is called before the scheduled “callClose” happens (this timer is setup on applicationWillResignActive handler and it is scheduled to happen half a second after). In this case, the SDK will initialize and then be de-initialized when callClose triggers, causing any subsequent call to Branch’s methods will make the application to crash with the error: “Branch SDK Error: making request before init succeeded!".

We've been unable to reproduce this but logically, it is possible.

@derrickstaten

### DIFF
--- a/Branch-SDK/Branch-SDK/Branch.m
+++ b/Branch-SDK/Branch-SDK/Branch.m
@@ -1094,6 +1094,7 @@ NSString * const BNCShareCompletedEvent = @"Share Completed";
 #pragma mark - Application State Change methods
 
 - (void)applicationDidBecomeActive {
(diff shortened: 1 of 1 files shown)


derrickstaten:
Any changes to lifecycle make me nervous. I think this needs to be extensively tested by QA before pushing to master. It seems relatively innocuous but could do great harm if there are cases we aren't anticipating. 👍  pending rigorous QA.

aaustin:
Partner also confirmed this resolved his issue.

PR #444, opened 2016-09-13
Fix unknown credits crash

aaustin:
We have reports of a rare but unreproducible crash in `loadRewards`. The crash report indicated that there was a `Fatal Exception: NSInvalidArgumentException` on line 51. This pull request adds extensive type checking to ensure arguments are valid.

@agrimn @derrickstaten

### DIFF
--- a/Branch-SDK/Branch-SDK/Requests/BranchLoadRewardsRequest.m
+++ b/Branch-SDK/Branch-SDK/Requests/BranchLoadRewardsRequest.m
@@ -49,7 +49,7 @@
     if ([responseKeys count] && ([response.data isKindOfClass:[NSDictionary class]] || [response.data isKindOfClass:[NSMutableDictionary class]])) {
         
         for (NSString *key in response.data) {
-            if (!key) { continue; }
+            if (![key isKindOfClass:[NSString class]]) { continue; }
             
             NSInteger credits = [preferenceHelper getCreditCountForBucket:key];
             if (response.data[key] && [response.data[key] respondsToSelector:@selector(integerValue)]) {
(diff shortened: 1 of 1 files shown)


derrickstaten:
this looks fine to me. pending testing, 👍

Branch-SDK/Branch-SDK/Requests/BranchLoadRewardsRequest.m
```diff
@@ -46,18 +46,26 @@ - (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
...
     NSArray *storedKeys = [currentCreditDictionary allKeys];
 
-    if ([responseKeys count]) {
+    if ([responseKeys count] && ([response.data isKindOfClass:[NSDictionary class]] || [response.data isKindOfClass:[NSMutableDictionary class]])) {
+        
         for (NSString *key in response.data) {
-             NSInteger credits = [response.data[key] integerValue];
+            if (![key isKindOfClass:[NSString class]]) { continue; }
+            
+            NSInteger credits = [preferenceHelper getCreditCountForBucket:key];
```
derrickstaten: why did the `credits` variable change from being derived from the response to being pulled from preference helper?

derrickstaten: other than this, 👍

aaustin: that's harmless because it gets set in the pref helper down below

derrickstaten: got it. i guess this has no impact on `hasUpdated` which is all i cared about. certainly an edge case.

derrickstaten:
👍

PR #449, opened 2016-09-14
Crash Prevention For Request Fail

sojanpr:
There may a corner case where request in the  main queue is nullified
and may cause cars when trying to add to failed request queue

@aaustin


aaustin:
Can you re-open this against `staging`?

postmechanical:
Is there any plan to ship this fix?

PR #450, opened 2016-09-15
Fix BUO nullables

aaustin:
See https://github.com/BranchMetrics/ios-branch-deep-linking/issues/448. @derrickstaten


derrickstaten:
can you open against staging? other than that, the most recent commit looks good

PR #452, opened 2016-09-15
Crash Prevention For Request Fail

sojanpr:
Protection for  a corner case crash reported by partner. 

@aaustin


aaustin:
hmm - seems harmless 👍

PR #453, opened 2016-09-15
hotfix for facebook shouldWaitForInit functionality

derrickstaten:
@aaustin fix for bug where initSession would hang on cold launch if universal link tapped


aaustin:
👍

PR #463, opened 2016-09-21
Share Sheet Swift Crash

sojanpr:
Adding  NSInvalidArg pprotection.

Seems swift can return nil for UIActivityViewController  on initializing with place holder from issue reported below
https://github.com/BranchMetrics/ios-branch-deep-linking/issues/461
@aaustin @derrickstaten

### DIFF
--- a/Branch-SDK/Branch-SDK/BranchUniversalObject.m
+++ b/Branch-SDK/Branch-SDK/BranchUniversalObject.m
@@ -182,7 +182,12 @@
     // Log share initiated event
     [self userCompletedAction:BNCShareInitiatedEvent];
     UIActivityItemProvider *itemProvider = [self getBranchActivityItemWithLinkProperties:linkProperties];
-    NSMutableArray *items = [NSMutableArray arrayWithObject:itemProvider];
+    NSMutableArray *items;
+    if (itemProvider) {
+        items = [NSMutableArray arrayWithObject:itemProvider];
+    } else {
+        items = [[NSMutableArray alloc] init];
+    }
     if (shareText) {
         [items insertObject:shareText atIndex:0];
     }


derrickstaten:
Do we have a way to reproduce this crash? Other than that, this looks okay to me. 👍

Branch-SDK/Branch-SDK/BranchUniversalObject.m
```diff
@@ -182,7 +182,12 @@ - (void)showShareSheetWithLinkProperties:(BranchLinkProperties *)linkProperties
     // Log share initiated event
     [self userCompletedAction:BNCShareInitiatedEvent];
     UIActivityItemProvider *itemProvider = [self getBranchActivityItemWithLinkProperties:linkProperties];
-    NSMutableArray *items = [NSMutableArray arrayWithObject:itemProvider];
+    NSMutableArray *items;
+    if (itemProvider) {
```
aaustin: hmm - if this gets returned null, basically we won't be able to put a Branch link in there

derrickstaten: Good point. I would like to be able to reproduce this before making this change.

derrickstaten: *before making _any_ change

sojanpr: Couldn`t reproduce this with the test bed application. Setting a `itemProvider = nil` creates the exact same error log as mentioned in the ticket. Looking at the code there is no chance for this since it is returning the `UIActivityViewController`

E-B-Smith: I'm closing this PR since this doesn't seem to solve anything.   We can re-open this in the future if needed.

PR #473, opened 2016-09-30
add missing src files to XCode project

ahmednawar:
this fixes the build errors in the iMessage test-bed app. We forgot to include the new content discovery src files in the project. The SDK files are added as symbolic links to the original files but XCode can't see newly added files automatically.

@aaustin @E-B-Smith


aaustin:
👍

PR #477, opened 2016-10-04
do not send brtt info in GET requests

ahmednawar:
We are still getting encoding errors for brtt if we are sending GET requests because we don't support encoding dictionaries to query strings.

After a discussion with brian, we decided not to send brtt info in GET requests. Another error comes up because we also send a metadata dictionary in all requests even if it's empty which results in an encoding error as well. This PR changes that to only send metadata if it contains data.

@aaustin @derrickstaten @E-B-Smith

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCServerInterface.m
+++ b/Branch-SDK/Branch-SDK/BNCServerInterface.m
@@ -191,10 +191,8 @@ NSString *requestEndpoint;
 #pragma mark - Internals
 
 - (NSURLRequest *)prepareGetRequest:(NSDictionary *)params url:(NSString *)url key:(NSString *)key retryNumber:(NSInteger)retryNumber log:(BOOL)log {
-    NSMutableDictionary *preparedParams = [[self prepareParamDict:params key:key retryNumber:retryNumber] mutableCopy];
+    NSDictionary *preparedParams = [self prepareParamDict:params key:key retryNumber:retryNumber requestType:@"GET"];
     
-    // explicitly remove instrumentation info from GET requests. They are only meant to be sent in POST requests.
-    [preparedParams removeObjectForKey:BRANCH_REQUEST_KEY_INSTRUMENTATION];
     NSString *requestUrlString = [NSString stringWithFormat:@"%@%@", url, [BNCEncodingUtils encodeDictionaryToQueryString:preparedParams]];
     
     if (log) {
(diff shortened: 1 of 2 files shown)


Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -243,7 +245,9 @@ - (NSDictionary *)prepareParamDict:(NSDictionary *)params key:(NSString *)key re
     NSMutableDictionary *metadata = [[NSMutableDictionary alloc] init];
     [metadata addEntriesFromDictionary:self.preferenceHelper.requestMetadataDictionary];
     [metadata addEntriesFromDictionary:fullParamDict[BRANCH_REQUEST_KEY_STATE]];
-    fullParamDict[BRANCH_REQUEST_KEY_STATE] = metadata;
+    if (metadata.count) {
```
ahmednawar: any potential side effects for this check?

Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -191,8 +191,10 @@ - (BNCServerResponse *)genericHTTPRequest:(NSURLRequest *)request log:(BOOL)log
 #pragma mark - Internals
 
 - (NSURLRequest *)prepareGetRequest:(NSDictionary *)params url:(NSString *)url key:(NSString *)key retryNumber:(NSInteger)retryNumber log:(BOOL)log {
-    NSDictionary *preparedParams = [self prepareParamDict:params key:key retryNumber:retryNumber];
+    NSMutableDictionary *preparedParams = [[self prepareParamDict:params key:key retryNumber:retryNumber] mutableCopy];
     
+    // explicitly remove instrumentation info from GET requests. They are only meant to be sent in POST requests.
+    [preparedParams removeObjectForKey:BRANCH_REQUEST_KEY_INSTRUMENTATION];
```
aaustin: I'd prefer that you add this check in the below method, alongside this conditional `if (self.preferenceHelper.instrumentationDictionary.count) {`. I remember we had a history of concurrent modification issues with this type of stuff, which is where the structure of the `prepareParamDict` method came from.

ahmednawar: Ah ok, much cleaner for sure. I've updated the PR.

aaustin:
👍

PR #501, opened 2016-11-07
Clean up file presence warnings

aaustin:
In this new file design, we were spitting out a giant error message for preferences and the queue when the first time the SDK was installed. We should not be throwing errors or warnings unless there's a problem. @ahmednawar


ahmednawar:
but isn't the failure to load the prefs or queue files a problem that would cause the sdk to not function properly?

aaustin:
no - we save them later on but the first time, on install, they won't be there.

ahmednawar:
👍

PR #506, opened 2016-11-16
Queued Opens AIS-128

E-B-Smith:
This fixes the JIRA issue https://branch.atlassian.net/browse/AIS-128

This fix has been tested on iOS 10/9/8/7 and opens links as expected.

@aaustin you may want to look this PR over since it changes the way the network queue handles requests.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
@@ -609,38 +609,48 @@ static NSString * const BNC_BRANCH_FABRIC_APP_KEY_KEY = @"branch_key";
     return (NSDictionary *)[self readObjectFromDefaults:BRANCH_PREFS_KEY_ANALYTICS_MANIFEST];
 }
 
+
 #pragma mark - Writing To Persistence
 
+
 - (void)writeIntegerToDefaults:(NSString *)key value:(NSInteger)value {
-    self.persistenceDict[key] = @(value);
-    [self persistPrefsToDisk];
+    [self writeObjectToDefaults:key value:@(value)];
 }
 
 - (void)writeBoolToDefaults:(NSString *)key value:(BOOL)value {
-    self.persistenceDict[key] = @(value);
-    [self persistPrefsToDisk];
+    [self writeObjectToDefaults:key value:@(value)];
 }
 
 - (void)writeObjectToDefaults:(NSString *)key value:(NSObject *)value {
-    if (value) {
-        self.persistenceDict[key] = value;
+    @synchronized (self) {
+        if (value) {
+            self.persistenceDict[key] = value;
+        }
+        else {
+            [self.persistenceDict removeObjectForKey:key];
+        }
+        [self persistPrefsToDisk];
     }
-    else {
-        [self.persistenceDict removeObjectForKey:key];
-    }
-
-    [self persistPrefsToDisk];
 }
 
 - (void)persistPrefsToDisk {
     @synchronized (self) {
         if (!self.persistenceDict) return;
-        NSData *data = [NSKeyedArchiver archivedDataWithRootObject:self.persistenceDict];
+        NSData *data = nil;
+        @try {
+            data = [NSKeyedArchiver archivedDataWithRootObject:self.persistenceDict];
+        }
+        @catch (id exception) {
+            data = nil;
+            [self logWarning:
+                [NSString stringWithFormat:@"Exception creating preferences data: %@.",
+                    exception]];
+        }
+        if (!data) {
+            [self logWarning:@"Can't create preferences data."];
+            return;
+        }
         NSBlockOperation *newPersistOp = [NSBlockOperation blockOperationWithBlock:^ {
-            if (!data) {
-                [self logWarning:@"Can't create preferences archive."];
-                return;
-            }
             NSError *error = nil;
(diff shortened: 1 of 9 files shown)


Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1268,11 +1264,7 @@ - (void)processNextQueueItem {
                 [req processResponse:nil error:[NSError errorWithDomain:BNCErrorDomain code:BNCInitError userInfo:@{ NSLocalizedDescriptionKey: @"Branch User Session has not been initialized" }]];
                 return;
             }
-            
-            if (![req isKindOfClass:[BranchCloseRequest class]]) {
-                [self clearTimer];
-            }
-            
+
```
aaustin: Won't we need to use that boolean `appIsInBackground` to prevent this code from executing? Here's what I'm imagining:
context: init session network call in progress, 3 items in queue
0. app goes to background
1. new boolean set
2. init session network call returns
3. incorrect: 3 items in queue proceed to execute despite the fact that the app is in the background
expected: 3 items remain outstanding in queue until next session initializes

Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
```diff
@@ -631,13 +634,9 @@ - (void)writeObjectToDefaults:(NSString *)key value:(NSObject *)value {
 
 - (void)persistPrefsToDisk {
     @synchronized (self) {
-        NSDictionary *persistenceDict = [self.persistenceDict copy];
+        if (!self.persistenceDict) return;
```
E-B-Smith: I didn't check the commit history.  But the `NSData *data = [NSKeyedArchiver archivedDataWithRootObject:self.persistenceDict];` on the next line effectively does a deep copy of the dict and it's data.

The `[self.persistenceDict copy]` copy doesn't necessarily do a deep copy which can lead to a crash later.

E-B-Smith:
I've tested on iOS 10/9/8/7 on real devices, with and without the network conditioner, and so far the user experience for switching app context and clicking links seems to be appropriate.

I think this PR needs more testing for sure, but I want to confirm I'm headed in the right path before committing time and testing in the wrong direction.

Branch-SDK/Branch-SDK/Branch.m
E-B-Smith: My perceived experience is that when the app transitions to/from background, the main queue is paused so nothing executes there, but other work queues may fire a callback when starting/stopping.  So things don't run indefinitely in the background and the network queue isn't going to empty in the background.

But honest question:  Is it important that these events strictly happen after the next successful  initSession?

Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
aaustin: You can click this `copy]` line in Github while reviewing the master branch code in the browser and then scroll up and press the 'Blame' button. It'll show you exactly who committed this with hopefully some explanation of why.

aaustin:
Let me know when you're ready. I'd like to do a round of network tests as well

E-B-Smith:
@ahmednawar  Alex: Ahmed is going to test this out as too.

ahmednawar:
@E-B-Smith @aaustin I've performed some tests and they all seem to pass except for one:
Enable the network link conditioner with an Edge profile --> click link A --> background app quickly --> click link B --> background app quickly --> open app w/o clicking on a link.

I expect the session to initialize and not return any link data, however, it consistently returns link B data. I am not sure why but in the logs I usually see 4 attempts to replay requests (3 for `v1/close` and 1 for `v1/open` with link B's universal link url)

E-B-Smith:
@ahmednawar @aaustin Thanks for testing!   I agree that in this case no link data should be returned.  I'll look at this and make an update.

E-B-Smith:
@ahmednawar At the moment, a request that is being retried is not cancellable, which is how you found the issue in your above scenario.

In the future we can refactor the server request queue to make cancelling requests (and other things) possible. But that refactor is going to be a big effort, and this PR is a good intermediate fix for the problem we are seeing.

Branch-SDK/Branch-SDK/Branch.m
aaustin: Going through this PR again and it seems like you didn't address this comment? Are you still working on this?

Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
aaustin: Can you at least wrap this in a try/catch just in case? I looked through past commits and can't remember when we added it but I'm having terrible flashbacks to the late night crash reports coming in ~ a year ago for concurrent modification errors.

E-B-Smith: I wrapped this in a try/catch.

aaustin:
In case it was missed above, I think there's still a risk where this will happen:
```
-1. Slower network. Init session and 3 items queued up immediately
0. Init session in flight due to slow network
1. app goes to background
2. new boolean set
3. init session network call returns
will happen: 3 items in queue proceed to execute despite the fact that the app is in the background
expected: 3 items remain outstanding in queue until next session initializes
```

Have we confirmed this isn't happening?

E-B-Smith:
@aaustin Yes, I confirmed that the three items remain in the queue until the next init session happens.

Issue #507, opened 2016-11-18
Headers need to be updated to include correct nullable and nonnull for Swift


jackfreeman:
Right now none of the headers use the `nullable` and `nonnull` keywords which can cause unexpected results when using the Branch SDK with Swift. See https://developer.apple.com/swift/blog/?id=25 for more info on these keywords and their usage.

Specifically I'm seeing a crash from `initSessionWithLaunchOptions` if the Branch API cannot be reached. Due to not having the `nullable` keyword, the params are returned as a force unwrapped optional, even though they are sometimes nil, which causes a crash when accessing them.

`callbackWithParams` in Branch.h should look like this to correctly integrate with Swift:
          
    typedef void (^callbackWithParams) (nullable NSDictionary *params, nullable NSError *error);

E-B-Smith:
Great, Thanks!  I'll add this to our next release.

aaustin:
Hey @jackfreeman - Not sure how you're seeing that? If you check out `BNCCallbacks.h` https://github.com/BranchMetrics/ios-branch-deep-linking/blob/master/Branch-SDK/Branch-SDK/BNCCallbacks.h, you'll see the nullable flags there. Not sure where you're observing otherwise. Can you help us understand?

shortstuffsushi:
Hey guys, want to add notes to your updates here. The `callbackWithParams` and `callbackWithBranchUniversalObject` both mark their non-error properties as `_Nonnull`, but in actual use, they're frequently `nil`. I'm not writing an app in Swift, but this likely causes the crashes described above. I was actual coming here to note that you're calling "non-nullable" params with nil, but it seems this guy is running into it already.

Example:

[Successful init](https://github.com/BranchMetrics/ios-branch-deep-linking/blob/43d036927d6888bb8a6280329c990b0200145ecc/Branch-SDK/Branch-SDK/Branch.m#L1368) calls [`getLatestReferringBranchUniversalObject`](https://github.com/BranchMetrics/ios-branch-deep-linking/blob/43d036927d6888bb8a6280329c990b0200145ecc/Branch-SDK/Branch-SDK/Branch.m#L786-L792), which returns `nil` unless the open was actually triggered by a link click.

[...]

PR #508, opened 2016-11-18
Made callback params nullable.  Github #507

E-B-Smith:
I made the callback params nullable for Swift compatibility.


E-B-Smith:
Wrong branch.

PR #509, opened 2016-11-18
Updated nullability for Swift.

E-B-Smith:
I made the callback params nullable for Swift compatibility.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCCallbacks.h
+++ b/Branch-SDK/Branch-SDK/BNCCallbacks.h
@@ -11,11 +11,11 @@
 
 @class BranchUniversalObject, BranchLinkProperties;
 
-typedef void (^callbackWithParams) (NSDictionary * _Nonnull params, NSError * _Nullable error);
-typedef void (^callbackWithUrl) (NSString * _Nonnull url, NSError * _Nullable error);
+typedef void (^callbackWithParams) (NSDictionary * _Nullable params, NSError * _Nullable error);
+typedef void (^callbackWithUrl) (NSString * _Nullable url, NSError * _Nullable error);
 typedef void (^callbackWithStatus) (BOOL changed, NSError * _Nullable error);
 typedef void (^callbackWithList) (NSArray * _Nullable list, NSError * _Nullable error);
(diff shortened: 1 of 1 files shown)


aaustin:
Hmm - I don't think params can ever be null though, which is why I set them to `_Nonnull`. Can you share the line where they return null values?

aaustin:
Also, you need to tag someone if you want this reviewed. Maybe you didn't?

E-B-Smith:
Accidental nil values being passed to a call back can be tricky to detect, and are a certain crash on the Swift side, so I made the parameters nullable for the time being to prevent crashes.

ahmednawar:
IIUC, all server responses should hit this method https://github.com/BranchMetrics/ios-branch-deep-linking/blob/master/Branch-SDK/Branch-SDK/BNCEncodingUtils.m#L389-L415 which would return `{}` in case of `nil` or bad data

E-B-Smith:
I'm thinking of this, instance:

https://github.com/BranchMetrics/ios-branch-deep-linking/blob/master/Branch-SDK/Branch-SDK/Branch.m#L1302-L1304
```
        if (self.sessionInitWithParamsCallback) {
            self.sessionInitWithParamsCallback([self getLatestReferringParams], nil);
        }
```

If `[self getLatestReferringParams]` return nil for some reason, it's a crash in Swift.

aaustin:
I think it's impossible for that to happen? https://github.com/BranchMetrics/ios-branch-deep-linking/blob/master/Branch-SDK/Branch-SDK/Branch.m#L775 and https://github.com/BranchMetrics/ios-branch-deep-linking/blob/master/Branch-SDK/Branch-SDK/BNCEncodingUtils.m#L389. Since we already have those flags and the parter wrote in saying we don't, I think there's a different root cause at play here. Let's wait for him to reply to us.

E-B-Smith:
The best solution is to add strict nullability checking throughout the SDK which will allow the static analyzer to catch these cases for us.  Checking all the edge cases can be pretty time consuming for us humans but is easy for the analyzer.

Until then, I think we should change the 'nullable' flag here to prevent accidental partner crashes.  It's the fastest fix and prevents some pretty awful partner crashes.

PR #519, opened 2016-12-07
Fix keyboard hiding

vdugnist:
We start our application with active keyboard. After request to branch, our keyboard hides on calling `[self.secondWindow makeKeyWindow]` inside.
You can add SFSafariController as child view controller to prevent this.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
@@ -30,7 +30,6 @@ NSInteger const ABOUT_30_DAYS_TIME_IN_SECONDS = 60 * 60 * 24 * 30;
 
 @interface BNCStrongMatchHelper ()
 
-@property (strong, nonatomic) UIWindow *secondWindow;
 @property (assign, nonatomic) BOOL requestInProgress;
 @property (assign, nonatomic) BOOL shouldDelayInstallRequest;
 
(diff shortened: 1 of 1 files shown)


vdugnist:
In attached videos I close tutorial, tap on camera request, accept it.
Before: https://youtu.be/wuOrgTx9Frs
After: https://youtu.be/jE10DbVHTlo

E-B-Smith:
@vdugnist  Thank you for your patch!

We have a fix for this that is beta testing right now.
If you'd test our fix for this to make sure it fixes your problem it would be a huge help.

Here's the info for getting our beta code:
https://github.com/BranchMetrics/ios-branch-deep-linking/wiki/The-SDK-Beta-Project

vdugnist:
@E-B-Smith beta works well. Can I use it in production right now or it's better to use current stable version with my fix?

E-B-Smith:
This is the version that we will push to production provided no significant problems are found.

It's been tested on iOS 8/9/10.  I would love some feedback so I can release it soon!

vdugnist:
Ok, thx!

E-B-Smith:
Thanks : Please send me feedback either way!

vdugnist:
@E-B-Smith I've tried to run tests before PR, but it looks like they call outdated methods.
We'll test current beta and I'll give feedback in a few days.

E-B-Smith:
@vdugnist Great! Thanks.

PR #528, opened 2016-12-21
Fix for potential race condition

aeromusek:
Suggestion comes from https://mobilegrowth.org/discussions/thread/unity-branch-initsession-callback-never-fires-when-deep-linking/

Original notes:

> In my [Unity] app, probably due to how scenes are loaded, the Branch.initSession callback never fires when the app is launched from a branch link (handleDeepLink). After lengthy debugging, I found the following patch fixes the issue for me. Is this a legit fix that others might encounter?

### DIFF
--- a/Branch-SDK/Branch-SDK/Branch.m
+++ b/Branch-SDK/Branch-SDK/Branch.m
@@ -330,7 +330,16 @@ NSString * const BNCShareCompletedEvent = @"Share Completed";
 
 - (void)initSessionWithLaunchOptions:(NSDictionary *)options isReferrable:(BOOL)isReferrable explicitlyRequestedReferrable:(BOOL)explicitlyRequestedReferrable automaticallyDisplayController:(BOOL)automaticallyDisplayController {
     self.shouldAutomaticallyDeepLink = automaticallyDisplayController;
+    
+    // If the SDK is already initialized, this means that initSession is being called later in the app lifecycle
+    // and that the developer is expecting to receive deep link parameters via the callback block immediately
+    if (self.isInitialized) {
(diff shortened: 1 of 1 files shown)


dave-mcclurg:
I think there is probably a better solution.  If handleDeepLink tries to call the callback before it is setup by the app (via initSession), it should remember that (eg, callbackAttempted flag?) and then when initSession is called, it can check that flag and do the callback immediately.

aaustin:
@dave-mcclurg Just so I understand, the broken time sequence is as follows:
1. handleDeepLink or continueUserActivity receives the deep link URL
2. /v1/open is called with that deep link URL
3. parameters are returned to the app and stored in the ns user defaults
4. you then call initSession with an `options` dictionary filled with the URL and register your deep link handler
5. callback block doesn't trigger because initSession falls through

This is an edge case we've never supported as the normal iOS lifecycle never has `initSession` being called before `continueUserActivity` or `openUrl`.

If this is happening 100% of the time, can you just fetch the deep link parameters in your scene using our local nsuserdefaults method shown here? https://github.com/BranchMetrics/unity-branch-deep-linking#retrieve-session-install-or-open-parameters

I hesitate to open up support for this lifecycle flow since it's so unnatural, but let me think a bit more about this.

aaustin:
@dave-mcclurg Please read and comment on the above, but also note that I've changed the logic here. Would love to see if this new fix would work for you.

dave-mcclurg:
@aaustin - The patch works great. I’ll go with that.

aaustin:
Making against staging in this PR. Closing this out https://github.com/BranchMetrics/ios-branch-deep-linking/pull/535

PR #529, opened 2016-12-22
This patch fixes the iOS 10.2.2 app install/update reporting bug.

E-B-Smith:


E-B-Smith:
Tested with iOS 7/8/9/10 and 10.2.2.

PR #531, opened 2016-12-28
INTENG-3110 Revert lifecycle handling change in 0.12.18

aaustin:
@ahmednawar 

An explanation of the new logic Ed introduced in 0.12.18: Only set initSuccess when the app is in the foreground

Whereas the old logic set initSuccess any time init successfully completed.

The new logic fails on link click, where continueUserActivity will complete initialization before the app can enter the foreground, causing Branch to re-initialize unnecessarily, and clearing the existing referral data. This PR reverts the above logic back to the old mechanism.


mcocaro:
please notify me when this is merged! This has fixed the issue of the cold start that I was having. Thanks!

ahmednawar:
👍

aaustin:
Oops - let me merge into staging instead.

PR #536, opened 2017-01-04
Synchronous refer params AIS-8

E-B-Smith:
I added the method `- (NSDictionary*) getLatestReferringParamsSynchronous` to the Branch shared object.

I created a companion branch, `synchronous-refer-params-AIS-8-test` that has extra testing code included in Branch-TestBed.

That branch should not be merged, but it's useful for testing this code.  That branch already includes this code.


aaustin:
Hmm - can you pull in the latest changes from staging into this branch? It seems out of date with my fixed from last week, making it hard to review

aaustin:
@ahmednawar When you have a spare cycle, can you help me take a look over this one too? Just worried I'll miss something.

Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -138,10 +138,9 @@ - (void)genericHTTPRequest:(NSURLRequest *)request retryNumber:(NSInteger)retryN
             }
             
         }
-        dispatch_async(dispatch_get_main_queue(), ^{
-            if (callback)
-                callback(serverResponse, error);
-        });
+		//	Don't call on the main queue since it might be blocked.
+		if (callback)
```
aaustin: weird tabbing on these lines

Branch-SDK/Branch-SDK/Requests/BranchOpenRequest.m
```diff
@@ -173,4 +176,52 @@ - (NSString *)getActionName {
...
+            openRequestWaitQueueIsSuspended = YES;
+            dispatch_suspend(openRequestWaitQueue);
+        }
+    }
+}
+
++ (void) waitForOpenResponseLock {
+    NSLog(@"Wait for openRequestWaitQueue.");
+    dispatch_sync(openRequestWaitQueue, ^ {
+        NSLog(@"Finished waitForOpenResponseLock");
```
aaustin: can you clean up these logs?

aaustin: something feels gross about the syntax of this empty block. Maybe we could put a return in front of dispatch_sync and a return in the block for readability.

Branch-SDK/Branch-SDK/Requests/BranchOpenRequest.m
```diff
@@ -173,4 +176,52 @@ - (NSString *)getActionName {
...
+    });
+}
+
++ (void) releaseOpenResponseLock {
+    @synchronized (self) {
```
aaustin: should we synchronize on the static queue or bool rather than the instance of `BranchOpenRequest`? Hypothetically, it seems like it's possible for multiple branch open objects to enter this block of code and potentially clash on the boolean, although I'm having trouble imagining the edge case that could cause this.

aaustin:
Can you add a test case for when this is called in a non-main thread as well being called from the main thread?

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1181,16 +1197,17 @@ - (void)registerViewWithParams:(NSDictionary *)params andCallback:(callbackWithP
...
     }
 }
 
 - (void)applicationWillResignActive {
-    [self clearTimer];
-    self.sessionTimer = [NSTimer scheduledTimerWithTimeInterval:0.5 target:self selector:@selector(callClose) userInfo:nil repeats:NO];
+    self.appIsInBackground = YES;
+    [self callClose];
     [self.requestQueue persistImmediately];
+    [BranchOpenRequest setWaitNeededForOpenResponseLock];
```
aaustin: I don't understand why we need to call this here.

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1336,38 +1347,38 @@ - (void)initializeSession {
...
 
-    if ([self.requestQueue removeInstallOrOpen])
-        self.networkCount = 0;
-    BranchOpenRequest *req = [[clazz alloc] initWithCallback:initSessionCallback];
-    [self insertRequestAtFront:req];
-    [self processNextQueueItem];
+	@synchronized (self) {
+		if ([self.requestQueue removeInstallOrOpen])
+			self.networkCount = 0;
+		[BranchOpenRequest setWaitNeededForOpenResponseLock];
```
aaustin: I also don't understand why we need to call this here if it's called up in initialize?

Branch-SDK/Branch-SDK/Branch.m
aaustin: Oh I think I see: because you increment it every time the open call completes.

Branch-SDK/Branch-SDK/Branch.m
aaustin: Now I understand I think. If you call this here though, seems like you can get rid of the call in initialize.

Branch-SDK/Branch-SDK/Requests/BranchOpenRequest.m
```diff
@@ -82,6 +82,7 @@ - (void)makeRequest:(BNCServerInterface *)serverInterface key:(NSString *)key ca
 
 - (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
     if (error) {
+        [BranchOpenRequest releaseOpenResponseLock];
```
aaustin: Does this cover the retry case? Where a request fails & returns a non-nil error, then triggers a retry call? Will the queue be re-incremented?

Branch-SDK/Branch-SDK/Requests/BranchOpenRequest.m
aaustin: Ah maybe because this is called on the static instance, it's fine.

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -94,6 +93,7 @@ @interface Branch() <BranchDeepLinkingControllerCompletionDelegate, FABKit>
 @property (assign, nonatomic) BOOL delayForAppleAds;
 @property (assign, nonatomic) BOOL searchAdsDebugMode;
 @property (strong, nonatomic) NSMutableArray *whiteListedSchemeList;
+@property (assign, nonatomic) BOOL appIsInBackground;
```
aaustin: I think this boolean is unused? right?

Branch-SDK/Branch-SDK/Requests/BranchOpenRequest.m
E-B-Smith: By the time `processResponse:error:` is called, the network call has exceeded the retries.

I want to unlock the lock so that the client app won't freeze.  

But if there's a network error, Branch isn't really initialized correctly, and `getLatestReferringParamsSynchronous` will return stale information.  Maybe the `getLatestReferringParamsSynchronous` should return a dictionary like:

```
{
    error = BNCErrorNetworkError
}
```

to indicate that there was a problem?

Branch-SDK/Branch-SDK/Branch.m
E-B-Smith: True.  Better synchronization of the initialization made this unnecessary.

Branch-SDK/Branch-SDK/Requests/BranchOpenRequest.m
E-B-Smith: I will clean up the logs with some debug logging.

With the debug logging the block I think the code won't look weird.

E-B-Smith:
@aaustin Wrote: 

> Can you add a test case for when this is called in a non-main thread as well being called from the main thread?

There are test cases in the `synchronous-refer-params-AIS-8-test` branch, including tests for deadlocks.

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -796,6 +797,20 @@ - (NSDictionary *)getLatestReferringParams {
     return origSessionParams;
 }
 
+- (NSDictionary*) getLatestReferringParamsSynchronous {
+    if (!self.isInitialized &&
+        !self.preferenceHelper.shouldWaitForInit &&
+        ![self.requestQueue containsInstallOrOpen]) {
+		NSLog(@"[Branch] Warning: getLatestReferringParamsSynchronous called before session init.");
+		NSLog(@"[Branch] Make sure Branch `initSessionWithLaunchOptions:`, `handleDeepLink:`, or "
```
ahmednawar: should we return an empty dictionary here better than making the app wait forever in case of user error?

Branch-SDK/Branch-SDK/Requests/BranchOpenRequest.m
```diff
@@ -173,4 +176,52 @@ - (NSString *)getActionName {
     return @"open";
 }
 
+
+#pragma - Open Response Lock Handling
+
+
+//	Instead of semaphores, the lock is handled by scheduled dispatch_queues.
```
ahmednawar: nitpick: is a tab needed between the `//` and the comment text? on a broader note, perhaps we could start thinking of having a linter for the sdk and decide on a style.

Branch-SDK/Branch-SDK/BNCDeviceInfo.m
```diff
@@ -86,22 +86,22 @@ - (id)init {
     static NSString* browserUserAgentString = nil;
 
     void (^setUpBrowserUserAgent)() = ^() {
-        browserUserAgentString =
-            [[[UIWebView alloc]
-              initWithFrame:CGRectZero]
-                stringByEvaluatingJavaScriptFromString:@"navigator.userAgent"];
+		if (!browserUserAgentString) {
```
ahmednawar: This is out of scope for this PR, but I would propose checking this value every 30 days similar to what we do with strong matching instead with every request and cache it in the pref helper. AFAIU, this value will not change unless the user updates their OS. Allocating a `WebView` for every single request is a bit expensive.

Branch-SDK/Branch-SDK/Requests/BranchOpenRequest.m
E-B-Smith: By habit I use a tab.  I frequently see people use two spaces after the comment delimiter too.

On a broader note, I started an experimental SDK project that has a clang-format definition file.  The format file is called '.clang-format'.

The project is here:  https://github.com/BranchMetrics/BranchLabs-iOS

We can play with formatting there before deciding.  But I think having a formatter is a good idea.

Branch-SDK/Branch-SDK/BNCDeviceInfo.m
```diff
@@ -83,27 +83,60 @@ - (id)init {
...
 
-    return self;
+    static NSString* browserUserAgentString = nil;
+	void (^setBrowserUserAgent)() = ^() {
+		if (!browserUserAgentString) {
+			browserUserAgentString =
+				[[[UIWebView alloc]
+				  initWithFrame:CGRectZero]
+					stringByEvaluatingJavaScriptFromString:@"navigator.userAgent"];
+			NSLog(@"[Branch] userAgentString: '%@'.", browserUserAgentString);
```
aaustin: We don't want this log in there

Branch-SDK/Branch-SDK/BNCDeviceInfo.m
```diff
@@ -83,27 +83,60 @@ - (id)init {
...
+		return browserUserAgentString;
+	}
+
+	//	Wait and yield to prevent deadlock:
+
+	dispatch_block_t agentBlock = dispatch_block_create_with_qos_class(
+		DISPATCH_BLOCK_DETACHED | DISPATCH_BLOCK_ENFORCE_QOS_CLASS,
+		QOS_CLASS_USER_INTERACTIVE,
+		0,  ^ {
+			NSLog(@"Will do.");
```
aaustin: or these logs - this and 2 below

Branch-SDK/Branch-SDK/BNCServerInterface.m
aaustin: Did you see this comment?

Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -138,10 +138,9 @@ - (void)genericHTTPRequest:(NSURLRequest *)request retryNumber:(NSInteger)retryN
             }
             
         }
-        dispatch_async(dispatch_get_main_queue(), ^{
-            if (callback)
-                callback(serverResponse, error);
-        });
+		//	Don't call on the main queue since it might be blocked.
```
aaustin: This callback block has traditionally always been handled on the main thread. Changing this could break a lot of existing integrations that might assume we're always calling their code on the main thread.

aaustin: Oh - I see you've moved it further down into `initSessionCallback`. Will delete this comment.

Branch-SDK/Branch-SDK/Branch.m
aaustin: Please confirm this ^

Branch-SDK/Branch-SDK/Requests/BranchOpenRequest.m
```diff
@@ -173,4 +177,53 @@ - (NSString *)getActionName {
...
+    if (self != [BranchOpenRequest self])
+        return;
+    openRequestWaitQueue =
+        dispatch_queue_create("io.branch.sdk.openqueue", DISPATCH_QUEUE_CONCURRENT);
+}
+
++ (void) setWaitNeededForOpenResponseLock {
+    @synchronized (self) {
+        if (!openRequestWaitQueueIsSuspended) {
+            NSLog(@"Suspend openRequestWaitQueue.");
```
aaustin: Still a lot of logs in the code

Branch-SDK/Branch-SDK/Requests/BranchOpenRequest.m
aaustin: Won't the call to self.callback(NO, error) on line 88 passthrough that network error? I believe it will but you should confirm while testing with no network & network attenuation

aaustin:
Hey - you can review my outstanding comments by clicking the "Files changed" tab https://github.com/BranchMetrics/ios-branch-deep-linking/pull/536/files. Let's address each one before proceeding.

Lastly, can you update the README with the new API?

Branch-SDK/Branch-SDK/BNCDeviceInfo.m
E-B-Smith: I'll remove it after the final QA round.  It'll be much easier to see if this is breaking with the logging in.

Branch-SDK/Branch-SDK/BNCServerInterface.m
E-B-Smith: Would you prefer it all on one line?

Branch-SDK/Branch-SDK/Requests/BranchOpenRequest.m
E-B-Smith: Again, I'll remove them after the final QA pass.

Branch-SDK/Branch-SDK/Branch.m
E-B-Smith: The call to `setWaitNeededForOpenResponseLock` in initialize sets the lock because there are no `latestReferringParams` to return yet.

Later I set it because there will be some `latestReferringParams` to return, but we haven't received them yet.

Calling `setWaitNeededForOpenResponseLock` twice does no harm.

Branch-SDK/Branch-SDK/Requests/BranchOpenRequest.m
E-B-Smith: I'll test for it.

Branch-SDK/Branch-SDK/BNCServerInterface.m
aaustin: No it's not that. See line 140? It's got 8 spaces whereas these two lines have 2 tabs. It's just a weird indentation thing for these 2 lines alone.

E-B-Smith: Ah.  I must have an editor that is converting spaces to tabs or vice versa.  I can fix that.

aaustin:
Ok I think I'm comfortable for this to move to QA. Please please please please please remember to move the logs. There's nothing worse than an SDK that print debug logs in production apps. 👍

PR #538, opened 2017-01-04
Warn when a user event commerce conflicts DANA-77

E-B-Smith:
Added a warning when a user event conflicts with a Branch commerce event.


derrickstaten:
maybe show the full ios method name? `sendCommerceEvent:metadata:withCompletion:`

derrickstaten:
your call ultimately. then 👍

E-B-Smith:
@derrickstaten Good catch.  I updated the message to mention the `sendCommerceEvent:metadata:withCompletion:` method.

PR #539, opened 2017-01-05
Fixed a crash when the root view controller is UINavigationController…

n8tr:
Here is what we observed:
- Upon first launch of the app after install, if a user taps on our sign up button within 3 seconds, this causes a set of events that quickly lead to a crash, with the underlying cause being the BNCMatchView.

Repro steps:
-Fresh install the Clarity Money App
- Launch the app
- Tap on Sign up within 3 seconds
- No sign up screen gets displayed
- A back button appears at the top
- Tapping the back button pops off the welcome screen and user is left with a 100% black screen
- Tapping on the Sign Up button again still doesn't display anything differently
- After tapping on Sign up a second time, if press back again, app crashes.

Detailed explanation of problem:
- It appears that Branch inserts a view controller instance into the view hierarchy (via BNCMatchView)
- BNCMatchViewController exists in the navigation stack, but as a child view controller of the navigation controller, not a child VC of a VC pushed on the navigation controller stack
- BNCMatchView is inserted in the view hierarchy outside of the navigation stack
- If a user taps to push a new VC on the nav stack before BNCMatchView removes itself, the navigation controller tries to push the sign up view controller
- While that sign up VC is trying to animated on the screen, BNCMatchView and BNCMatchViewController removes itself
- In removing itself, it breaks the navigation controller animation and/or the view hierarchy in that the pushed sign up VC never gets added or gets quickly removed from the view hierarchy. 
- The pushed sign up VC does still exist in memory, but not never displays in the view hierarchy

Solution:
- Inspect window's root view controller for navigation controller, tab bar, or presented view controller and fetch that class's currently presented VC and then let BNCView and BNCMatchViewController attach itself there.

Welcome Screen
welcome

If the user waits for more than 3 seconds -OR- launches again after a crash, this view is properly pushed on the nav stack
whatshouldhappen

Welcome screen when tap on sign up within 3 seconds.  Notice no pushed sign up screen and notice the back button at top
welcomeaftertapsignuptoosoon

Flow where user can get black screen:
blackscreen

View hierarchy capture on cold launch within 3 seconds
coldlaunch_

View hierarchy on second launch or if wait more than 3 seconds:
2ndlaunch_

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
@@ -282,11 +282,13 @@
     self.matchView.alpha = 1.0;
     [self.matchView addSubview:self.matchViewController.view];
 
-    [self.primaryWindow.rootViewController addChildViewController:self.matchViewController];
-    UIView *parentView = self.primaryWindow.rootViewController.view ?: self.primaryWindow;
+    UIViewController *rootViewController = [self topViewController:self.primaryWindow.rootViewController];
+
+    [rootViewController addChildViewController:self.matchViewController];
+    UIView *parentView = rootViewController.view ?: self.primaryWindow;
     [parentView insertSubview:self.matchView atIndex:0];
 
-    [self.matchViewController didMoveToParentViewController:self.primaryWindow.rootViewController];
+    [self.matchViewController didMoveToParentViewController:rootViewController];
 
     return YES;
 }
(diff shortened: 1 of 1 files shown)


E-B-Smith:
Fantastic analysis and patch.  Thank you.  
I'll QA this and make a release today.

n8tr:
great thank you

E-B-Smith:
FYI:  The new version of Branch is released with this fix in it.  Thanks again.

n8tr:
great thanks for merging the fix

JakeSc:
We're also seeing this crash in our app. Confirmed that 0.12.24 fixes the bug.

PR #550, opened 2017-01-16
Add Commerce Example to TestBed-Swift AIS-177

E-B-Smith:


derrickstaten:
can you change currency to USD in all examples? Our backend only accepts valid currencies. We may even add these as an enum to the SDK

E-B-Smith:
Great.  I changed 'Smackeroos' to 'USD'.

Issue #548, opened 2017-01-14
Crash on BNCServerInterface.m line 158 -[BNCServerInterface genericHTTPRequest:retryNumber:log:callback:retryHandler:]


itchingpixels:
Hi,

We've seen 55 of these crashes coming from Branch in the last 7 days (10k sessions per day).

![branch-crash-1](https://cloud.githubusercontent.com/assets/5416288/21954635/ee895ec2-da4e-11e6-80fd-a529f0c02bc4.png)

itchingpixels:
http://crashes.to/s/117dea8947c

E-B-Smith:
Thanks for the heads up.  I'm looking into this now.

Which version of the SDK are you using?

itchingpixels:
hi!

thank you.
we're using the latest version ( - v0.12.24)

ahmednawar:
@itchingpixels I am unable to reproduce any of the crashes you've reported. Could you please send us your Branch-related code so that we can take a look? Feel free to send it to integrations@branch.io and reference the github issue in the email. Also we have a [Swift Testbed](https://github.com/BranchMetrics/ios-branch-deep-linking/tree/master/Branch-TestBed-Swift) that you can take a look at as a reference implementation. Please let us know.

itchingpixels:
Thank you. The testbed project looks exactly how we have Branch integrated.
Is there anything I can do? We're also seeing another crash that can be connected (affecting a lot more users - 2-3% of the sessions). It appeared at the same time as the others.
http://crashes.to/s/117dea8947c

We're investigating the latter - I'm not sure if it's connected.

We'd love to sort these out as soon as possible.

We started seeing these crashes after upgrading first to 0.12.19 (may be a good clue)

ahmednawar:
We had changes in this version related to `preferenceHelper`, however, we don't have any reports of crashes lately. 

1. What was the version prior to 0.12.19 that you were running?
2. Is it possible to share the code with us?

itchingpixels:
1. Actually, I may got myself confused regarding when this crash appeared. Let me investigate it a bit more closely, I'll get back to you tomorrow! 
2. Happy to share it (isn't really touching any of the business logic):
[branch-related-code.txt](https://github.com/BranchMetrics/ios-branch-deep-linking/files/709438/branch-related-code.txt)

itchingpixels:
Are you guys online now? I've got a crash in the simulator - debugger is paused currently:)

[...]

PR #554, opened 2017-01-18
Potentially fixed a crash casued by a prematurely dellocated static variable in BNCServerInterface.m

itchingpixels:
Details in #548


E-B-Smith:
Thank you for your patch! 👍

PR #555, opened 2017-01-20
Don't register iad attribution if not actually installed from search ad

djfink-carglass:
I believe that currently there is a bug with apple search ads integration. The SDK works correctly when the app is installed from a search ad, or when `requestAttributionDetailsWithBlock:` returns no dictionary. However, sometimes a dictionary is returned with the format:
```javascript
{
  "Version3.1": {
    "iad-attribution": "false"
  }
}
```
In this case, the SDK incorrectly reports that the install was from a search ad. This one line change should solve the problem.


dmitrig01:
we actually just solved this problem on the backend -- cc @derrickstaten

djfink-carglass:
great, I'll close the request

dmitrig01:
should be deployed around Wednesday of next week.

derrickstaten:
@djfink-carglass short term you could absolutely patch on your end. As @dmitrig01 said, we should have a fix in production on our backend with our next deploy.

djfink-carglass:
Sounds good, thanks for keeping me in the loop!

PR #558, opened 2017-01-21
INTENG-3137 Handle duplicate local attribution calls

aaustin:
There's a flaw with our local, async attribution checks where we don't support the case where the developer calls both Facebook and Apple. Whichever one returns first will trigger an initialization, potentially ignoring the attribution from the second. This PR resolves this issue by better handling the case where two are in flight.

I've also added an additional boolean to be sent back to Branch where we'll know if the SDK checked for Apple search ad attribution.

@E-B-Smith


Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -370,9 +372,10 @@ - (void)initSessionWithLaunchOptions:(NSDictionary *)options isReferrable:(BOOL)
         if (![options.allKeys containsObject:UIApplicationLaunchOptionsURLKey] &&
             ![options.allKeys containsObject:UIApplicationLaunchOptionsUserActivityDictionaryKey]) {
             
+            self.asyncRequestCount = 0;
+            BOOL willNotCheckAsyncAttributionAPIs = ![self checkFacebookAppLinks] && ![self checkAppleSearchAdsAttribution];
```
E-B-Smith: Do you mean the short circuit `&&` here?  If a Facebook deferred applink is being checked, then it will _*not*_ check for Apple Search Ads.

Maybe this is clearer and doesn't short-circuit:
```objc
self.asyncRequestCount = 0;
BOOL checkingFacebook = [self checkFacebookAppLinks];
BOOL checkingAppleAds = [self checkAppleSearchAdsAttribution];

if (!(checkingFacebook || checkingAppleAds)) {
...
```

E-B-Smith:
After from my one comment is addressed, I can QA and get this in production very shortly.

Branch-SDK/Branch-SDK/Branch.m
aaustin: mm yea good catch

aaustin:
Corrected

E-B-Smith:
👍

Issue #612, opened 2017-04-05
branch.framework 0.14.10 build failed


adeb6600:
If there is a bug or error in this repo, please file the issue here. 
branch.framework 0.14.10 failed to build 
BranchShareLink.h not found
If you need assistance integrating a Branch SDK, please email support@branch.io. Thank you.

E-B-Smith:
How are you integrating Branch into your project?  Through direct integration, CocoaPods, or Carthage?

Caiopia:
I'm also getting this error with 0.14.10. It seems like `BranchShareLink.h ` is not in `Branch.framework/Headers/`. Adding it there seems to fix it for me.

E-B-Smith:
Fixed in release 0.14.11.

PR #613, opened 2017-04-06
Added BranchShareLink.h to public headers.

E-B-Smith:


E-B-Smith:
Fixes GH issue #612.

ahmednawar:
would this add it to the public headers in Carthage as well?

E-B-Smith:
The header is already added to the Carthage build.

ahmednawar:
👍

PR #625, opened 2017-05-16
Update BNCStrongMatchHelper to handle UISplitViewController

hongrich:
Similar to how UINavigationController and UITabBarController is
currently being handled when adding BNCMatchView. If we don't handle
UISplitViewController this way, BNCMatchViewController will end up in
the split view's viewControllers property which is unexpected and may
lead to crashes later. Split view's viewControllers should only have 1
or 2 view controllers depending on its collapsed / expanded state.

The special handling of UINavigationController and UITabBarController
was added in https://github.com/BranchMetrics/ios-branch-deep-linking/pull/539

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m
@@ -254,7 +254,7 @@
 }
 
 /**
-  Find the top view controller that is not of type UINavigationController or UITabBarController
+  Find the top view controller that is not of type UINavigationController, UITabBarController, UISplitViewController
  */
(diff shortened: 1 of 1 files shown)


BranchMacMini:
Result of [Integration 1](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/36c413b26e7abf951dbb3e5f3d474dd6/36c413b26e7abf951dbb3e5f3d475904)
---
*Duration*: 4 minutes and 20 seconds
*Result*: **Perfect build!** All 169 tests passed. :+1:
*Test Coverage*: 57%

E-B-Smith:
Great addition -- Thanks!

PR #630, opened 2017-05-23
[Don't Merge] Condition removed to check title and canonicalId - INTENG-3409

parthkalavadia:
Condition removed to check title and canonicalId while creating shortUrl in Branch.

### DIFF
--- a/Branch-SDK/Branch-SDK/BranchUniversalObject.m
+++ b/Branch-SDK/Branch-SDK/BranchUniversalObject.m
@@ -151,10 +151,6 @@
 #pragma mark - Link Creation Methods
 
 - (NSString *)getShortUrlWithLinkProperties:(BranchLinkProperties *)linkProperties {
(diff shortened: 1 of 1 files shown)


BranchMacMini:
Result of [Integration 1](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/b388335146db7e936b7215574da4f84e/b388335146db7e936b7215574da503dd)
---
*Duration*: 4 minutes and 50 seconds
*Test Coverage*: 57%

aaustin:
Messaged you asking for a bit more explanation for this change.

E-B-Smith:
So these fields are used to de-dup links -- Awesome.  Now we now.

Let's figure out the right we to do this and why Android isn't doing it.

E-B-Smith:
This will break existing compiles.  Better to let it silently fail?

PR #642, opened 2017-06-23
Make it easier to use the Branch test key - AIS-287

E-B-Smith:
Right now, switching between the `key_test` and the `key_live` confusing.

If the user can set a boolean or explicitly set the key before initialization then switching between live/test is a simple one line code change for the partner.
Options

### DIFF
--- a/Branch-SDK/Branch-SDK/Branch.h
+++ b/Branch-SDK/Branch-SDK/Branch.h
@@ -135,6 +135,14 @@ typedef NS_ENUM(NSUInteger, BranchCreditHistoryOrder) {
 ///--------------------------------
 
 /**
+ Gets the global, test Branch instance.
+
+ @warning This method is not meant to be used in production!
+*/
++ (Branch *) getTestInstance __attribute__((deprecated(("Use `Branch.useTestBranchKey = YES;` instead."))));
+
+
+/**
  Gets the global, live Branch instance.
  */
 + (Branch *)getInstance;
--- a/Branch-SDK/Branch-SDK/Branch.h
+++ b/Branch-SDK/Branch-SDK/Branch.h
@@ -140,13 +140,6 @@ typedef NS_ENUM(NSUInteger, BranchCreditHistoryOrder) {
 + (Branch *)getInstance;
 
 /**
- Gets the global, test Branch instance.
- 
- @warning This method is not meant to be used in production!
- */
-+ (Branch *)getTestInstance;
-
-/**
  Gets the global Branch instance, configures using the specified key
  
  @param branchKey The Branch key to be used by the Branch instance. This can be any live or test key.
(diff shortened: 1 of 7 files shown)


BranchMacMini:
Result of [Integration 1](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/f88f43c5166eeaf93fa88b73ca0ade2b/f88f43c5166eeaf93fa88b73ca0ae025)
---
*Duration*: 5 minutes and 39 seconds
*Test Coverage*: 58%

ahmednawar:
I also vote for not removing the `getTestInstance` because it will break for current partners and it has been in our docs for a long time.

BranchMacMini:
Result of [Integration 2](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/f88f43c5166eeaf93fa88b73ca0ade2b/f88f43c5166eeaf93fa88b73ca85bbb6)
---
*Duration*: 5 minutes and 29 seconds
*Result*: **Build failed 7 tests** out of 179

E-B-Smith:
I've added back `getTestInstance`.  It's now deprecated.

Once the Branch object is instantiated with a key, test or live, the key can't be changed.  So `getTestInstance` returns the same instance that `getInstance` will.  It's what ever instance was created first.  So the order of use is important and a little confusing.  That's why I want to deprecate it.

E-B-Smith:
As far as different configurations go, I usually use an xcconfig file.  The values is the xcconfig file can be used in the Info.plist, so I can switch between live and test pretty easily.

BranchMacMini:
Result of [Integration 3](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/f88f43c5166eeaf93fa88b73ca0ade2b/f88f43c5166eeaf93fa88b73caa8086d)
---
*Duration*: 4 minutes and 30 seconds
*Test Coverage*: 58%

Branch-SDK/Branch-SDK/Branch.h
```diff
@@ -154,6 +155,33 @@ typedef NS_ENUM(NSUInteger, BranchCreditHistoryOrder) {
  */
 + (Branch *)getInstance:(NSString *)branchKey;
 
+/// TODO: Document:
```
jdee: Did you consider using class properties here? I like using `@property (class, readonly) BOOL useTestKey` because of the way it's bridged to Swift. With a class property you have `Branch.useTestBranchKey` in Swift. With a class method you have `Branch.useTestBranchKey()`. This is a minor concern. I'm mainly curious if you have any reason not to use class properties, maybe because of Xcode 7 support or something.

E-B-Smith: Yeah, unfortunately class properties don't exist prior to Xcode 8.

I've defined these class methods in a 'class property' style, so maybe we can turn this on later.

PR #643, opened 2017-06-23
Added a WebViewExample-Test schema

jdee:
This illustrates how to select the Branch live or test key without a) making any temporary changes to the code or configuration that later have to be commented or reverted; or b) keeping separate copies of the Info.plist for different configurations. Use the WebViewExample schema for the Branch live environment and the WebViewExample-Test schema for the Branch test environment. This differs a bit from @E-B-Smith's recommendation to use xcconfig files, but amounts to the same thing, I believe. It follows the process outlined [here](https://github.com/BranchMetrics/react-native-branch-deep-linking/blob/master/docs/branch-environments.md). It may be that an alternate process would be simpler and preferable. However, this is something requested by users at times, and this can be a useful illustration.

### DIFF
--- a/Examples/WebViewExample/Podfile
+++ b/Examples/WebViewExample/Podfile
@@ -20,3 +20,9 @@ pod "TextAttributes"
 # ----- Application target
 
 target "WebViewExample"
+
+# Add custom configurations. This informs CocoaPods that Test-Debug should be treated like
+# a debug build, not a release build, when building pods. In particular this means the
+# Branch SDK will be built with the DEBUG flag when using this configuration.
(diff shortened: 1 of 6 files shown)


BranchMacMini:
Result of [Integration 1](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/f88f43c5166eeaf93fa88b73ca2ca957/f88f43c5166eeaf93fa88b73ca2cb83d)
---
*Duration*: 5 minutes and 18 seconds
*Test Coverage*: 58%

E-B-Smith:
Thanks for the example.  I added an example of using xcconfig to change values in the Info.plist.

I like the xcconfig route because you can use one single file to configure your project and sub-projects, like an app that also has a Watch app and Today Widget.

👍

PR #646, opened 2017-06-27
[Don't merge] Add test schema to webview example -- plist keys

E-B-Smith:
This shows how I use xcconfig files to set configuration options in plists based on different schemes.

I thought @jdee would find it interesting.

### DIFF
--- a/Examples/WebViewExample/Podfile
+++ b/Examples/WebViewExample/Podfile
@@ -20,3 +20,9 @@ pod "TextAttributes"
 # ----- Application target
 
 target "WebViewExample"
+
+# Add custom configurations. This informs CocoaPods that Test-Debug should be treated like
+# a debug build, not a release build, when building pods. In particular this means the
+# Branch SDK will be built with the DEBUG flag when using this configuration.
+# https://github.com/BranchMetrics/react-native-branch-deep-linking/blob/master/docs/branch-environments.md#step-7-update-podfile
+project "WebViewExample", "Test-Debug" => :debug
--- a/Examples/WebViewExample/README.md
+++ b/Examples/WebViewExample/README.md
@@ -1,4 +1,4 @@
-# webview_example
+# WebViewExample
 
 This app presents a list of the planets in a UITableView. When each
 row is tapped, a custom ArticleView is displayed using the a UINavigationController.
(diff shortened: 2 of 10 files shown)


BranchMacMini:
Result of [Integration 1](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/f88f43c5166eeaf93fa88b73cac983e4/f88f43c5166eeaf93fa88b73cac992e4)
---
*Duration*: 5 minutes and 22 seconds
*Test Coverage*: 58%

E-B-Smith:
The idea here is that all the configuration parameters are in the xcconfig file.  

The conditional Info.plist value, the app name, or nearly anything can be configured in one file.  This avoids #ifdefs or other conditional code.

jdee:
I like the idea of using xcconfig files this way and avoiding conditional code, multiple Info.plist files, multiple Branch keys in a given file.

I notice that this changes the way the project behaves though. In this branch, it's set up so that the Release configuration use the live key, and the Debug configuration uses the test key. This is an improvement over commenting things out, but it removes something that was already present in the project. The version in the staging branch lets you choose the Branch environment based on build scheme rather than configuration. In particular it lets you debug with the live key and do release builds with the test key (e.g., for distribution on Crashlytics or TestFlight, something partners have specifically expressed an interest in). This was the reason for the other two configurations. The Test-Debug and Test-Release configurations selected the test Branch key (using a compilation flag) rather than the live one, which was used by Debug and Release. The second scheme used these configurations rather than the ones for the live key. To retain this flexibility, I think it makes sense to restore the other configurations but set them up so that the base configuration for each is:

- Debug: LiveBranch.xcconfig
- Release: LiveBranch.xcconfig
- Test-Debug: TestBranch.xcconfig
- Test-Release: TestBranch.xcconfig

And restore the WebViewExample-Test scheme to use Test-Debug and Test-Release.

I also notice that the path relative to the group is wrong for TestBranch.xcconfig, which is in the repo but shows up missing from the project.

I haven't worked much with xcconfig files before. I'd be happy to make these changes to get a chance to play with this.

jdee:
I misspoke slightly. It appears that the schemes select the Branch environment as before, but now WebViewExample uses the Release configuration for all tasks, and WebViewExample-Test uses the Debug configuration. This probably isn't the best approach. I know this was offered as an illustration. I can submit a PR using xcconfig files to support the same behavior as is currently in the staging branch.

I like this approach, and I think it's something I'm likely to use in my own apps in the future, but now that I've taken a stab at it, I think it's more complicated than using conditional compilation (partly because of CocoaPods; it's not so complicated if you don't have to worry about including the CocoaPods xcconfig). Of course, this is sort of an advanced case. It may just be a question of documentation and perhaps automation.

E-B-Smith:
You can set up as many build configs as you like, I was just lazy so I showed only two for the demo code.

I did this as more of an academic exercise because I thought you might find it helpful at some point and it's not an common way to set up projects, so maybe you haven't seen it before.

I'm not planning on merging this.

E-B-Smith:
This is simply an example.

PR #654, opened 2017-07-21
BNCDeviceInfo thread safety

jdee:
This is a bit of a shot in the dark to fix a very rare crash that may involve thread safety.

In stack traces, the crashes never occur until line 301. The deviceInfo.hardwareId and deviceInfo.hardwareIdType are both strings that appear to be safely used above line 301. However, unlike the later properties, these are copied. This doesn't seem a very likely cause, and since the crash has sometimes occurred below line 301, it may just be random to some degree.

With this in mind, I changed every reference property in BNCDeviceInfo to have `copy` semantics rather than `strong` reference semantics. This means that every access to these properties now involves a copy operation. This can improve thread safety, since it means that no two threads can ever handle the same instance of these properties at the same time.

Also:
- Changed them all to be atomic rather than nonatomic, for better thread safety.
- Eliminated all calls to self in init. This is usually OK, but at least considered a slightly unsafe practice.
- Made all these properties readonly as well, since they are never externally assigned.
- Added a lazy initialization to BNCDeviceInfo.vendorId since UIDevice.identifierForVendor (and BNCSystemObserver.getVendorId) can return nil once and non-nil later, according to Apple's docs (see code comment)

It's not obvious that this change will fix the crash, but BNCDeviceInfo is now entirely thread safe.

The only place the instance and its properties are ever called is in the BNCServerInterface when preparing a POST. This adds a small amount of overhead for locking and copying that should be negligible compared to an HTTPS request.

This relates to INTENG-3524.

**Updated:**

Since the crash seems to involve a bad message recipient, calls to self were separated from calls to deviceInfo, to verify which recipient is the problem in future crash reports.

Also, all the calls to `safeSetValue:forKey:onDict:` throughout the SDK are probably unnecessary, as indicated in a comment.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCServerInterface.m
+++ b/Branch-SDK/Branch-SDK/BNCServerInterface.m
@@ -297,20 +297,45 @@ NSString *requestEndpoint;
         dict[BRANCH_REQUEST_KEY_HARDWARE_ID_TYPE] = hardwareIdType;
         dict[BRANCH_REQUEST_KEY_IS_HARDWARE_ID_REAL] = isRealHardwareId;
     }
-    
-    [self safeSetValue:deviceInfo.vendorId forKey:BRANCH_REQUEST_KEY_IOS_VENDOR_ID onDict:dict];
-    [self safeSetValue:deviceInfo.brandName forKey:BRANCH_REQUEST_KEY_BRAND onDict:dict];
-    [self safeSetValue:deviceInfo.modelName forKey:BRANCH_REQUEST_KEY_MODEL onDict:dict];
-    [self safeSetValue:deviceInfo.osName forKey:BRANCH_REQUEST_KEY_OS onDict:dict];
-    [self safeSetValue:deviceInfo.osVersion forKey:BRANCH_REQUEST_KEY_OS_VERSION onDict:dict];
-    [self safeSetValue:deviceInfo.screenWidth forKey:BRANCH_REQUEST_KEY_SCREEN_WIDTH onDict:dict];
-    [self safeSetValue:deviceInfo.screenHeight forKey:BRANCH_REQUEST_KEY_SCREEN_HEIGHT onDict:dict];
 
-    [self safeSetValue:deviceInfo.browserUserAgent forKey:@"user_agent" onDict:dict];
-    [self safeSetValue:deviceInfo.country forKey:@"country" onDict:dict];
-    [self safeSetValue:deviceInfo.language forKey:@"language" onDict:dict];
+    NSString *vendorId = deviceInfo.vendorId;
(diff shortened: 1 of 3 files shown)


BranchMacMini:
Result of [Integration 1](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/40294f4db6a6c8ac6f2d52fa56da18f9/40294f4db6a6c8ac6f2d52fa56da2411)
---
*Duration*: 7 minutes and 55 seconds
*Test Coverage*: 57%

BranchMacMini:
Result of [Integration 2](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/40294f4db6a6c8ac6f2d52fa56da18f9/40294f4db6a6c8ac6f2d52fa56fbd6f7)
---
*Duration*: 6 minutes and 34 seconds
*Test Coverage*: 57%

aaustin:
This looks good to me 👍

Branch-SDK/Branch-SDK/BNCServerInterface.m
```diff
@@ -284,32 +284,58 @@ - (void) collectInstrumentationMetrics {
...
-    [self safeSetValue:deviceInfo.screenWidth forKey:BRANCH_REQUEST_KEY_SCREEN_WIDTH onDict:dict];
-    [self safeSetValue:deviceInfo.screenHeight forKey:BRANCH_REQUEST_KEY_SCREEN_HEIGHT onDict:dict];
-
-    [self safeSetValue:deviceInfo.browserUserAgent forKey:@"user_agent" onDict:dict];
-    [self safeSetValue:deviceInfo.country forKey:@"country" onDict:dict];
-    [self safeSetValue:deviceInfo.language forKey:@"language" onDict:dict];
-
-    dict[BRANCH_REQUEST_KEY_AD_TRACKING_ENABLED] = @(deviceInfo.isAdTrackingEnabled);
+
+    NSString *vendorId = deviceInfo.vendorId;
```
E-B-Smith: Why is this better than setting the device info variables directly? Is this for debugging?

jdee: Yes. (And it isn't better. I'd be inclined to revert it later after the problem is identified.) The stack trace consistently shows EXC_BAD_ACCESS from objc_msgSend, which this SO post says is due to a bad message recipient: https://stackoverflow.com/a/7684896/875725. It stands to reason, since obc_msgSend should not care about the args. When you have `[self safeSetValue:deviceInfo.vendorId etc:...]`, there are two messages sent in the same line. If the crash is consistent after this change, we can be certain it's the `BNCDeviceInfo` that's at fault rather than the `BNCServerInterface` (or vice versa).

PR #660, opened 2017-07-26
Adjust podspec for networking changes

jdee:
Builds with CocoaPods (e.g. WebViewExample) were failing because of the recent networking changes. This change fixes that. It also cleans up the podspec a little to avoid some repetition and modernize a Ruby Hash literal.

### DIFF
--- a/Branch.podspec
+++ b/Branch.podspec
@@ -24,6 +24,7 @@ Use the Branch SDK (branch.io) to create and power the links that point back to 
     "Branch-SDK/Branch-SDK/Networking/*.{h,m}",
     "Branch-SDK/Branch-SDK/Networking/Requests/*.{h,m}",
     "Branch-SDK/Fabric/*.h"
+  s.default_subspec = 'Core'
 
   s.subspec 'Core' do |core|
     core.source_files = source_files
--- a/Branch.podspec
+++ b/Branch.podspec
@@ -15,26 +15,31 @@ Use the Branch SDK (branch.io) to create and power the links that point back to 
   s.homepage         = "https://branch.io"
   s.license          = 'Proprietary'
   s.author           = { "Branch" => "support@branch.io" }
-  s.source           = { :git => "https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK.git", :tag => s.version.to_s }
+  s.source           = { git: "https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK.git", tag: s.version.to_s }
   s.platform         = :ios, '7.0'
   s.requires_arc     = true
 
+  source_files =
+    "Branch-SDK/Branch-SDK/*.{h,m}",
+    "Branch-SDK/Branch-SDK/Networking/*.{h,m}",
+    "Branch-SDK/Branch-SDK/Networking/Requests/*.{h,m}",
+    "Branch-SDK/Fabric/*.h"
+
   s.subspec 'Core' do |core|
-    core.source_files = "Branch-SDK/Branch-SDK/*.{h,m}", "Branch-SDK/Branch-SDK/Requests/*.{h,m}", "Branch-SDK/Fabric/*.h"
+    core.source_files = source_files
     core.private_header_files = "Branch-SDK/Fabric/*.h"
     core.frameworks = 'AdSupport', 'MobileCoreServices'
   end
 
   s.subspec 'without-IDFA' do |idfa|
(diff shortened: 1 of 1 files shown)


BranchMacMini:
Result of [Integration 1](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c8e561f6311c1c559ca1339dc8079030/c8e561f6311c1c559ca1339dc8079ea4)
---
*Duration*: 7 minutes and 59 seconds
*Test Coverage*: 60%

Branch.podspec
```diff
@@ -15,26 +15,31 @@ Use the Branch SDK (branch.io) to create and power the links that point back to
   s.homepage         = "https://branch.io"
   s.license          = 'Proprietary'
   s.author           = { "Branch" => "support@branch.io" }
-  s.source           = { :git => "https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK.git", :tag => s.version.to_s }
+  s.source           = { git: "https://github.com/BranchMetrics/iOS-Deferred-Deep-Linking-SDK.git", tag: s.version.to_s }
   s.platform         = :ios, '7.0'
   s.requires_arc     = true
 
+  source_files =
```
jdee: I'm pretty sure this can just be `s.source_files = ...`. https://guides.cocoapods.org/syntax/podspec.html#subspec That attribute should be inherited by all subspecs unless overridden.

Branch.podspec
```diff
@@ -15,26 +15,31 @@ Use the Branch SDK (branch.io) to create and power the links that point back to
...
   end
 
   s.subspec 'without-IDFA' do |idfa|
-    idfa.source_files = "Branch-SDK/Branch-SDK/*.{h,m}", "Branch-SDK/Branch-SDK/Requests/*.{h,m}", "Branch-SDK/Fabric/*.h"
+    idfa.source_files = source_files
     idfa.private_header_files = "Branch-SDK/Fabric/*.h"
     idfa.frameworks = 'MobileCoreServices'
   end
 
   s.subspec 'without-Safari' do |safari|
```
jdee: I'm also not entirely sure there's any difference between the Core and with-Safari subspecs.

Branch.podspec
jdee: Nevermind. Each subspec requires at minimum source_files or a number of other fields not present here. These have to stay, so using the local variable is probably the easiest thing to do. The private_header_files can probably belong to the root spec, however.

BranchMacMini:
Result of [Integration 2](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c8e561f6311c1c559ca1339dc8079030/c8e561f6311c1c559ca1339dc86fee1d)
---
*Duration*: 7 minutes and 30 seconds
*Test Coverage*: 60%

jdee:
There is a bit of a pathology in the podspec as it stands. By default, CocoaPods installs all subspecs of a pod. So as it stands, you get this with `pod "Branch"`:

```
PODS:
  - Branch (0.17.0):
    - Branch/Core (= 0.17.0)
    - Branch/without-IDFA (= 0.17.0)
    - Branch/without-Safari (= 0.17.0)
  - Branch/Core (0.17.0)
  - Branch/without-IDFA (0.17.0)
  - Branch/without-Safari (0.17.0)
```

It builds all three subspecs into the same Branch.framework. I suspect this just happens to work because it merges the dependencies from all subspecs. I doubt that it actually increases the size of the library when using the pod, but I'm not sure.

I added a commit that sets `s.default_subspec = "Core"` to avoid this, so when you specify `pod "Branch"` you only get the default:

```
PODS:
  - Branch (0.17.0):
    - Branch/Core (= 0.17.0)
  - Branch/Core (0.17.0)
```

I'm not sure if subspecs are the best way to define build variants like this with CocoaPods. I didn't quickly find examples of open-source pods that do this sort of thing. I think the usual approach though with AdSupport is not to make it a dependency of the pod. Users are just instructed to add it to their projects if they want it. See the podspecs for Fabric and Crashlytics, e.g. No mention anywhere of AdSupport:

https://github.com/CocoaPods/Specs/blob/master/Specs/0/a/c/Fabric/1.6.12/Fabric.podspec.json
https://github.com/CocoaPods/Specs/blob/master/Specs/3/6/0/Crashlytics/3.8.5/Crashlytics.podspec.json

It would probably be much simpler to remove the subspecs altogether and have a single spec without these frameworks. If they are in the app, the SDK will use them. If not, nothing happens. I also think this probably matches what users have to do with manual integration and Carthage.

PR #666, opened 2017-07-29
INTENG-3579 Default dynamic links fix

agrimn:
By default, Dynamic links are created every time internet connectivity is down.
- Set forceLinkCreation to YES by default.
- Fixed indentation.

### DIFF
--- a/Branch-SDK/Branch-SDK/Branch.m
+++ b/Branch-SDK/Branch-SDK/Branch.m
@@ -1087,23 +1087,23 @@ static BOOL bnc_enableFingerprintIDInCrashlyticsReports = YES;
 #pragma mark - ShortUrl methods
 
 - (NSString *)getShortURL {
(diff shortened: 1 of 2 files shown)


BranchMacMini:
Result of [Integration 1](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/0bf1fc775fcb4dc220f39b63f14f253b/0bf1fc775fcb4dc220f39b63f14f300c)
---
*Duration*: 6 minutes and 56 seconds
*Test Coverage*: 60%

ahmednawar:
I am still confused. Why does forcing link creation generate a dynamic link by default? Don't we already generate dynamic (long) links in the case of network failures?

E-B-Smith:
@ahmednawar @agrimn Agrim looked into this:  Apparently Android generated long links when there was a network failure, but iOS didn't.  Agrim made this change so that both SDKs worked the same.

ahmednawar:
iOS does the same right now as well. If you enable Airplane mode, then try to generate a link on the testbed, you will get a long link.

PR #668, opened 2017-08-01
Support deduping our NSUserActivity-based indexed Spotlight items

aaustin:
We've recently had requests that would allow partners to index their content with our BUO methods but continue to update the the metadata over time. The current production version of this implementation makes no effort to de-dupe across each BUO item because we previously found it impossible to do. I've experimented a bit this evening and found that `weakRelatedUniqueIdentifier` actually solves this without breaking our Spotlight implementation. Unfortunately, it appears that Apple only lets you update an item one single time. I tested many different mechanisms to get around this limitation, like setting the modification date and resigningCurrent before calling becomeCurrent. Nothing seems to work. Still, this is better than nothing.

Also, I updated the testbed to share the same keys as the Android one.

@jdee


BranchMacMini:
Result of [Integration 1](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/77ca53d7b0df82cb52d0785ab10e18a9/77ca53d7b0df82cb52d0785ab10e244a)
---
*Duration*: 7 minutes and 35 seconds
*Test Coverage*: 60%

E-B-Smith:
Apple only lets you change the information one time?  That seems like a very odd design choice on their part. 

👍

E-B-Smith:
I'll pass along the change of keys to Parul -- She'll need to update her testing run book.

Thanks for this. We are going to dig into other Spotlight issue as well this week.

PR #676, opened 2017-08-04
Big cleanup of memory and threading errors.

E-B-Smith:
* SDK now passes the address sanitizer and thread sanitizer tests.
* Added more tests to catch these errors early.


BranchMacMini:
Result of [Integration 1](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/1ba939e28abf51c124246b3ad333d16c/1ba939e28abf51c124246b3ad333ddd3)
---
*Duration*: 9 minutes and 13 seconds
*Result*: **Perfect build!** All 199 tests passed. :+1:
*Test Coverage*: 61%

Branch-SDK/Branch-SDK/NSMutableDictionary+Branch.h
```diff
@@ -10,7 +10,7 @@
 #import <Foundation/Foundation.h>
 
 
-void ForceNSMutableDictionaryToLoad(void);
+void ForceNSMutableDictionaryToLoad(void) __attribute__((constructor));
 
 
 @interface NSMutableDictionary (Branch)
```
jdee: I think it might be possible to remove this whole category and just use subscript assignment, e.g. `dict[@"key"] = nil;` is just fine, I believe. This would also include the `safeSetValue` methods here and there in the SDK. I don't think it does any harm, but less is more.

scripts/test-unit-tests
```diff
@@ -64,9 +64,22 @@ function analyzeBuild() {
         -workspace Branch-TestBed.xcworkspace \
         -scheme Branch-TestBed \
         -destination name="iPhone 7",OS=latest \
-            clean analyze -quiet
+            analyze -quiet
```
jdee: One of these days I'll give you one Fastfile that replaces this whole folder. :)

PR #688, opened 2017-08-15
[WIP: Comments Welcome] Added NSNotifications and a Branch delegate (AIS-271).

E-B-Smith:
Branch can be awkward to integrate in certain kinds of apps since the Branch object only communicates state through a callback block.  I added a Branch delegate and NSNotifications.

1) Added a Branch delegate with optional callback methods:
   `branch:willOpenURL:`
   `branch:didOpenURL:linkParameters:error:`

2) Added Branch NSNotifications:
    `BNCBranchWillOpenURLNotification` and `BNCBranchDidOpenURLNotification`

3) Added an example in the Branch-Testbed that demonstrates usage.

To Do:
1) Get feedback from stake holders.
2) Add documentation.


BranchMacMini:
Result of [Integration 1](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/c57398147d6bbaee710ac6bf1097645b)
---
*Duration*: 9 minutes and 8 seconds
*Result*: **Perfect build!** All 200 tests passed. :+1:
*Test Coverage*: 60%

BranchMacMini:
Result of [Integration 2](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/c57398147d6bbaee710ac6bf10bcad85)
---
*Duration*: 7 minutes and 49 seconds
*Result*: **Perfect build!** All 200 tests passed. :+1:
*Test Coverage*: 60%

Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
```diff
@@ -345,6 +346,20 @@ - (void)setExternalIntentURI:(NSString *)externalIntentURI {
     }
 }
 
+- (NSString*) referredUrl {
+    @synchronized (self) {
+        if (!_referredUrl) _referredUrl = [self readStringFromDefaults:@"referredUrl"];
```
aaustin: The rest of this file uses constants with a specific pattern for the keys to ns user defaults - why are you deviating here?

aaustin:
I think my main concern with adding this is that I'm not sure where in the documentation you'd put it? https://dev.branch.io. If it's not in the docs, no one will use it and therefore it's superfluous.

Related to the above, have we received feedback that people need this? If so, what problem are they trying to solve that can't be solved with the delegate callback block? I don't want to add ideological methods without a customer need or use case.

Branch-SDK/Branch-SDK/BranchNotifications.h
```diff
@@ -0,0 +1,31 @@
...
+//  Branch-TestBed
+//
+//  Created by edward on 6/30/17.
+//  Copyright © 2017 Branch Metrics. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+@class Branch;
+
+@protocol BranchDelegate <NSObject>
```
jdee: This is a nice addition to the SDK. I would at least add a delegate method like `branch:didOpenBranchUniversalObject:withLinkProperties:` and include a BUO and BranchLinkProperties in the notification payload. Using untyped data (params) without a data model is just bad app design in a typed language and should be discouraged, though I don't see any harm in exposing the raw params for those who want access to them for diagnostics or something. I would also argue for a lot more semantic distinction. This change adds new notification mechanisms as alternatives to a single callback block in the app delegate, which is a big improvement. But it's still confusing for lots of people (myself included) that the same callback used to notify users that a link was opened (which is generally all they care about) is called when no link is available. It's not a bad idea to give feedback that the SDK was initialized successfully, and no link is available. But calling the app when the SDK does a deferred deep link check on foreground and then notifying the app that nothing happened is not the best idea. And in any case, I think much more semantic distinction is required. I'd like to see distinct notifications and delegate methods for: initialization success; initialization failure; link open success; link open failure. The open events should only occur when a user actually opened a link without an error, and errors should be distinct events, not parameters in the success callback. A method like `didEncounterError:` is extremely common in Obj-C delegates. But maybe additional delegate methods and notifications would be a separate change.

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1919,6 +1951,35 @@ - (void)handleInitSuccess {
     }
 }
 
+- (void) sendOpenNotificationWithOriginalURL:(NSURL*)originalURL
+                              linkParameters:(NSDictionary*)linkParameters
+                                       error:(NSError*)error {
+
+    NSMethodSignature *signature = nil;
```
jdee: I'm curious why the NSInvocation here. This would be a lot simpler with
```Obj-C
if ([self.delegate respondsToSelector:selector]) {
    [self.delegate branch:self didOpenURL:originalURL linkParameters:linkParameters: error:error];
}
```

I'm sure there's a good reason.

E-B-Smith: The reason is that I was showing Parth how to use invocations.

I will change it to the obvious and intuitive method.

Branch-SDK/Branch-SDK/BranchNotifications.h
E-B-Smith: 1.  I agree that untyped data is not a good idea.  I think changing these notifications to a BUO is a great idea.  Have untyped data is hard for the SDK user, and we are trying to move everything to a BUO anyway.

2. Thanks for the feedback on semantic naming of the notifications / delegate methods.

Maybe these names are better:

`branch:willOpenURL:`  (When URL is non-null).
`branchWillCheckForDeferredDeepLink:` (When URL is null).

`branch:didOpenBranchUniversalObject:withLinkProperties:`
`branch:failedToOpenURL:withError:`

what other events and notifications would be helpful?  How should they be named?

E-B-Smith:
As far as documentation goes, I plan on making a small section explaining the delegate methods and notifications and making it clear that they are _optional_ alternatives to the handling everything in the callback.

Branch-SDK/Branch-SDK/BranchNotifications.h
jdee: I think the changes you suggested sound good. I'm sure this will evolve if there are more things that are really needed. :+1:

aaustin:
Can you modify the readme so that I can review how you plan on recommending the implementation alongside the current integration path?

E-B-Smith:
I'll add the docs shortly. Let me finish up some other stuff first.

ethanneff:
@E-B-Smith ping me once the readme is complete so I can make sure the changes go in the new docs

BranchMacMini:
Result of [Integration 3](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/1f920e0ff1c6754b9c4407240debe081)
---
*Duration*: 8 minutes and 38 seconds
*Result*: **Build failed 1 test** out of 200

BranchMacMini:
Result of [Integration 4](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/3a74b4a7fc6611bd0ef3cc8456142a0b)
---
*Duration*: 9 minutes and 48 seconds
*Result*: **Build failed 1 test** out of 200

BranchMacMini:
Result of [Integration 5](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/3a74b4a7fc6611bd0ef3cc84563b3c8a)
---
*Duration*: 8 minutes and 33 seconds
*Result*: **Build failed 1 test** out of 200

BranchMacMini:
Result of [Integration 6](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/c6b42f1b298fcd3b1544429c2429e34e)
---
*Duration*: 11 minutes and 45 seconds
*Result*: **Build failed 2 tests** out of 200

BranchMacMini:
Result of [Integration 6](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/c6b42f1b298fcd3b1544429c2429e34e)
---
*Duration*: 11 minutes and 45 seconds
*Result*: **Build failed 2 tests** out of 200

BranchMacMini:
Result of [Integration 8](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/e2044af8e1ee7c22a9ec25423529afaa)
---
*Duration*: 9 minutes and 37 seconds
*Result*: **Build failed 1 test** out of 204

BranchMacMini:
Result of [Integration 9](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/e2044af8e1ee7c22a9ec2542354f2595)
---
*Duration*: 10 minutes and 12 seconds
*Result*: **Build failed 2 tests** out of 204

BranchMacMini:
Result of [Integration 10](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/c43e569ae63caa771d39544766d865bf)
---
*Duration*: 10 minutes and 1 second
*Result*: **Build failed 1 test** out of 204

BranchMacMini:
Result of [Integration 11](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/c43e569ae63caa771d39544766fe63a7)
---
*Duration*: 9 minutes and 25 seconds
*Result*: All 204 tests passed, but please **fix 1 warning**.
*Test Coverage*: 62%

BranchMacMini:
Result of [Integration 12](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/d5a4b2fd72dec1b24b96bc996cc17a95)
---
*Duration*: 2 minutes and 10 seconds
*Result*: **4 errors, failing state: build-errors**

BranchMacMini:
Result of [Integration 13](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/d5a4b2fd72dec1b24b96bc996cc2b4a0)
---
*Duration*: 13 minutes and 58 seconds
*Result*: **-1 error, failing state: internal-build-error**

BranchMacMini:
Result of [Integration 14](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/d5a4b2fd72dec1b24b96bc996cc7b99d)
---
*Duration*: 14 minutes and 31 seconds
*Result*: **-1 error, failing state: internal-build-error**

BranchMacMini:
Result of [Integration 15](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/d5a4b2fd72dec1b24b96bc996cc85302)
---
*Duration*: 9 minutes and 59 seconds
*Result*: All 204 tests passed, but please **fix 1 warning**.
*Test Coverage*: 62%

BranchMacMini:
Result of [Integration 16](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/a63029b6e050121ba0af3689eb84921e)
---
*Duration*: 9 minutes and 35 seconds
*Result*: All 204 tests passed, but please **fix 1 warning**.
*Test Coverage*: 62%

E-B-Smith:
Hey @jdee , Would you mind taking a look at this again and let me know what you think?

I've added your suggestions for this PR to make the delegate and notifications more useful, and I've add a small example app called BareBones.

The relevant files to look at are the delegate / notification header file and the BareBones documentation, part of which I plan on rolling into our bigger doc at some point.

Header file:
https://github.com/BranchMetrics/ios-branch-deep-linking/blob/d04e66eb8ab0d651f834e77bcc6ab3b949cdf257/Branch-SDK/Branch-SDK/BranchDelegate.h

Document:
https://github.com/BranchMetrics/ios-branch-deep-linking/blob/d04e66eb8ab0d651f834e77bcc6ab3b949cdf257/Examples/BareBones/README.md

jdee:
This is a nice implementation and a nice example. I do see something wrong with BareBones, however. If I open a link, it routes to a view (LinkVC, I think) after showing the WaitingVC. If I then background the app and put it in the foreground again, I get the WaitingVC again, though no link was opened.

I think this gets to the heart of my concern about the API for the Branch SDK in the first place. There are a number of issues that originally confused me when I was first working with the Branch SDK (not to mention QA at the last job) and seem to confuse partners on a regular basis, because the behavior of the SDK is often unexpected. This is just my opinion, but I suspect a lot of partners might agree. The following remarks are motivated by practical concerns about supportability, though I think these things might also fall under the heading of best practices. And since you asked, here are my two cents:

I've often seen people remark in GitHub issues that something isn't working, "and the callback fires with `{"+clicked_branch_link": false, "+is_first_session": false}` when I foreground the app. Obviously something is broken." In fact, they have some other problem, but the fact that this behavior is unexpected means it always has to be explained and always arouses suspicion. If the SDK behaved in a way that was obvious and didn't have to be explained, it would be a big improvement. The current behavior is a little like a software implementation of Homer Simpson's Everything's OK Alarm.

Everything's OK

With recent changes to WebViewExample, I witnessed some of this behavior in spades, and it confused me greatly in recent testing until I realized what was happening. Because `handleDeepLinkWithNewSession:` passes every link to the server, even if the method returns NO, and it passes every response back to the app, the app receives quite a lot of `+non_branch_link` responses when using the About the App view. I think that could stand to be improved.

I can't find it now, but I read somewhere (possibly the Apple Human Interface Guidelines) a very useful piece of advice: Never notify the user of an error that occurred in response to an action they did not initiate. Consumers of this SDK are developers, not end users, but a similar idea could be put forth here: Don't notify the app when 1) an error occurred; or 2) nothing happened in response to an event that was entirely initiated by the SDK, not the user or the app. If the user tries to open a link, and that fails, of course an error should be reported. Arguably, if the SDK would certainly have returned a link in response to a deferred deep link check, but cannot because of an error, that should be reported. One problem is that the SDK doesn't know this before calling the server. So the willStart notification as it stands can really only mean will check for a deferred deep link but may not actually be opening a link. Perhaps the willStart notification could be moved somewhere like `handleDeepLink` or `continueUserActivity`, so that the SDK can be fairly sure that it is actually responding to an open from the app, not a deferred deep link check, and events can be generated only when those methods return YES, so there's reason to expect the link open to succeed. It could also happen at initialization if the appropriate launch option key is available, so the app knows it was launched from a link.

Further, the data model here is not exactly request-response. A request gets only a single response. (Yes, there are some odd models here and there that don't conform, but this is generally the case.) The model here is a subscription to a stream of asynchronous events. In fact, with the addition of these notifications, the subscription is established through the NotificationCenter, and subscription now clearly (and correctly) is unrelated to initialization. Calling `initSession` may result in a series of requests with individual responses, but that's a matter of implementation. It's most natural to expect the app to be called back only when events of interest occur. Every request to the server will generate a response. Every response need not (and should not) be passed back to the app, only those that indicate events of interest to the app. And the notifications should be used to distinguish different types of event.

I think this is the right idea:

- BranchSessionDidStartNotification: Indicates initialization success. Generated only once, because the app initiated the session.
- BranchSessionDidFailToStartNotification: Indicates initialization failure. Generated only once, because the app initiated the session.
- BranchLinkWasOpenedNotification: Indicates that a link was successfully opened. Generated every time a link is actually opened and available. Includes the link and all relevant data from the server. _This is generated in addition to BranchSessionDidStartNotification if an app is launched from a link._ Links are irrelevant to initializing a session except as an implementation detail. There should be no need to handle it in more than one notification.
- BranchLinkDidFailToOpenNotification: Indicates that an open request initiated by the app failed.

Arguably, you can live with the same notification used for success and error, with an optional error element that indicates failure. I think smaller interfaces are better, and so is finer semantic distinction. There is precedent in lots of delegate protocols, e.g. WKNavigationDelegate.

I don't think there's any need to hold up merging this PR, though the behavior of BareBones might be improved, at least in app code. I think this could easily live in staging for a while, and you could refine it over time.

aaustin:
Thanks for this write up Jimmy. I think the analogy is sound and the original pattern does have its flaws. My original thought when implementing the consistent callback was that "developers will call initialize session and register a deep link handler callback expecting a response, so we better respond 100% of the time a session initializes." That's why the methods are named as such. There were some unwanted behaviors that we tied the definition of session to foreground/background, like session reinitialization in system alerts (push notifs, location etc), but it delivered a consistent response when a session was initialization otherwise.

However, your point here is very relevant. Developers are registering for notifications and should not be alerted when there's no data to process. It looks like we have this functionality under `BranchLinkWasOpenedNotification`, so it's possible for the developer to register for just the relevant stuff. The issue you're taking is mostly with the testbed, no?

E-B-Smith:
Thanks for taking the time with all the thoughtful feedback.

I don't want to leave your comments hanging but I don't have much time today to respond. In short, maybe we should better communicate that a Branch session isn't just the first time the app opens, it can happen multiple times during a normal app life time.  Foreground / Notifications and other events can start a new Branch session.  That said, often the first app run is special. Do you think it would be convenient to have a special indicator for first Branch initialization?

Also I want to mention that the issue with the BareBones example showing the last link opened is a bug that I've fixed. I'm still testing the fix before I push it.

jdee:
Thanks @aaustin and @E-B-Smith. Nothing's ever perfect, and hindsight is always 20/20. I'm not always sure if my point will get across, so I may have both overstated and diluted what I wanted to say.

Here it is in bullet points:

- The current API for the SDK is low-level. Works great, and lots of apps build a great deep-linking experience on it. But there's currently only one entry point, often a block literal in the AppDelegate, that just tells the app that something somewhere happened, maybe a link open, maybe an error, maybe nothing. Please examine the data to determine what happened and what you need to do.
- What an app really wants is to be told that an event of interest occurred and receive just the relevant data.
- Changing the existing behavior might be good, but changing things is risky, and that probably at least requires some review.
- This is a really nice improvement to the SDK, and it offers an excellent opportunity to address these semantic issues.
- I like what's been done, but I would argue for more notifications and more delegate methods with a finer grain.

I'll just add an example to show why I think this is still a little complicated. This is from BareBones:

```swift
    @objc func branchDidStartSession(notification: Notification) {
        WaitingViewController.hide()

        let url : URL? = notification.userInfo?[BranchURLKey] as? URL

        if let error = notification.userInfo?[BranchErrorKey] as? Error {
            if let url = url {
                self.showAlert(
                    title: "Couldn't Open URL",
                    message: "\(url.absoluteString)\n\n\(error.localizedDescription)"
                )
            } else {
                self.showAlert(title: "Error Starting Branch Session", message: error.localizedDescription)
            }
            return
        }
        if let buo = notification.userInfo?[BranchUniversalObjectKey] as? BranchUniversalObject {
            let linkViewController = LinkViewController.instantiate()
            linkViewController.branchObject = buo
            linkViewController.branchURL = url
            navigationController?.pushViewController(linkViewController, animated: true)
            AppStats.shared.linksOpened += 1
            return
        }
    }
```

That's a large method with multiple purposes. If the delegate methods were more finely-grained:

```swift
@objc func branchDidStartSession(branch: Branch) {
    NSLog("Branch started successfully") // nothing to do
}

@objc func branch(branch: Branch, didEncounterInitializationError error: Error) {
    showAlert(title: "Error Starting Branch Session", message: error.localizedDescription)
}

@objc func branch(branch: Branch, didEncounterLinkOpenError error: Error, withURL url:URL) {
    WaitingViewController.hide()
    showAlert(title: "Couldn't Open URL", message: "\(url.absoluteString)\n\n\(error.localizedDescription)")
}

@objc func branch(branch: Branch, didOpenLink link: BranchLink, withURL url: URL) {
    WaitingViewController.hide()

    let linkViewController = LinkViewController.instantiate()
    linkViewController.branchObject = link.universalObject
    linkViewController.branchURL = url
    navigationController!.pushViewController(linkViewController, animated: true)
    AppStats.shared.linksOpened += 1
}
```

Note: No optionals. No guards. No pyramid code. Trivial to read and maintain because
no contextual interpretation is necessary. When an open event happens, the URL and
BranchLink are not optional. How can you open a link without a link? The original uses
notifications, which are generic objects, so you'll always at least have something like:

```swift
let url = notification.userInfo![BranchURLKey] as! URL
```

This is always a trade-off between notifications and delegates, so I chose the delegate
in order to illustrate the point more clearly, but you can simplify notification handling
along these lines as well using multiple notification types. And multiple notifications can let you fan out these tiny tasks to multiple code units. With a single notification, these events always
force you through a single selector, and you have to do the fanout yourself.
When a single notification or delegate method is reused with lots of parameters, you always have to have if-else blocks, guards and so on.

Again, I clicked "approve" on the original PR, and except for the bug fix Ed's working on, I'd say this can be merged. But I'd also give some thought to further improvements along these lines, since this is new.

aaustin:
I think I need to review more in depth. I don’t think it’s a good idea to implement a notification that alerts every session open. It should only alert when link data is present. Can’t the developer register for the BranchLinkWasOpened notification and not need all that ghetto logic to check it link data is present? Listening to every session is optional and shouldn’t be recommended. Otherwise, it’s the exact same thing as my original callback block implementation.

BranchMacMini:
Result of [Integration 17](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/892fe5e52c6501f514715b92607a308e)
---
*Duration*: 9 minutes and 53 seconds
*Result*: **1 error, failing state: build-errors**

jdee:
I have a feeling we could arrive at a clear understanding if we all jumped on a 10-min. video call or something.

On a tangential note, with regard to notifications, Apple's [MPMusicPlayerController](https://developer.apple.com/documentation/mediaplayer/mpmusicplayercontroller) has two methods: [beginGeneratingPlaybackNotifications](https://developer.apple.com/documentation/mediaplayer/mpmusicplayercontroller/1624199-begingeneratingplaybacknotificat) and [endGeneratingPlaybackNotifications](https://developer.apple.com/documentation/mediaplayer/mpmusicplayercontroller/1624247-endgeneratingplaybacknotificatio). These are useful when working with that utility because the alternative is to constantly remove and add observers for several different notifications generated by the MPMusicPlayerController. I think the NotificationCenter and OperationQueue together are smart enough to short-circuit early if there are no observers for a notification, so there's no significant cost to posting a notification if no one's listening. The point is that there is precedent for not generating notifications at all by default and requiring the app to turn them on if it wants to use them. That would be an option here. I'm not sure if it's needed or even a good idea. It's just food for thought.

Examples/BareBones/BareBones/AppStatsViewController.swift
```diff
@@ -0,0 +1,92 @@
...
+import UIKit
+import Branch
+
+class AppStatsViewController: UIViewController {
+
+    @IBOutlet weak var statsLabel: UILabel!
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        NotificationCenter.default.addObserver(
```
jdee: You might want to call `NotificationCenter.default.removeObserver(self)` in deinit or viewWillUnload or something, as you did in AppStats. This is just a matter of best practices, since in practice the root VC never goes away, but probably a good idea in an example app, especially when adding a new feature.

BranchMacMini:
Result of [Integration 19](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/43b802e8876ab06629237126544ed7b6)
---
*Duration*: 9 minutes and 57 seconds
*Result*: All 204 tests passed, but please **fix 1 warning**.
*Test Coverage*: 62%

BranchMacMini:
Result of [Integration 20](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/4a950b5436627d63296817d0a011735b)
---
*Duration*: 9 minutes and 43 seconds
*Result*: **Build failed 1 test** out of 204

BranchMacMini:
Result of [Integration 21](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/4a950b5436627d63296817d0a0952788)
---
*Duration*: 10 minutes and 13 seconds
*Result*: All 204 tests passed, but please **fix 1 warning**.
*Test Coverage*: 62%

BranchMacMini:
Result of [Integration 22](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/4a950b5436627d63296817d0a0bfa182)
---
*Duration*: 10 minutes and 22 seconds
*Result*: **Build failed 2 tests** out of 204

BranchMacMini:
Result of [Integration 23](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/9d6ee9418c71092603096f17920f4309)
---
*Duration*: 10 minutes and 16 seconds
*Result*: **Build failed 2 tests** out of 204

BranchMacMini:
Result of [Integration 24](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/48d5d74b0c7e389b607bef2bcd983472)
---
*Duration*: 10 minutes and 5 seconds
*Result*: **Build failed 1 test** out of 204

BranchMacMini:
Result of [Integration 25](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/48d5d74b0c7e389b607bef2bcdeb7a58)
---
*Duration*: 10 minutes and 2 seconds
*Result*: All 204 tests passed, but please **fix 1 warning**.
*Test Coverage*: 62%

BranchMacMini:
Result of [Integration 26](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/f8df1cbd9bb77d91ba9bf0ea6e1091c2)
---
*Duration*: 9 minutes and 26 seconds
*Result*: All 204 tests passed, but please **fix 1 warning**.
*Test Coverage*: 62%

BranchMacMini:
Result of [Integration 28](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/f8df1cbd9bb77d91ba9bf0ea6e83a037)
---
*Duration*: 9 minutes and 28 seconds
*Result*: **Build failed 1 test** out of 204

BranchMacMini:
Result of [Integration 29](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/f8df1cbd9bb77d91ba9bf0ea6eb92689)
---
*Duration*: 9 minutes and 18 seconds
*Result*: **1 error, failing state: build-errors**

BranchMacMini:
Result of [Integration 30](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/be1c3435bf74e788e66de87a2201220a)
---
*Duration*: 9 minutes and 55 seconds
*Result*: All 204 tests passed, but please **fix 1 warning**.
*Test Coverage*: 62%

BranchMacMini:
Result of [Integration 31](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/be1c3435bf74e788e66de87a224f8dde)
---
*Duration*: 10 minutes and 36 seconds
*Result*: All 204 tests passed, but please **fix 1 warning**.
*Test Coverage*: 62%

BranchMacMini:
Result of [Integration 32](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/b50b91a0106aec66ea880379cf418cbd)
---
*Duration*: 12 minutes and 17 seconds
*Result*: All 204 tests passed, but please **fix 1 warning**.
*Test Coverage*: 62%

BranchMacMini:
Result of [Integration 34](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/b50b91a0106aec66ea880379cfb907a3)
---
*Duration*: 9 minutes and 14 seconds
*Result*: All 221 tests passed, but please **fix 1 warning**.
*Test Coverage*: 65%

BranchMacMini:
Result of [Integration 35](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/b50b91a0106aec66ea880379cfe029f6)
---
*Duration*: 9 minutes and 21 seconds
*Result*: All 221 tests passed, but please **fix 1 warning**.
*Test Coverage*: 65%

BranchMacMini:
Result of [Integration 36](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/68c5e063bf07377454b62d4229084e17)
---
*Duration*: 9 minutes and 19 seconds
*Result*: All 221 tests passed, but please **fix 1 warning**.
*Test Coverage*: 65%

BranchMacMini:
Result of [Integration 37](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/68c5e063bf07377454b62d4229c49a49)
---
*Duration*: 9 minutes and 29 seconds
*Result*: All 221 tests passed, but please **fix 1 warning**.
*Test Coverage*: 65%

BranchMacMini:
Result of [Integration 38](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/2b09c88671bcd5625e4de9189b7d0b42)
---
*Duration*: 11 minutes and 28 seconds
*Result*: **Build failed 1 test** out of 221

BranchMacMini:
Result of [Integration 39](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/2b09c88671bcd5625e4de9189baa6d41)
---
*Duration*: 10 minutes and 23 seconds
*Result*: All 221 tests passed, but please **fix 1 warning**.
*Test Coverage*: 65%

BranchMacMini:
Result of [Integration 40](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/f91d458930e1d7c7b58655c6ed7b3a04)
---
*Duration*: 10 minutes and 39 seconds
*Result*: **Build failed 1 test** out of 223

BranchMacMini:
Result of [Integration 41](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/f91d458930e1d7c7b58655c6eda48063)
---
*Duration*: 10 minutes and 28 seconds
*Result*: **Build failed 1 test** out of 223

BranchMacMini:
Result of [Integration 42](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/f91d458930e1d7c7b58655c6edccc3b2)
---
*Duration*: 9 minutes and 49 seconds
*Result*: All 223 tests passed, but please **fix 1 warning**.
*Test Coverage*: 65%

BranchMacMini:
Result of [Integration 43](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/1c453d4ef252b2829ed43be85aa45508)
---
*Duration*: 10 minutes and 53 seconds
*Result*: **1 error, failing state: build-errors**

BranchMacMini:
Result of [Integration 44](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/d581f40ccbec70312356d8c4ae161bc1)
---
*Duration*: 10 minutes and 18 seconds
*Result*: All 223 tests passed, but please **fix 1 warning**.
*Test Coverage*: 65%

BranchMacMini:
Result of [Integration 45](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c57398147d6bbaee710ac6bf10975bd1/ccee3a03a506642ef5411203d5150299)
---
*Duration*: 10 minutes and 25 seconds
*Result*: All 223 tests passed, but please **fix 1 warning**.
*Test Coverage*: 62%

PR #697, opened 2017-08-24
Added ability to turn logging on and off via links / Info.plist.

E-B-Smith:
Since we now use certificate pinning to secure our network traffic, Charles Proxy no longer works to debug app network problems.

I've added a way to turn logging on and off via crafted app links.  Using an app link like `myapp://?BranchLogLevel=BNCLogLevelAll` (or a universal link with the same query string) turns on the logging for the app.  The log messages can be read using the Console app or Xcode on a Mac.

Also, per @jdee , I added a way to set the log level in the app's Info.plist file.  Set the `BranchLogLevel` key in the Info.plist to the desired log level string.  The strings are defined in the `BNCLog.h`file.

### DIFF
--- a/Branch-SDK/Branch-SDK/Branch.m
+++ b/Branch-SDK/Branch-SDK/Branch.m
@@ -69,6 +69,7 @@ NSString * const BNCPurchasedEvent = @"Purchased";
 NSString * const BNCShareInitiatedEvent = @"Share Started";
 NSString * const BNCShareCompletedEvent = @"Share Completed";
 
+NSString * const BNCLogLevelKey = @"io.branch.sdk.BNCLogLevel";
 
 #pragma mark - Load Categories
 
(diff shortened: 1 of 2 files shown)


BranchMacMini:
Result of [Integration 1](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/4c35a3a414e39682e5d9ee1cca4cc2c7/4c35a3a414e39682e5d9ee1cca4cc4d3)
---
*Duration*: 9 minutes and 32 seconds
*Result*: **Perfect build!** All 199 tests passed. :+1:
*Test Coverage*: 60%

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -129,16 +130,30 @@ + (void) load {
...
+            BNCLogSetDisplayLevel(BNCLogLevelWarning);  // Default
+
+            // Try loading from the Info.plist
+            NSString *logLevelString = [[NSBundle mainBundle] infoDictionary][@"BranchLogLevel"];
+            if ([logLevelString isKindOfClass:[NSString class]]) {
+                BNCLogLevel logLevel = BNBLogLevelFromString(logLevelString);
+                BNCLogSetDisplayLevel(logLevel);
+            }
+
+            // Try loading from user defaults
```
jdee: I wonder about loading from user defaults. In particular, I'd expect the Info.plist to override the user defaults. If I make a change in the Info.plist, and then rerun in Xcode using this change, I believe the Info.plist change would be ignored. I'm not sure what value the user defaults add here. If you're setting it through the Info.plist, that will always be present. If you're setting it programmatically, that will also always be present. Is the concern that a logging URI like the one introduced here should persistently override the log level specified by the app? What if someone sends a debugging URI to a release build in production and forever turns on extremely verbose logging? Maybe it would be best if the debugging URI's effect were not persistent.

E-B-Smith: We need to be able to turn logging on in a production app so that Integrations Engineering can debug partner issues without using Charles proxy.

If the Info.plist overrides the user defaults then turning on the logging via Branch Link won't work when debugging issues that involve links that open a non-running app.

I think the possibility of turning on logging by accident is so low it would have to be a malicious attack, and the only side effect is that log messages would appear in the iPhone user's console log.

jdee: I see your point. No matter what workarounds you put in place for other builds, an AppStore release should have pinning enabled and logging disabled. This provides something.

I agree that the likelihood of this causing any serious problem is small.

:+1:

PR #691, opened 2017-08-22
CoreSpotlight implementation using searchable item [AIS-62]

parthkalavadia:
Spotlight index using Searchable Item implemented with following functionality:
- Index private BUOs using CSSearchableItem
- Index multiple BranchUniversalObjects using CSSearchableItem
- remove content from Spotlight if indexed using SearchableItem

PR #743, opened 2017-10-30
View controller manager

jdee:
Relevant to #691.

This is a utility class to find the current view controller for use with `presentViewController:animated:completion:`. The implementation is probably not too far from correct. Please review it. At least this allows implementing this logic in one place and reusing it throughout the SDK (and in wrapper SDKs). This can replace the implementation in Parth's PR that was borrowed from RN (and that can be replaced if this is released).

I also want to write some tests for this, but wanted to give everyone time to have a look.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCViewControllerManager.h
+++ b/Branch-SDK/Branch-SDK/BNCViewControllerManager.h
@@ -1,0 +1,31 @@
+//
+//  BNCViewControllerManager.h
+//  Branch
+//
+//  Created by Jimmy Dee on 10/30/17.
+//
+
+#import <Foundation/Foundation.h>
+
+/**
+ * Utility class to find the current view controller for use with presentViewController:animated:completion:.
+ */
+@interface BNCViewControllerManager : NSObject
+
+/**
+ * The current view controller to use.
+ *
+ * This can be nil in an extension, when UIApplication is not defined, if UIApplication.sharedApplication.keyWindow is nil
+ * or if the current view controller does not respond to presentViewController:animated:completion:.
+ */
+@property (readonly, nullable, nonatomic) UIViewController *currentViewController;
+
+/**
+ * Convenience method to present a view controller from the currentViewController.
+ * @param viewControllerToPresent an instance of UIViewController to present
+ * @param flag whether the presentation should be animated
+ * @param completion an optional completion block called once the presentation is complete
+ */
+- (void)presentViewController:(UIViewController * _Nonnull)viewControllerToPresent animated: (BOOL)flag completion:(void (^ __nullable)(void))completion;
+
+@end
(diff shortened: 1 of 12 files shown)


BranchMacMini:
Result of [Integration 1](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/68c5e063bf07377454b62d42299bf3b0/68c5e063bf07377454b62d42299bffec)
---
*Duration*: 10 minutes and 28 seconds
*Result*: All 220 tests passed, but please **fix 1 warning**.
*Test Coverage*: 64%

E-B-Smith:
This will be pretty useful.  I've needed code does something similar in several places in the SDK. 
  
Maybe instead of a whole class, a single function would do, with a function signature like `UIViewController* nullable BNCCurrentViewController()`, or a method on the Branch shared instance?

Here's one place I've put a lot of work in to finding the current view controller:

https://github.com/BranchMetrics/ios-branch-deep-linking/blob/master/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m#L258-L287

It's doesn't worry about alerts.

Here's where how it's used:

https://github.com/BranchMetrics/ios-branch-deep-linking/blob/master/Branch-SDK/Branch-SDK/BNCStrongMatchHelper.m#L323-L335

I'm happy to work with you on this.

jdee:
I really just wanted to get the ball rolling here after it came up a couple of times in PR reviews. I was actually hoping... you'd just do the work @E-B-Smith. :smile: (Kidding, kidding.)

This solution was obviously tailored to `presentViewController:animated:completion:`, which is a common need. Part of the idea behind the manager class was also to support operations directly, so you can just call the method on the manager and let it send the message to currentViewController. I see in the strong match helper you add the SFSafariVC as a child VC of the currently visible VC, or something like that. So that's a somewhat different need, though the solution to that problem ought to solve the first problem as well. When presenting with a container VC, you can present from the container or the child, so it's not necessary to dig out the actual visible VC. When you need that, you may also want to account for modal presentation by following the chain of `presentedViewController`s as here (the child of the container may be presenting another container, with a child presenting another VC, etc.).

I wonder if there's a more general solution to the container VC issue, since you have not only UINavigationController, UITabBarController, UISplitViewController but also others like UIPageViewController and any custom container VC. It's probably possible to iterate over the child view controllers with some possible exceptions for things like nav controllers where you have the `topViewController` property for simplicity.

I don't have a strong opinion about implementation. It's probably a good idea to start simple and iterate if necessary.

E-B-Smith:
It is probably possible to go the the UIResponder chain starting with the current firstResponder and find the "top" view controller that way.  It's a more general solution that will probably work, but I haven't done any testing with it, where as I have used and tested the StrongMatch helper version that has special cases.

E-B-Smith:
See PR #755 Current ViewController.

E-B-Smith:
I added the `vc` branch version of this that adds a category on `[UIViewController bnc_currentViewController]`.  I've tested code like that a lot and it is general purpose and handles view controllers with children view controllers.

PR #755, opened 2017-11-17
Current ViewController

E-B-Smith:
This is a category on UIViewController that gets the current view controller.

This is a version that I know works well and has been tested in a number of production apps.

It works with view controllers embedded in UINavigationControllers, UITabControllers, etc. It intentionally ignores alert views and system windows. 

The Swift call would be `if let vc = UIViewController.bnc_currentViewController ...`.


BranchMacMini:
Result of [Integration 2](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/46ac32330c688f9d4f468f418999caf5/46ac32330c688f9d4f468f418999dc92)
---
*Duration*: 11 minutes and 37 seconds
*Result*: All 223 tests passed, but please **fix 1 warning**.
*Test Coverage*: 63%

BranchMacMini:
Result of [Integration 3](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/46ac32330c688f9d4f468f418999caf5/46ac32330c688f9d4f468f4189c3ee24)
---
*Duration*: 11 minutes and 21 seconds
*Result*: All 223 tests passed, but please **fix 1 warning**.
*Test Coverage*: 63%

jdee:
I did see this, and it looks pretty good. I've just been busy getting the v2 event stuff into RN. I should have time for a real review by the end of the week.

E-B-Smith:
@jdee Bump. If you think this will work for you I'll merge it.

E-B-Smith:
@jdee Bump

jdee:
Sorry, I didn't see your bumps until now. My only concern is that in some cases, I've known `presentViewController:animated:completion:` to crash or at least not work if called on the `rootViewController` when that VC is presenting another VC, and so it is usually necessary to call the method on the `presentedViewController`, which could itself have a `presentedViewController`, and so on. I'm not sure what the current state of affairs is in iOS 11, but that was a concern in iOS 9, say. I'm also not sure `presentedViewController` is always reliable going back to iOS 7 and before.

However, this is a nice improvement, since it puts the logic in one place. Now if that logic needs to be modified, it can be modified in one place and reused, so :shipit:.

E-B-Smith:
The method I added works recursively so it should handle the case of a child view controller presenting a view controller that is presenting a view controller.  It's view controllers all the way down. Yikes.

PR #798, opened 2018-03-13
This seem to be fixing the memory leak

deepshah7:
I have tried profiling the BareBones project before and after applying this fix. Before the fix it certainly leaks memory. After the fix it didnt leak the memory. 

That said I am not sure if this is the right fix for the problem. Could someone please throw some light on this?

### DIFF
--- a/Branch-SDK/Branch-SDK/Networking/BNCNetworkService.m
+++ b/Branch-SDK/Branch-SDK/Networking/BNCNetworkService.m
@@ -219,6 +219,7 @@
                     operation.stringFromResponseData);
                 if (operation.completionBlock)
                     operation.completionBlock(operation);
+                _session = nil;
             }];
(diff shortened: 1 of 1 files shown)


BranchMacMini:
Result of [Integration 1](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/8f3e18ed1506d72c8b2f5a6edad129ab/8f3e18ed1506d72c8b2f5a6edad13020)
---
*Duration*: 16 minutes and 0 second
*Result*: All 227 tests passed, but please **fix 3 warnings**.
*Test Coverage*: 63%

E-B-Smith:
@sojanpr

E-B-Smith:
Check out #802 , Handle Network Session Dealloc Better.

I want to release the network session, but only when we're really done with it.

PR #808, opened 2018-03-19
Add Gmail subject line when sharing.

E-B-Smith:
This works, but only sometimes, apparently. Which is better tan what we have.  It won't hurt.


BranchMacMini:
Result of [Integration 1](https://stlt.herokuapp.com/v1/xcs_deeplink/qabot.stage.branch.io/c4a08d815bcaafc35de319aa0e22e5e5/c4a08d815bcaafc35de319aa0e22e8df)
---
*Duration*: 15 minutes and 31 seconds
*Result*: All 227 tests passed, but please **fix 3 warnings**.
*Test Coverage*: 63%

PR #829, opened 2018-05-01
Test-Fest Reorg

E-B-Smith:
* Reorganized project as minimally as possible to make building and testing apps much easier.

Do not squash-merge!


E-B-Smith:
I've made sure that everything compiles and that none of the SDK files have moved. Only the test cases have moved.

PR #832, opened 2018-05-08
Updated the UITests for easier automated testing.

E-B-Smith:

### DIFF
--- a/Examples/UITestBed/UITestBed-UITests/UITestBed.strings
+++ b/Examples/UITestBed/UITestBed-UITests/UITestBed.strings
@@ -1,0 +1,38 @@
+/* 
+  UITest.strings
+  UITestBed
+
+  Created by Edward on 5/7/18.
+  Copyright © 2018 Branch. All rights reserved.
+*/
+
+"UITestSafari" =
+"
+{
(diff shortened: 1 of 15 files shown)


Examples/UITestBed/UITestBed-UITests/UITestBed.strings
```diff
@@ -0,0 +1,38 @@
...
+  Copyright © 2018 Branch. All rights reserved.
+*/
+
+"UITestSafari" =
+"
+{
+  \"~referring_link\" : \"https:\\/\\/branch-uitestbed.app.link\\/YZIfk1NgJM\",
+  \"~creation_source\" : 3,
+  \"$one_time_use\" : false,
+  \"+is_first_session\" : false,
```
sojanpr: Can this be always `false` in the test environment?

sojanpr:
LGTM except for the above comment 👍

Examples/UITestBed/UITestBed-UITests/UITestBed.strings
E-B-Smith: If a special test was constructed, then yes. But in this case it will be no.

PR #835, opened 2018-05-14
Podspec: Add SafariServices for `without-IDFA` subspec

ziogaschr:
Updated subspec for podspec in order to include `SafariServices` framework with `without-IDFA` subspec for more reliable session matching.


E-B-Smith:
Thanks for the update.

PR #844, opened 2018-06-19
Prevent strong match controller crash DEVEX-515

E-B-Smith:
* Added extra safety to prevent a strong match controller crash.
* Updated project to Xcode-9.4.1.


sojanpr:
LGTM . 👍

PR #846, opened 2018-07-11
Fix inconsistency between Objective-C and Swift example

dstranz:
In Swift example method `application(_:continue:restorationHandler:)` returns always `true` and ignoring Branch response. It's fine in Objective-C example.


E-B-Smith:
Thanks for your update!

PR #848, opened 2018-07-31
Don't send IDFA when in debug mode. DEVEX-685

E-B-Smith:
* Don't send idfa for v2-events when in debug mode.
* Fixed some unit test conditions.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCDeviceInfo.m
+++ b/Branch-SDK/Branch-SDK/BNCDeviceInfo.m
@@ -430,7 +430,9 @@ exit:
     addString(osVersion,            os_version);
     addString(extensionType,        environment);
     addString(vendorId,             idfv);
-    addString(adId,                 idfa);
+    if (![BNCPreferenceHelper preferenceHelper].isDebug) {
+        addString(adId,             idfa);
+    }
     addString(browserUserAgent,     user_agent);
     addString(country,              country);
     addString(language,             language);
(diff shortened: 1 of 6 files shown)


sojanpr:
@E-B-Smith In case of Andorid, we are sending a  fake HWID and `is_hardware_id_real =false` while debug is enabled. Is there a reason for doing it different in iOS?

E-B-Smith:
@sojanpr  I'm following @ahmednawar request on this.  

We still send the `idfv`, which isn't a problem like `idfa` is.

On Android, you're sending `is_hardware_id_real` on v2-event too?

sojanpr:
@E-B-Smith  in case of v2-Events we don't add a fake id but add `unidentified_device= true` in the request

Branch-SDK/Branch-SDK/BNCDeviceInfo.m
```diff
@@ -430,7 +430,11 @@ - (NSDictionary*) v2dictionary {
     addString(osVersion,            os_version);
     addString(extensionType,        environment);
     addString(vendorId,             idfv);
-    addString(adId,                 idfa);
+    if ([BNCPreferenceHelper preferenceHelper].isDebug) {
```
sojanpr: Not sure if there any other condition we cannot get `adId`? if so can we add `unidentified_device` for that case also

sojanpr:
LGTM other than question above 👍

Branch-SDK/Branch-SDK/BNCDeviceInfo.m
E-B-Smith: If we can't get idfa in other cases it's because of privacy settings.  In that case we use idfv instead, and that's legit. So it isn't an `unidentified_device` device in that case, it's just a less identified device.

sojanpr: 👍

Issue #855, opened 2018-09-01
Error calling Branch Events in Objective-C


rodrigobdz:
I am currently implementing Branch Events in an app and while reading the docs I used a function call I found in the [README](https://github.com/BranchMetrics/ios-branch-deep-linking/blame/5dd137d3e224b7f0ddc7c3b4906c3a2f6a380e51/README.md#L607):

```objc
[BranchEvent.standardEvent(BranchStandardEventAddToCart) logEvent];
```

Xcode throws an error. Shouldn't it be called as shown below?

```diff
- [BranchEvent.standardEvent(BranchStandardEventAddToCart) logEvent];
+ [[BranchEvent standardEvent:BranchStandardEventAddToCart] logEvent];
```

E-B-Smith:
Yes it should!  Thanks, I'll update the docs. It looks like the Swift and obj-c examples got intertwined.

PR #856, opened 2018-09-02
README: Fix Objective-C function call

rodrigobdz:
Objective-C function was being called with a "swifty" syntax and caused an error in Xcode.
Similar to Issue #855 which was fixed in 6d6a9049e5c5e21ea77aa113d5e449bf1150155f.

PR submitted against `master` branch because it's a docs change.


E-B-Smith:
Thank you!

rodrigobdz:
Glad to help!

PR #859, opened 2018-09-06
DEVEX-739 iOS 12 Preperation

E-B-Smith:
* Use Cocoapods for OCMock and OHHTTP mocks: Frameworks aren't ABI compatible.
* Updated some unit tests for iOS 12.


aaaronlopez:
👍

PR #861, opened 2018-09-07
DEVEX-775 Add preview link scraping for Apple Reminders

E-B-Smith:
Added Apple Reminders as a scraper to prevent deferred linking.


aaaronlopez:
👍

PR #867, opened 2018-09-28
DEVEX-816 Facebook URI Fix

E-B-Smith:
* Fix JSON string encoding so that Facebook auth works right. 
* Update unit tests.


aaaronlopez:
👍

PR #872, opened 2018-10-09
DEVEX-808 change url and enable ATS

echo-branch:
Update url to https://api2.branch.io
Enable ATS in the testbed app


E-B-Smith:
In iOS we get TLS 1.2 for free!
We allowed arbitrary loads in the TestBed app for debugging network problems with certain proxies.

👍

PR #876, opened 2018-10-10
DEVEX-837 Updated the SDK to use secure coding.

E-B-Smith:


echo-branch:
Don't we also need to update any place we don't use the default initWithCoder?

`- (instancetype)initWithCoder:(NSCoder *)decoder {
    self = [super initWithCoder:decoder];
	if (!self) return self;

	self.commerceDictionary = [decoder decodeObjectForKey:@"commerceDictionary"];
	self.metadata = [decoder decodeObjectForKey:@"metaData"];
    return self;
}`

`id obj = [decoder decodeObjectOfClass:[MyClass class] forKey:@"myKey"];`

E-B-Smith:
Oh yeah. Updated.

echo-branch:
Looks good!

PR #882, opened 2018-10-17
DEVEX-861 Fix inconsistent unit tests.

E-B-Smith:
Unit tests were inconsistently failing and succeeding, mostly due to assumed app state. This PR fixes some of that.

Also, I think that iOS 12 unearthed some latent race condition bugs too.


echo-branch:
👍Code looks good and the unit tests pass for me.

PR #893, opened 2018-11-09
DEVEX-875 add new standard events

echo-branch:
**Changes due to updated spec**

1. Ad type constants are now CAPS  
2. BUO is sent
3. All events are routed through the same logic.  There is no difference between the custom and standard API.  This is how the current implementation works.

**Old description:**

These are new standard events for FB and Tune.  Most of these new standard events do NOT support the BranchUniversalObject.  Only RESERVE supports the BUO.

Also there is a behavior change.  Previously standard events created via the customEvent API would go to the standard server endpoint.  Now all events created via the customEvent API will go to the custom server endpoint.

All events created via BranchUniversalObject's userCompletedAction are treated as custom events.

### DIFF
--- a/Branch-SDK/Branch-SDK/BranchEvent.m
+++ b/Branch-SDK/Branch-SDK/BranchEvent.m
@@ -38,13 +38,13 @@ BranchStandardEvent BranchStandardEventUnlockAchievement      = @"UNLOCK_ACHIEVE
 
 // TODO: move these new events to appropriate sections
 
-BranchStandardEvent BranchStandardEventInvite = @"INVITE";
-BranchStandardEvent BranchStandardEventLogin = @"LOGIN";
-BranchStandardEvent BranchStandardEventReserve = @"RESERVE";
-BranchStandardEvent BranchStandardEventSubscribe = @"SUBSCRIBE";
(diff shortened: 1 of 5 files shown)


Branch-SDK/Branch-SDK/BranchEvent.m
```diff
@@ -161,9 +180,30 @@ - (void) setContentItems:(NSMutableArray<BranchUniversalObject *> *)contentItems
     }
 }
 
+// Objective-C does not have string enums, convert for the server
```
E-B-Smith: See here how to make obj-c 'string enums'.  

https://github.com/BranchMetrics/ios-branch-deep-linking/blob/7276778cf3ca1ec6af376950a37eed5290186348/Branch.framework/Versions/A/Headers/BranchEvent.h#L15

echo-branch: Removed misleading comment.  I was thinking string enums in other languages that have built in type enforcement.  If I were to use the obj-c string enum macro, I'd still have to validate.

echo-branch:
Spec has been updated.  There are significant changes since the original PR, see updates section in the description.

Issue #892, opened 2018-11-09
App install returns no params but after install and link click there is


hebertpalomino:
I'm having an issue where the app will return clicked_branch_link as false but after install and link click it shows and true and returns params. 

Link is: https://mens-essentials-mag.app.link/Ljzb4tWxGR

tested using both test and live environments (neither worked, although it was working a few days ago). 
A few things I did:
- I made sure to disable setUseTestBranchKey when not using test
- checked [Branch branchKey] to make sure right key was being used
- Tried different URL schemes and URI schemes in info and dashboard
- info's branch_app_domain is set to mens-essentials-mag.app.link
- Tried to run [[Branch getInstance] validateSDKIntegration] but get crash on `Checking for URI scheme correctness`
- in validateSDKIntegration I tried going to the URL it spits out ( [https://api2.branch.io/v1/app-link-settings/key_live_klv8U4RjmhgF66BqfWC4xgjissaM23vd?sdk=ios0.25.8&retryNumber=0](url) ) in browser and more info loads than in mobile safari on phone.. don't know if that's useful.. but something I noticed.

Let me know what other info you might need

hebertpalomino:
solved this by rolling back to 0.25.2
not sure if it was server issue or not but just in case this happens to anyone else

E-B-Smith:
Thanks for your great break down of your issue. 

The deferred deep links indeed may not have worked because the back end couldn't guarantee a match between your web browser session and the initial run of the app.  

I just tested deferred deep linking several times and it was working for me, so hopefully it will work for you too now.

I've made a test branch with fixes to the validator.  Would you be willing to test it with your configuration?

If you'd test your deep links too, to make sure any issues are taken care of, that would be great.

The fixes are on the `beta` branch of this GitHub repo.  Here are some instructions on setting your project to try this branch:
 
https://github.com/BranchMetrics/ios-branch-deep-linking/wiki/The-Branch-SDK-Beta-Project

PR #898, opened 2018-12-03
GH-892 Fix Validator Error

E-B-Smith:
Fixed a crash bug in the validator code.


echo-branch:
👍Code looks good and manual integration in the Monster Factory app looks good.

PR #899, opened 2018-12-04
SDK-76 nsexception in setBranchKey

echo-branch:
**Replace NSException with NSError in setBranchKey.**

Remove NSException use. This fixes the crash on misconfiguration.

Add an NSError out parameter. This allows runtime handling of branch initialization errors. Technically, they could have done it with a try catch, but Apple discourages that for typical errors.

Add a method that maintains the old API signature and just prints the error.

Add unit tests to verify changes.  This includes a new unhosted test target to avoid initialization done by our host app.  This test suite must run against simulator.

**Testing**

Check out branch.  Review and run unit tests in the Branch-SDK-Unhosted-Tests suite.


aaustin:
Looks great ! 👍

PR #908, opened 2019-03-14
Update BNCKeyChain kSecAttrAccessible value

JoeFerrucci:
Same effect as previous setting, but disallows being restored (from backup) to a _different_ device.
More secure this way.

When adding these items to the Keychain, the App does not explicitly exclude the items from device backups, which will allow sensitive data to get exported to iTunes and iCloud backups.

Having sensitive data sent to iCloud exposes it to both Apple, Inc. and an attacker with the ability to compromise the user's iCloud account (similarly to the iCloud celebrity hack, described in http://en.wikipedia.org/wiki/August2014celebrityphotoleaks/).

Additionally, exporting data to iTunes backups allows an attacker with physical access to the device and its passcode to use the iTunes encrypted backup functionality to extract all the Keychain items (that are flagged for backups) from the device.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCKeyChain.m
+++ b/Branch-SDK/Branch-SDK/BNCKeyChain.m
@@ -180,7 +180,7 @@ CFStringRef SecCopyErrorMessageString(OSStatus status, void *reserved) {
 
     dictionary[(__bridge id)kSecValueData] = valueData;
     dictionary[(__bridge id)kSecAttrIsInvisible] = (__bridge id)kCFBooleanTrue;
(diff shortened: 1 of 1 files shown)


JoeFerrucci:
@E-B-Smith @echo-branch - Would you mind having a quick look at this please?

echo-branch:
Hi,

The change looks reasonable, however we don't store sensitive data in keychain.  It's just install timestamps and bundle id to help with attribution.  We have a backlogged ticket to look into moving this out of keychain since there's no reason for it to be in a secure store.

JoeFerrucci:
Ok, thank you!

PR #916, opened 2019-04-19
Update Branch.m

cgx2012:
UIApplicationOpenURLOptionsSourceApplicationKey 和 UIApplicationOpenURLOptionsAnnotationKey 的定义是在ios9 以上，ios 8上访问会导致崩溃


echo-branch:
We used to support really old versions of Xcode and were unable to use the available check.  We recently we switched to a higher min Xcode version and can include these types of changes in the future.   

We have a backlogged ticket to go through and modernize this type of issue, I'll include this change when we do that ticket.

PR #921, opened 2019-07-05
SDK-301 remove search ads debug.  Apple provides a sample payload on …

echo-branch:
Remove redundant Apple Search Ads debug mode.  Apple provides the same type of placeholder data on all debug and simulator builds.  This feature should never be used in production.


Zhukn1:
Bumped into this today with another client and confirmed through Charles proxy that Apple is indeed sending the test data for every install coming in through the Testflight beta distribution system. 

Here's another discussion on the same topic:
https://stackoverflow.com/questions/50972230/applesearchingads-why-iad-attribution-is-always-true-for-dev-build

If we are going to merge this we probably also need to remove this method from the docs.

Issue #918, opened 2019-05-16
crash w/ branch 0.27.0


jaz-ah:
looks like you guys tried to make a fix in this area before - we're seeing many crashes in this area via crashlytics (no reproducible steps):

#0Crashed: com.apple.main-threadEXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x000000066a6f56b0 Raw Text
--
0 | libobjc.A.dylib | objc_msgSend + 16
1 | Branch | BNCEncodingUtils.m line 145+[BNCEncodingUtils sanitizedStringFromString:]
2 | Branch | BNCEncodingUtils.m line 208+[BNCEncodingUtils encodeDictionaryToJsonString:]
3 | Branch | BNCEncodingUtils.m line 185+[BNCEncodingUtils encodeDictionaryToJsonString:]
4 | Branch | BNCEncodingUtils.m line 159+[BNCEncodingUtils encodeDictionaryToJsonData:]
5 | Branch | BNCServerInterface.m line 777-[BNCServerInterface preparePostRequest:url:key:retryNumber:]
6 | Branch | BNCServerInterface.m line 567__75-[BNCServerInterface genericHTTPRequest:retryNumber:callback:retryHandler:]_block_invoke_2
7 | libdispatch.dylib | _dispatch_call_block_and_release + 24
14 | UIKitCore | UIApplicationMain + 212

echo-branch:
Thanks for bringing this to my attention.  I believe the issue is an unexpected key type is making it all the way to serialization.  I put in some protection around it and merged it to staging.

jaz-ah:
thanks @echo-branch - what is your ETA to getting this into a released sdk?

csalmi-branch:
We will hopefully be releasing a new version of the SDK in our next sprint cycle.

leandromperez:
👍 waiting for the fix here too!

jaz-ah:
@csalmi-branch is the fix in the 0.27.1 release?

csalmi-branch:
Yes! Sorry about the delay in notification. The fix has been released in v0.27.1 which is now available.

LucasVanDongen:
I'm still experiencing these crashes after updating to 0.27.1. I have no idea what strings cause them so I have no idea what to test against to begin with.

echo-branch:
[...]

PR #925, opened 2019-07-30
Sdk 415 GitHub reported crash around retry logic

echo-branch:
Crash reported around retryHandler, very likely due to a nil retryHandler block.  Add basic protection around it.


csalmi-branch:
This is for issue #918

PR #932, opened 2019-08-30
INTENG-7505 uiwebview deprecated

echo-branch:
Replace/remove UIWebView usage.  

1. Replaced UIWebView with Webkit.
2. Removed old code around Branch Views.

User agent is no longer a sync lazy loaded call.  Now it is loaded during Branch init.  First time it will be a bit slow, subsequent calls should be from storage.


echo-branch:
Not ready for full review yet.  The user agent code needs to be reworked since it relied on a clever hack to avoid dead lock.

PR #938, opened 2019-09-13
Add BNCThreads to public headers in carthage

evandcoleman:
This adds BNCThreads to the public headers in the framework built for Carthage. It is used by the Segment iOS integration https://github.com/BranchMetrics/Segment-Branch-iOS


echo-branch:
I'm going to remove that dependency from our Segment repo.  It never should have pulled in a project scoped utility class like that.

evandcoleman:
Awesome, thanks! I'll close this.

PR #945, opened 2019-09-25
INTENG-7729 race condition with search ads or slow init

echo-branch:
This is a quick fix for the issue.  I really want to refactor this as the current solution is difficult to maintain.


rkim-branch:
👍

PR #951, opened 2019-10-11
INTENG-7695 check for tune

echo-branch:
Repurposes the deprecated update flag with tune migration status information.

### DIFF
--- a/Branch-SDK/Branch-SDK/Networking/Requests/BranchOpenRequest.m
+++ b/Branch-SDK/Branch-SDK/Networking/Requests/BranchOpenRequest.m
@@ -18,6 +18,7 @@
 #import "Branch.h"
 #import "BNCApplication.h"
 #import "BNCAppleReceipt.h"
+#import "BNCTuneUtility.h"
 
 @interface BranchOpenRequest ()
 @property (assign, nonatomic) BOOL isInstall;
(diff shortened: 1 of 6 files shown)


Branch-SDK/Branch-SDK/Networking/Requests/BranchOpenRequest.m
```diff
@@ -99,42 +100,22 @@ - (void)makeRequest:(BNCServerInterface *)serverInterface key:(NSString *)key ca
...
-//  BNCUpdateStateError         = 3,    // Error determining update state.
-//  BNCUpdateStateReinstall     = 4     // App was re-installed.
+// Deprecated states, however older SDKs still send them.  The server will ignores these values.
+//    BNCUpdateStateInstall       = 0,    // App was recently installed.
+//    BNCUpdateStateNonUpdate     = 1,    // App was neither newly installed nor updated.
+//    BNCUpdateStateUpdate        = 2,    // App was recently updated.
+//    BNCUpdateStateError         = 3,    // Error determining update state.
+//    BNCUpdateStateReinstall     = 4,    // App was re-installed.
+    
+    BNCUpdateStateUnknown = 0,       // App update status is unknown
```
bklastaitis-branch: Seems like if an old SDK was sending `BNCUpdateStateInstall` it could potentially get interpreted as `BNCUpdateStateUnknown` since the int value is the same.

bklastaitis-branch: Also, since all other enum values were deprecated and it is now a two state enum, wouldn't it make sense to drop the use of the enum altogether and just use `[BNCTuneUtility isTuneDataPresent]` wherever the distinction is needed? It would prolly need to be renamed to something like `isFormerTuneUser`.

echo-branch: 1. Server has been ignoring values 0-4 for some time.  I'm just consolidating them into 0.
2. I can't switch away from enum since the server expects a numerical value in that field.  Even though it is effectively a boolean now.  

This only applies to iOS though.  On android that field is valid.

PR #979, opened 2019-12-06
improve BranchEvent's contentItems and customData properties

bryanmontz:
The contentItems and customData properties of BranchEvent are unnecessarily using mutable collection objects, rather than immutable ones.

Casting a NSDictionary to NSMutableDictionary does not actually make it a mutable dictionary. It just makes the compiler *think* that it is mutable, allowing the developer to call a method that only applies to mutable collections on an immutable one, which causes a crash.


echo-branch:
Looks good!

PR #980, opened 2019-12-06
INTENG-8009

bklastaitis-branch:
@echo-branch here is the current approach, as I said with this approach I still need to check that we do not auto-initialize when delayed initSession block exists. But just wanted to share this to discuss if the overall approach looks right.


echo-branch:
Looks good!

PR #993, opened 2020-01-17
INTENG-8456 allow overriding the placeholder url

echo-branch:
This does not have any unit tests as it's a UI change.  We should eventually add more UI tests.

### DIFF
--- a/Branch-TestBed/Branch-TestBed/AppDelegate.m
+++ b/Branch-TestBed/Branch-TestBed/AppDelegate.m
@@ -32,7 +32,7 @@ didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
     
     
     // test pre init support
-//    [self testDispatchToIsolationQueue:branch]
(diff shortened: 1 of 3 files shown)


Branch-TestBed/Branch-TestBed/AppDelegate.m
```diff
@@ -42,7 +42,7 @@ - (BOOL)application:(UIApplication *)application
     //[branch validateSDKIntegration];
 
     // Check for Apple Search Ad attribution (trade-off: slows down app startup):
-    [branch delayInitToCheckForSearchAds];
+    //[branch delayInitToCheckForSearchAds];
```
bklastaitis-branch: Intended?

echo-branch: Yes.  By default apple always sends a test payload and I was getting tired of seeing it.

PR #997, opened 2020-02-06
INTENG-8944 rare open crash

echo-branch:
I can demo the suspected root cause on a test app, I wasn't able generate a good unit test for it.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
+++ b/Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
@@ -778,19 +778,20 @@ static NSString * const BRANCH_PREFS_KEY_ANALYTICS_MANIFEST = @"bnc_branch_analy
             NSDictionary *persistenceDict = nil;
             @try {
                 NSError *error = nil;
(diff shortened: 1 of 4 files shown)


Branch-SDK/Branch-SDK/BNCPreferenceHelper.m
```diff
@@ -778,19 +778,20 @@ - (NSMutableDictionary *)persistenceDict {
             NSDictionary *persistenceDict = nil;
             @try {
                 NSError *error = nil;
-                NSData *data = [NSData dataWithContentsOfURL:self.class.URLForPrefsFile
-                    options:0 error:&error];
-                if (!error && data)
+                NSData *data = [NSData dataWithContentsOfURL:self.class.URLForPrefsFile options:0 error:&error];
```
bklastaitis-branch: it would be better to replace `0` with `NSDataReadingMappedIfSafe`

echo-branch: Probably, I didn't update the existing code.  Trying to resist the temptation to refactor a bunch of this.

PR #1007, opened 2020-03-03
SDK-842 disable ad network callout option 

echo-branch:
This is work in progress.  There's end to end testing and API documentation that's needed.

See ticket for details.


Branch-SDK-Tests/BNCDeviceInfoTests.m
```diff
@@ -140,6 +140,8 @@ - (void)testV2Dictionary {
     XCTAssertNotNil([dict objectForKey:@"os"]);
     XCTAssertNotNil([dict objectForKey:@"sdk"]);
     XCTAssertNotNil([dict objectForKey:@"sdk_version"]);
+    
+    XCTAssertNil([dict objectForKey:@"disable_ad_network_callouts"]);
```
jdee: Seems like this would all be more readable with more standard dictionary notation, e.g. `XCTAssertNil(dict[@"disable_ad_network_callouts"]);`. Just a general style/syntax comment. No need to change anything in this PR as far as I'm concerned.

Branch-SDK/Branch-SDK/BNCDeviceInfo.m
```diff
@@ -149,6 +149,11 @@ - (NSDictionary *)v2dictionary {
     @synchronized (self) {
         [self checkAdvertisingIdentifier];
         
+        BOOL disableAdNetworkCallouts = [BNCPreferenceHelper preferenceHelper].disableAdNetworkCallouts;
+        if (disableAdNetworkCallouts) {
+            [dictionary setObject:[NSNumber numberWithBool:disableAdNetworkCallouts] forKey:@"disable_ad_network_callouts"];
```
jdee: And:

```Obj-C
dictionary[@"disable_ad_network_callouts"] = @(disableAdNetworkCallouts);
```

(as just below, in fact).

Branch-SDK-Tests/BNCDeviceInfoTests.m
echo-branch: That's a good call.  I just kept going with the previous code, but really no need to.

PR #1009, opened 2020-03-13
SDK-886 `getDefaultUriScheme breaks if URI scheme used for Branch is not first`

illphil610:
This PR is a solution to SDK-886 where the validation fails if the URI scheme used for Branch is not the first in the plist file.  I did notice that `getDefaultUriScheme()` is used in more areas than just in the validation logic which is why I decided to create a new function to keep it separate to not interfere with internal logic.  I think since `getDefaultUriScheme()` only checks for the 1st entry we should refactor that to use the validated URI scheme that matches the server side URI scheme (if it exists) but didn't want to make that decision in this PR.  Just some thoughts!

Unit Testing Note:
I am having issues running unit tests locally.  I have unit tests created for the new methods but whenever I try to run them I am unable to build successful to run the tests thus I am not sure if they working correctly.  The tests are not included in this PR, but will be after I figure out the issue.  If we could have a working session to possibly solve my problem faster, I'd appreciate it!  Just wanted to give you a look into the solution in case I went down the wrong path to save time.

### DIFF
--- a/Branch-SDK/Branch-SDK/BNCSystemObserver.h
+++ b/Branch-SDK/Branch-SDK/BNCSystemObserver.h
@@ -19,6 +19,7 @@
                           andType:(NSString *__autoreleasing*)type;
 + (NSString *)getVendorId;
 + (NSString *)getDefaultUriScheme;
++ (NSArray *) getAllClientUriSchemes;
 + (NSString *)getAppVersion;
 + (NSString *)getBundleID;
 + (NSString *)getTeamIdentifier;
--- a/Branch-SDK/Branch-SDK/Branch+Validator.m
+++ b/Branch-SDK/Branch-SDK/Branch+Validator.m
@@ -28,6 +28,11 @@ static inline void BNCAfterSecondsPerformBlockOnMainThread(NSTimeInterval second
     dispatch_after(BNCDispatchTimeFromSeconds(seconds), dispatch_get_main_queue(), block);
 }
 
+typedef struct {
+    NSString *branchUriScheme;
+    BOOL doesMatchServerScheme;
+} ValidatedClientUriSchemeModel;
+
 #pragma mark - Branch (Validator)
 
 @implementation Branch (Validator)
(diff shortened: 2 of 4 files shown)


Branch-SDK/Branch-SDK/Branch+Validator.m
```diff
@@ -28,6 +28,11 @@ static inline void BNCAfterSecondsPerformBlockOnMainThread(NSTimeInterval second
     dispatch_after(BNCDispatchTimeFromSeconds(seconds), dispatch_get_main_queue(), block);
 }
 
+typedef struct {
+    NSString *uriScheme;
+    BOOL doesMatchServerScheme;
+} ValidatedClientUriSchemeModel;
```
echo-branch: Prefix the struct definition with BNC.

Branch-SDK/Branch-SDK/BNCSystemObserver.h
```diff
@@ -19,6 +19,7 @@
                           andType:(NSString *__autoreleasing*)type;
 + (NSString *)getVendorId;
 + (NSString *)getDefaultUriScheme;
++ (NSArray *) getAllClientUriSchemes;
```
echo-branch: If possible, let's type our collections in new code.

Branch-SDK/Branch-SDK/Branch+Validator.m
```diff
@@ -55,6 +60,29 @@ - (void) startValidation {
...
+ This method will loop through the clients URI schemes and compare each index with the uri scheme in Branch to look for a match.
+ @param clientUriSchemes NSArray representing each URI scheme present in the users plist file
+ @param serverUriScheme NSString representing the URI scheme Branch has for the client
+ @returns ValidatedClientUriSchemeModel : Struct which contains the necessary info to toggle validation logic in testing
+ */
+- (ValidatedClientUriSchemeModel) validateClientScheme:(NSArray*)clientUriSchemes withServerScheme:(NSString*)serverUriScheme {
+    ValidatedClientUriSchemeModel model;
+    model.doesMatchServerScheme = NO;
+    
+    for (id uriScheme in clientUriSchemes) {
```
echo-branch: Let's avoid id and use typed collections whenever possible.

echo-branch:
Let's add some unit tests around this.  I saw a dictionary length crash when testing in my test app.  Usually this is cause a dictionary is used instead of a string.  Didn't dig into it too much, but the error doesn't occur on stage.

Branch-SDK/Branch-SDK/Branch+Validator.m
E-B-Smith: Also can't be a struct: use an object instead.

Structs can't hold obj-c objects.

echo-branch: That should have triggered an error. :/

PR #1010, opened 2020-03-16
SDK-777 event status callback

echo-branch:
BranchEvent.logEvent with a callback.

Events with a callback are not part of the default retry logic, it will callback and report failure.

There should be no change in the existing BranchEvent.logEvent


Branch-SDK/Branch-SDK/BNCCallbackMap.h
```diff
@@ -0,0 +1,26 @@
...
+
+@interface BNCCallbackMap : NSObject
+
++ (instancetype)shared;
+
+- (void)storeRequest:(BNCServerRequest *)request withCompletion:(void (^_Nullable)(BOOL success, NSString *statusMessage))completion;
+
+- (BOOL)containsRequest:(BNCServerRequest *)request;
+
+- (void)callCompletionForRequest:(BNCServerRequest *)request withSuccessStatus:(BOOL)status message:(NSString *)statusMessage;
```
jdee: Did you think about using an `NSError *` instead of `NSString *`? There's lots of stuff on iOS where (in Swift) you often do something like:

```swift
(success: Bool, error: Error?) in

guard success, error == nil else {
    print("Error: \(error!.debugDescription)")
    MessageWindow.show(message: error!.localizedDescription)
    return
}
```

I'm not sure that's called for here, just a thought.

Branch-SDK/Branch-SDK/Branch.m
```diff
@@ -1873,7 +1874,20 @@ - (void) processRequest:(BNCServerRequest*)req
         error.code == BNCBadRequestError ||
         error.code == BNCDuplicateResourceError) {
 
-        BNCPerformBlockOnMainThreadSync(^{ [req processResponse:response error:error]; });
+        BNCPerformBlockOnMainThreadSync(^{
+            [req processResponse:response error:error];
+            if ([req isKindOfClass:[BranchEventRequest class]]) {
+                if (error) {
```
jdee: And then you could replace this whole block with:

```obj-c
[[BNCCallbackMap shared] callCompletionForRequest: req withSuccessStatus: (error == nil) error: error];
```

jdee:
Pretty sure I can pass that NSError straight up to React Native too.

Branch-SDK/Branch-SDK/BNCCallbackMap.h
echo-branch: Yea you're probably right.  I talked myself out of it for reasons that don't really apply in this situation.  I'll change it in the morning.

PR #1050, opened 2020-08-19
CORE-1209 SKAdNetwork

echo-branch:
Optional SKAdNetwork callouts.

If configured on the dashboard, the SDK will callout to SKAdNetwork on install, open or event.


Branch-SDK/BNCSKAdNetwork.m
```diff
@@ -0,0 +1,80 @@
...
+@property (nonatomic, strong, readwrite) Class skAdNetworkClass;
+@property (nonatomic, assign, readwrite) SEL skAdNetworkRegisterAppForAdNetworkAttribution;
+@property (nonatomic, assign, readwrite) SEL skAdNetworkUpdateConversionValue;
+
+@end
+
+@implementation BNCSKAdNetwork
+
++ (BNCSKAdNetwork *)sharedInstance {
+    static BNCSKAdNetwork *singleton;
```
jdee, line 25: Should this use `@synchronized(self)` around the whole thing? I think it is theoretically possible for two racing threads to try to initialize the `static` vars simultaneously.

PR #1060, opened 2020-09-29
Add files to interpret and implement branch.json

adrianyg7:
Add files to interpret and implement branch.json

### DIFF
--- a/Branch-TestBed/Branch-TestBed.xcodeproj/project.pbxproj
+++ b/Branch-TestBed/Branch-TestBed.xcodeproj/project.pbxproj
@@ -8,6 +8,12 @@
 
 /* Begin PBXBuildFile section */
 		1F241CE72639A862AD0840CF /* libPods-Branch-SDK-Tests.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 423ACD2DEBB9A2E4FAF64B08 /* libPods-Branch-SDK-Tests.a */; };
+		3A78D574251EB59B002A25CF /* BranchJsonConfig.h in Headers */ = {isa = PBXBuildFile; fileRef = 3A78D573251EB59B002A25CF /* BranchJsonConfig.h */; };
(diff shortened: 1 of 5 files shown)


Branch-TestBed/Branch-TestBed.xcodeproj/project.pbxproj
```diff
@@ -1212,7 +1223,7 @@
 						ProvisioningStyle = Automatic;
 					};
 					6700165F1940F51400A9E103 = {
-						DevelopmentTeam = R63EM248DP;
```
jdee, line 1215: We should probably back this out, at least manually (with vi or whatever). I can do that if it's tough otherwise.

jdee (commented):
Looks good. :+1: I guess we could stand to have some unit tests for the code I wrote and never bothered writing tests for. But I think this is the right idea. I'll leave it to Ernest to approve the PR.

Branch-TestBed/Branch-TestBed.xcodeproj/project.pbxproj
adrianyg7: Right, could you please do that?
I had to use our team and even when reverting outside xcode seems easy, I wouldn't like to mess with it.

jdee: Ernest or I can revert once this is merged, since this is from your fork. You can also use any text editor to modify the field in the file. You should still be able to build and deploy to a simulator. You can also perhaps go to a previous commit from before this change, copy out `Branch-TestBed.xcodeproj/project.pbxproj` to `/tmp` or anywhere, then copy it back in after reverting to this `HEAD` revision, depending if there were other changes to the file in the same commit. Not a big deal. There are numerous ways to deal with it now or later.

echo-branch (approved):
Looks good.  I'll cleanup the project settings on the staging branch.

Update.  Actually fixed it with the App Clip PR.

PR #1061, opened 2020-10-02
CORE-1185 app clips data

echo-branch:
Add data sharing between an App Clip and a subsequent full App install.

Testing for this is currently a manual integration of an app clip followed by a full app install.  We will need to add an integration test for it at some point.


Branch-SDK/Branch.h
```diff
@@ -676,6 +676,13 @@ typedef NS_ENUM(NSUInteger, BranchCreditHistoryOrder) {
  */
 - (void)ignoreAppleSearchAdsTestData;
 
+/**
+ Set the AppGroup used to share data between the App Clip and the Full App.
+ 
+ This must be set before initSession is called.
+ */
+- (void)setAppClipAppGroup:(NSString *)appGroup;
```
jdee, line 684: This seems like the kind of thing we could easily support via configuration as well, e.g. Info.plist (or branch.json).

jdee (approved):
:+1: Not sure if I'll be able to set up and test this. Might get to that, but looks good. The main thought is just that we could think of supporting the params for this in configuration, particularly the app group.

Branch-SDK/Branch.h
echo-branch: Yea agree.

PR #1066, opened 2020-10-12
Fix intermittent crash in `sanitizedStringFromString`

benski:
A fix for the issue described in https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/issues/918

We are using Branch in a large application and get a small handful of crashes in `[BNCEncodingUtils sanitizedStringFromString:]`

I looked through the code and I believe I have found a potential source to the problem. In `prepareParamDict`, the NSMutableDictionary `instrumentationDictionary` is returned without making a copy. Given the `@synchronized` code around access to this dictionary, it is likely to be updated by multiple queues at once.

The correct fix is to create a new method which makes a copy of this dictionary _while still within the @synchronized lock`!


echo-branch:
Thanks for the submission.  I'll take a look at it in the next sprint.

echo-branch (approved):

PR #1070, opened 2020-11-02
CORE-1354 make calls to the close api optional

echo-branch:
Add option to disable close requests.

Normally, this is set by a flag in the header of the install/open response. It is also possible to set in code for testing purposes.


Branch-SDK/BNCPreferenceHelper.m
```diff
@@ -597,6 +597,23 @@ - (void) setTrackingDisabled:(BOOL)disabled {
...
+        if ([b isKindOfClass:NSNumber.class]) return [b boolValue];
+        
+        // by default, we send close requests
+        return true;
+    }
+}
+
+- (void)setSendCloseRequests:(BOOL)disabled {
+    @synchronized(self) {
+        NSNumber *b = [NSNumber numberWithBool:disabled];
```
jdee, line 612: `NSNumber *b = @(disabled);`? Or even just `[self writeObjectToDefaults: @(disabled)];`? Not that it really matters.

jdee (approved):
:+1: All good. Minor stylistic point, but feel free to ignore. :D

Branch-SDK/BNCPreferenceHelper.m
echo-branch: updated

PR #1072, opened 2020-11-09
CORE-1354 switch default to not send close requests

echo-branch:
Change default behavior to not send close requests.


jdee (approved):
:+1:

PR #1074, opened 2020-12-12
CORE-1521 optional partner parameters for FB

echo-branch:
Facebook accepts partner parameters S2S.  For clients who wish to take advantage fo this, they may provide SHA256 hashed values to pass along to the FB API.  See Facebook documentation for details.

The Branch SDK sanity checks that the submitted value looks like a SHA256 hash.  
If present, these values will be passed up on install, opens and v2 events.  
We do not store these values, so they must be provided on every app start.
If the app needs to clear these values outside an app restart, we provide `clearPartnerParameters ` method.


echo-branch:
I need to add a button in the Testbed to exercise this feature.  This is still in progress.

Branch-SDK/BNCPartnerParameters.h
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ Parameters that clients wish to share with partners
+ */
+@interface BNCPartnerParameters : NSObject
+
++ (instancetype)shared;
+
+// FB partner parameters, see FB documentation for details
+// Values that do not looks like a valid SHA-256 hash are ignored
```
jdee, line 21: "do not look like"

Branch-SDK/BranchOpenRequest.m
```diff
@@ -85,6 +86,11 @@ - (void)makeRequest:(BNCServerInterface *)serverInterface key:(NSString *)key ca
                     forKey:BRANCH_REQUEST_KEY_SEARCH_AD
                     onDict:params];
     }
+    
+    NSDictionary *partnerParameters = [[BNCPartnerParameters shared] parameterJson];
+    if (partnerParameters.count > 0) {
+        [self safeSetValue:partnerParameters forKey:BRANCH_REQUEST_KEY_PARTNER_PARAMETERS onDict:params];
```
jdee, line 92: Just a general thought. I know this method has been around for a while, but it seems like this would make more sense as a category on NSDictionary, e.g. `[params safeSetValue:partnerParameters forKey:BRANCH_REQUEST_KEY_PARTNER_PARAMETERS]`. I think when using modern syntax though, you may be able to assign nil, i.e. `params[BRANCH_REQUEST_KEY_PARTNER_PARAMETERS] = partnerParameters // ok if it's nil`. But that's more a question of general refactoring.

jdee (approved):
:+1: LGTM

Branch-SDK/BranchOpenRequest.m
echo-branch: Yea I agree.  I think some of these old methods could use modernization.

PR #1084, opened 2021-01-31
SDK-1111 support error return for latd method

justinfurstossbranch:

### DIFF
--- a/Branch-SDK/Branch.h
+++ b/Branch-SDK/Branch.h
@@ -1120,7 +1120,7 @@ typedef NS_ENUM(NSUInteger, BranchCreditHistoryOrder) {
  @param window attribution window in days.  If the window is 0, the server will use the server side default.  If the window is outside the server supported range, it will default to 30 days.
  @param completion callback with attribution data
  */
-- (void)lastAttributedTouchDataWithAttributionWindow:(NSInteger)window completion:(void(^) (BranchLastAttributedTouchData * _Nullable latd))completion;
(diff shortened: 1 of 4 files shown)


echo-branch:
@justinfurstossbranch 
The class code change looks good, but the public API needs to be updated to match the change or a new API with error is required. 

https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/blob/master/Branch-SDK/Branch.m#L1288-L1293

justinfurstossbranch:
@echo-branch 
So i updated the latd method in Branch.m to handle the additional error return.

In regards to the public API. On Android we are able to support this :https://github.com/BranchMetrics/android-branch-deep-linking-attribution/blob/master/Branch-SDK/src/main/java/io/branch/referral/ServerRequestGetLATD.java#L86 
and they seem to be hitting the same endpoint: 
Android: https://github.com/BranchMetrics/android-branch-deep-linking-attribution/blob/master/Branch-SDK/src/main/java/io/branch/referral/Defines.java#L245
iOS: https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/blob/master/Branch-SDK/BranchConstants.m#L86

echo-branch:
Yea, I was referring to the SDK public API not the class public API.  Should have been clearer on that.

justinfurstossbranch:
Alright sweet! So my recent commit addresses that https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/pull/1084/commits/5202ece9ae1a7ba0a3bb1cfb11ca9cb798406c88 

Thanks @echo-branch !

PR #1085, opened 2021-02-08
CORE-1677 add static xcframework

echo-branch:
Add a static xcframework target. Also split deploy-release scripts into prep_release.sh and release_sdk.sh, to match the actual QA and release process. 

Running the scripts/prep_release.sh should build two xcframeworks and zip them up for distribution.
carthage-files/Branch.zip - contains dynamic xcframework. Supports iOS and tvOS
carthage-files/Branch_static.zip - contains static xcframework. Supports iOS only. tvOS does not support static frameworks.

You can verify the correct files were generated by running file on the component frameworks of the xcframework.


scripts/prep_release.sh
```diff
@@ -0,0 +1,56 @@
...
+echo "- Make sure that the ChangeLog.md has been updated, spell checked, and is coherent."
+echo "- Unit tests pass in Xcode"
+echo ""
+if ! ./scripts/askYN.sh "Build Branch release candidate version ${version}?"; then
+    echo ">>> Nothing built." 1>&2
+    exit 1
+fi
+
+# Prompt for editor input for ChangeLog.
+nano ChangeLog.md
```
jdee, line 38: One thing you might consider here is `$VISUAL ChangeLog.md` or `${VISUAL:nano} ChangeLog.md`, in case anyone else ever has to cut a release. I'm not familiar with nano and had to find my way out when trying this script. I think `VISUAL` is still valid, or maybe `EDITOR` to allow a user to override with `VISUAL=my-fave-editor scripts/prepare_release.sh`. But not a big deal if that doesn't seem to work well.

jdee (approved):
:+1:

scripts/prep_release.sh
echo-branch: Good call, i'll make that change.

PR #1086, opened 2021-02-09
SDK-1106 remove initialize use

echo-branch:
Removes a couple very old data copy methods when upgrading from SDK version 0.12.13.


jdee (approved):
:+1:

PR #1092, opened 2021-02-19
CORE-1711 reduce default SKAdNetwork timeout from 72 hours to 24 hours

echo-branch:
Reduce the default timeout to 24 hours.


jdee (approved):
:+1:

PR #1095, opened 2021-03-02
CORE-1715 att opt in status

echo-branch:
Add ATT opt in status to install and open. See jira ticket for details on test scenarios and sample payloads from an iOS 14.5 device.


Branch-SDK/BNCDeviceInfo.m
```diff
@@ -146,6 +146,7 @@ - (NSString *)userAgentString {
 
 // IDFA should never be cached
 - (void)checkAdvertisingIdentifier {
+    self.optedInStatus = [BNCSystemObserver attOptedInStatus];
```
jdee, line 149: So the idea is this provokes the prompt? Does it block until the user responds?

jdee (approved):
:+1: Looks good. Just a question.

Branch-SDK/BNCDeviceInfo.m
echo-branch: It never provokes the prompt.  It does check what the status is though, assuming the client didn't exclude the code via Linker flag.

echo-branch: Mostly cause the user could have changed it via menu instead of via prompt. So it's easier to just check it in a neutral way when we check for idfa.

PR #1104, opened 2021-03-22
CORE-1768 att opt in dialog event

echo-branch:
A public API to help measure ATT prompt performance.

### DIFF
--- a/Branch-SDK/Branch.h
+++ b/Branch-SDK/Branch.h
@@ -704,7 +704,7 @@ typedef NS_ENUM(NSUInteger, BranchCreditHistoryOrder) {
  Add a Partner Parameter for Facebook.
  Once set, this parameter is attached to install, opens and events until cleared or the app restarts.
  
- See Facebook's documentation for details on valid parameters0
+ See Facebook's documentation for details on valid parameters
  */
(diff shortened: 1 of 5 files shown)


Branch-SDK/Branch.h
```diff
@@ -683,6 +683,17 @@ typedef NS_ENUM(NSUInteger, BranchCreditHistoryOrder) {
...
+/**
+ Pass the AppTrackingTransparency authorization status to Branch to measure ATT prompt performance.
+ This method should be called from the callback of ATTrackingManager.requestTrackingAuthorization.
+ 
+ Note:
+ Before prompting the user, check that ATTrackingManager.trackingAuthorizationStatus is notDetermined.
+ Otherwise the prompt will not display and the completion will be called with current status.
+ This will inflate the number of OPT_IN and OPT_OUT events tracked by Branch.
+ */
+- (void)handleOptInStatus:(unsigned long)status;
```
jdee, line 695: Did you think of making this an `NSUInteger`? I think that would make it more transparent to Swift, as a `UInt` without casting.

jdee (approved):
Looks good. :+1:

Branch-SDK/Branch.h
echo-branch: Good point, I'll take a look at that.

PR #1114, opened 2021-05-05
CORE-1893 add timeout to apple attribution token

echo-branch:
We're seeing reports that the call to the Apple Attribution token api can hang.

Added a 100ms timeout to the Apple Attribution token call.


echo-branch:
Looking over the issue again, seems to be impacting simulators only. I'm thinking of replacing this PR with one that just skips the attribution token check on simulator.

jdee (approved):
:+1:

PR #1117, opened 2021-05-17
CORE-1916 fix keychain and queue

echo-branch:
Since we added support for secure coding on newer versions of iOS, we found archiving on Keychain and the network queue was failing due to mismatched types. This PR fixes both those issues.

For keychain, we only store NSDate. The code has been updated to always expect NSDate.

For the network queue, there was a refactor to improve testability around archive and unarchive of the network queue. The main change is that we only allow some classes to be archived and retrieved.  For older versions of iOS and the SDK, we filter on archive and retrieval.


jdee (commented):
:+1: Looks good. Sorry for the delay.

PR #1131, opened 2021-06-15
[CORE-1627] Trigger import.yml workflow in ios-spm repo on release

### LINKED ISSUE #1053: Git repository is very large, takes a long time to fetch using Xcode + Swift Package Manager
I was really excited to see Branch now supports SPM as of 0.35.0, however I noticed it now takes an awfully long time to resolve my dependencies.

Upon further inspection it looks like the repo is >750MB in size, owing to including binary frameworks inside the repository.

Xcode for some reason chooses to fetch the entire directory and therefore it takes an inordinate amount of time to download (over 7 minutes). 
Screen Shot 2020-08-24 at 2 55 43 PM

I'm not entirely sure what the solution is to this, but I wanted to raise the issue to hopefully stimulate a discussion.

If this repo was starting again I would respectfully suggest including binary frameworks in the releases section of the GitHub site, as opposed to including them in the repo itself. I realize this isn't an ideal solution now the repo has grown to this size. Hopefully someone has a good solution to this issue.

jdee:
Fix #1053

This change introduces a new public repository for use with Swift Package Manager. This effectively mirrors the current repo tag by tag, starting with version 0.35.0, the first release to support SPM. It excludes binaries from earlier releases and squashes together all commits between tags for minimal history and repo size, making for a much faster installation with SPM. The default branch here is `main`, not `master`.

This change to release.yml automatically triggers the manual import.yml workflow in that repo, which imports a specified tag. The mechanics here are subject to iteration.

In the long term, we'll deprecate this repo once SPM improves, e.g. providing support for shallow clones. We'll continue to support it indefinitely but will recommend that users no longer use this repo once it's no longer needed. The solution is  a bit brittle and likely to break from time to time, but this is the only way we can make things usable for devs using SPM in the near term.

~~Public docs updates on the way.~~
Public docs updated: https://help.branch.io/developers-hub/docs/ios-basic-integration#via-swift-package-manager

**Updated:** Renamed SPM repo to ios-branch-sdk-spm. Docs above also updated.

@BranchMetrics/core-team 

Screen Shot 2021-06-15 at 9 42 31 AM


.github/workflows/release.yml
```diff
@@ -168,4 +168,15 @@ jobs:
         uses: actions/github-script@v4
         with:
```
rkim-branch, line 169: We need to supply github-token. branchlet token should do, but that is internal token, so I recommend we create a new token that can have only access to our public repos

rkim-branch (approved):

rkim-branch (approved):

jdee:
I'm going to wait to merge this until @echo-branch has a chance to review when he's back next week.

ESA-Security-Group (approved):
#1137

PR #1159, opened 2021-11-04
CORE-2483 update build upload to include prebuild no idfa frameworks

echo-branch:
Add the no idfa binaries to the release script


GolfNoiise2020:
good

PR #1167, opened 2022-02-24
[SDK-1363] Changed the check for installParams

nsingh-branch:
Removed the check to see if storedParams is empty and now only checking isInstall


echo-branch (commented):
Just go ahead and delete the old code.

echo-branch (approved):

PR #1170, opened 2022-03-11
[SDK-1168] Added getLongAppLinkURLWithParams()

nsingh-branch:
Created a method that returns a long URL with app.link. This is used to properly show the app.link URL, link title, and link image in the Share Sheet preview

It fetches the user's app.link from preferenceHelper.userUrl, then shortens it to just the *.app.link. If one isn't available, then the usual bnc.lt long URL is generated like before.


ahmednawar:
Deferring to Ernest for the final review. LGTM

echo-branch (approved):

PR #1180, opened 2022-06-09
SDK-1391 : Don't Delete response fields from shared preference if the…

NidhiDixit09:
…y are empty

Updated class BranchOpenRequest and BranchSetIdentity to save session_id, randomized_bundle_token (identity_id), randomized_device_token (device_fingerprint_id) and
link, in shared preferences, only if they are non-empty.


echo-branch (commented):
Code looks good. Did we have tests around this? I know our request objects have low testability so we might need to revisit it.

NidhiDixit09:
I tested it manually by changing the response data while debugging. Turned off internet, so there was no response data. Then commented error checks in - (void) processRequest:(BNCServerRequest*)req response (BNCServerResponse*)response error:(NSError*)error method. This way processResponse methods were called with empty response fields.

PR #1181, opened 2022-06-14
Add missing `LinkPresentation` dep to podspec

mattrobmattrob:
https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/pull/1122 added the dependence on `LinkPresentation`. This adds it so `weak_framework` since it's only available in iOS 13.0 and up.


nsingh-branch (approved):

nsingh-branch:
Thank you for the contribution. We'll include this in the next release.

PR #1183, opened 2022-06-16
Change `URLSession` delegate queue to have `background` QOS

devioustree:
The current QOS of `userInteractive` is meant for interactive UI only. Network responses don't meet this bar and seem more suited to `background` which is meant for systems that the user is unaware of.

This also removes the runtime API check since the Branch SDK only supports iOS versions with the QOS API.

Fixes: https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/issues/1182


echo-branch:
@devioustree 
I like the change. Feel like we could just omit the `qualityOfService` line altogether since it should default to `NSQualityOfServiceBackground`.

PR #1196, opened 2022-09-09
SDK-1625 Protect BNCDeviceInfo

echo-branch:
## Reference
SDK-1625
INTENG-15831

## Summary
Updated BranchPluginSupport to safely use BNCDeviceInfo. This mirrors the code within BNCDeviceInfo itself.

## Motivation
Reports of a rare crash in this area code when using Adobe Launch iOS.

## Type Of Change

## Testing Instructions
The crash is fairly rare, the only reliable repro we've seen is using AdobeLaunch iOS on a real device running an older version of iOS. We suspect this is due to a threading issue where BNCDeviceInfo is being mutated from multiple threads.

### DIFF
--- a/Branch-SDK/BranchPluginSupport.m
+++ b/Branch-SDK/BranchPluginSupport.m
@@ -28,28 +28,29 @@
 
 
 - (NSDictionary<NSString *, NSString *> *)deviceDescription {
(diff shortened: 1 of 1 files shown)


NidhiDixit09 (approved):
It Looks good.

PR #1199, opened 2022-10-04
Sdk 1480 - UIPasteControl Support

NidhiDixit09:
## Reference
SDK-1480 - UIPasteControl iOS 16 - Allow customers to use their own UIPasteControl and pass params to Branch SDK

## Summary
* Added API passPasteItemProviders
* Added class BranchPasteControl
* Updated class BranchOpenRequest and BranchInstallRequest to send localURL if present.
* Added a new view in Testbed App which has BranchPasteControl and UIPasteControl buttons.
* Added unit test for API passPasteItemProviders

## Motivation 

Provide support for iOS 16 new control - UIPasteControl

 ` // Pass pasteboard data
 [[Branch getInstance] passPasteItemProviders:itemProviders];`

Added class BranchPasteControl - wrapper over UIPasteControl

`CGRect rectBC = CGRectMake(0, 0, _branchPasteControlView.frame.size.width, _branchPasteControlView.frame.size.height);
        BranchPasteControl *bc = [[BranchPasteControl alloc] initWithFrame:rectBC AndConfiguration:nil];
        [_branchPasteControlView addSubview:bc];`

## Type Of Change
- [ X] New feature (non-breaking change which adds functionality)
- [ X] This change requires a documentation update

## Testing Instructions

* Run unit test - testPassPasteControl
* Click on the link https://bnctestbed.app.link/nativelink
* Install TestBed 
* Press button "Show Paste control view" -> Press UIPasteControl or BranchPasteControl button. Verify logs.


echo-branch:
Code looks good and I was able to integrate into my test app. Just doing some edge case testing now.

echo-branch:
Code looks good and I was able to integrate into my test app. Just doing some edge case testing now.

Branch-SDK/BNCPreferenceHelper.h
```diff
@@ -29,9 +29,10 @@ NSURL* /* _Nonnull */ BNCURLForBranchDirectory(void);
 @property (copy, nonatomic) NSString *sessionID;
 @property (copy, nonatomic) NSString *linkClickIdentifier;
 @property (copy, nonatomic) NSString *spotlightIdentifier;
-@property (copy, nonatomic)    NSString *universalLinkUrl;
-@property (copy, nonatomic)    NSString *initialReferrer;
+@property (copy, nonatomic) NSString *universalLinkUrl;
+@property (copy, nonatomic) NSString *initialReferrer;
 @property (copy, nonatomic) NSString *userUrl;
+@property (copy, nonatomic) NSString *localUrl;
```
echo-branch, line 35: I think we need to track whether we've sent this value up. BNCPreferenceHelper is persistent, so setting it here without later clearing it means we send it up with subsequent opens.

echo-branch: We could also just keep it in memory instead of saving it to preference helper, but that could have other side effects.

nsingh-branch (approved):

Branch-SDK/BNCPreferenceHelper.h
NidhiDixit09: @echo-branch OK. Will clear it after first successful open. I thought it will be useful to send it with every open.

NidhiDixit09: @echo-branch I have added check for, sending local_url only on first open after data is copied, here in this commit https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/pull/1199/commits/3beb379491320c691ed0749afd6a48042c672eeb

Branch-SDK/BranchOpenRequest.h
```diff
@@ -12,6 +12,7 @@
 @interface BranchOpenRequest : BNCServerRequest
 
 @property (nonatomic, copy) callbackWithStatus callback;
+@property (assign, nonatomic) BOOL clearLocalURL;
```
echo-branch, line 15: Kindof a minor thing, usually we use YES and NO instead of TRUE and FALSE. I don't think it makes a difference here.

echo-branch: This comment was for when we use clearLocalURL later on.

Branch-SDK/Branch.h
```diff
@@ -10,6 +10,7 @@
 @import Foundation;
 #else
 #import <Foundation/Foundation.h>
+#import <UniformTypeIdentifiers/UniformTypeIdentifiers.h>
```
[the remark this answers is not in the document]
NidhiDixit09: Yes ! For compile time ... we only support latest SDK, so it won't be an issue. For unliking, its used in passPasteboardItem, which is available for iOS 16 only.

Branch-SDK/BranchOpenRequest.h
echo-branch: In my testing this doesn't quite work as expected. The main issue is that repeated paste button presses result in copying the url again. We could try clearing it from the pasteboard when we're certain it's a branch link. Another option is to check if the local_url is the same, then ignore.

echo-branch (approved):

PR #1208, opened 2022-11-09
Sdk 1659 skan4.0

NidhiDixit09:
## Reference
SDK-1659 --SKAN 4.0

## Summary
 SKAN 4.0 Initial Commit
* Preferences will store value of current SKAN Window and HighestConversionValueSent
* BNCSKAdNetwork : Added wrapper for new SKAN 4.0 API - updatePostbackConversionValue:coarseValue:lockWindowcompletionHandler.
*  BNCSKAdNetwork : Helper functions added for calculating current window from install timestamp, for parsing SKAN related keys from response Data. These functions will return default values if keys are not found in response. 
* BNCSKAdNetwork : Added function to decide whether updatePostbackConversionValue will be called based on currentSKANWindow and EnforceHighestConversionValue.
* BranchEvent : If iOS > 16.1, then new SKAN 4.0 API will be called.
* BranchOpen ( Install) - If its install, updatePostbackConversionValue will be called with default values - fineValue: 0, coarse : low, locked: false
* BranchOpen - Now updatePostbackConversionValue will be called for open also.

## Motivation
SKAN 4.0 Support

## Type Of Change

## Testing Instructions
This pull request in for code review only. Unit tests will be added in next PR. Integration testing can be done only after updated /v2/event and /v1/open APIs are available in staging


Branch-SDK/BNCSKAdNetwork.m
```diff
@@ -91,4 +101,94 @@ - (void)updatePostbackConversionValue:(NSInteger)conversionValue
...
+    }
+}
+
+- (int) calculateSKANWindowForTime:(NSDate *) currentTime{
+    
+    int firstWindowDuration = 2 * 24 * 3600;
+    int secondWindowDuration = 7 * 24 * 3600;
+    int thirdWindowDuration = 35 * 24 * 3600;
+    
+    NSTimeInterval timeDiff = [currentTime timeIntervalSinceDate:self.installDate];
```
jf-branch, line 122: @NidhiDixit09 to confirm self.installDate is currentInstallTimestamp and not first install timestamp correct?

Branch-SDK/BNCSKAdNetwork.m
```diff
@@ -91,4 +101,94 @@ - (void)updatePostbackConversionValue:(NSInteger)conversionValue
...
+    
+    if (timeDiff <= firstWindowDuration) {
+        return 1;
+    } else if (timeDiff <= secondWindowDuration) {
+        return 2;
+    }else if (timeDiff <= thirdWindowDuration) {
+        return 3;
+    }
+    return 0;
```
jf-branch, line 131: what is 0? we default to 0 for skanCurrentWindow, i would think 0 is the first window Duration and not 1, we can discuss this further during working session.

Branch-SDK/BNCSKAdNetwork.m
```diff
@@ -91,4 +101,94 @@ - (void)updatePostbackConversionValue:(NSInteger)conversionValue
...
+}
+
+- (NSString *) getCoarseConversionValueFromDataResponse:(NSDictionary *) dataResponseDictionary{
+    
+    NSString *coarseConversionValue = dataResponseDictionary[BRANCH_RESPONSE_KEY_COARSE_KEY] ;
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 160100
```
jf-branch, line 137: why are we using this syntax over     if (@available(iOS 16.1, *)) unless this is just how github is reporting it.

Branch-SDK/BranchOpenRequest.m
```diff
@@ -195,7 +194,55 @@ - (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
...
+                BOOL shouldCallUpdatePostback = [[BNCSKAdNetwork sharedInstance] shouldCallPostbackForDataResponse:data];
+                
+                BNCLogDebug([NSString stringWithFormat:@"SKAN 4.0 params - conversionValue:%@ coarseValue:%@, locked:%d, shouldCallPostback:%d", conversionValue, coarseConversionValue, lockWin, shouldCallUpdatePostback]);
+                
+                if(shouldCallUpdatePostback){
+                    [[BNCSKAdNetwork sharedInstance] updatePostbackConversionValue: conversionValue.longValue coarseValue:coarseConversionValue lockWindow:lockWin completionHandler:^(NSError * _Nullable error) {
+                        if (error) {
+                            BNCLogError([NSString stringWithFormat:@"Update conversion value failed with error - %@", [error description]]);
+                        } else {
+                            BNCLogDebug([NSString stringWithFormat:@"Update conversion value was successful. Conversion Value - %@", conversionValue]);
```
jf-branch, line 229: @NidhiDixit09 We should probably log all params, conversion, coarse, and lock window for fail safe debugging.

Branch-SDK/BNCSKAdNetwork.m
NidhiDixit09: @jf-branch Will discuss this during working session. Window 0 means There is no valid SKAN window, and it will help in initializing values.

Branch-SDK/BranchOpenRequest.m
NidhiDixit09: @jf-branch Check at line 222, all params are logged before functions call. However I think we should log current window also.

Branch-SDK/BNCSKAdNetwork.m
jf-branch: agreed, this makes sense!

Branch-SDK/BranchOpenRequest.m
jf-branch: agreed, and thanks for pointing 222 out, i see now.

Branch-SDK/BNCSKAdNetwork.m
NidhiDixit09: @jf-branch Imported it for enum values SKAdNetworkCoarseConversionValueHigh etc. We can replace them with string constants. But though Apple can change their values ... so its better to use enum constants. For function call we are using function pointers. We couldn't fetch pointers to string Constants. Correct me if I am wrong or there is way to do so.

echo-branch: After our discussion yesterday, I feel just using the raw string values will be ok.

NidhiDixit09: Yes ! In last commit, I did remove Storekit framework and using raw strings.

Branch-SDK/BNCSKAdNetwork.m
echo-branch: Maybe we should consider using an enum to make it a bit clearer that we mean 0 is no valid window.

Branch-SDK/BNCSKAdNetwork.m
```diff
@@ -91,4 +101,88 @@ - (void)updatePostbackConversionValue:(NSInteger)conversionValue
...
+
+    BOOL ascendingOnly = YES;
+    if([dataResponseDictionary[BRANCH_RESPONSE_KEY_ASCENDING_ONLY] isKindOfClass:NSNumber.class])
+        ascendingOnly = ((NSNumber *)dataResponseDictionary[BRANCH_RESPONSE_KEY_ASCENDING_ONLY]).boolValue;
+    return ascendingOnly;
+}
+
+- (BOOL) shouldCallPostbackForDataResponse:(NSDictionary *) dataResponseDictionary {
+    
+    BOOL shouldCallUpdatePostback = YES;
```
echo-branch, line 163: Minor thing, but I feel safer defaulting to NO on something like this and verifying that we should make the call.

echo-branch (approved):

PR #1209, opened 2022-11-25
INFRA-XX Auto-update GHA actions

kamil-branch:
Auto-generated GHA actions upgrade ios-branch-deep-linking-attribution. Need to upgrade actions, because of Node v.12 is deprecated (https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/) and set-output and save-state commands are deprecated  (https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -9,13 +9,13 @@ jobs:
     runs-on: macos-latest
     steps:
       - name: Check out code
-        uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


JagadeeshKaricherla-branch:
I think action version updates part of this PR could be better handled with dependabot with configuration file containing
`package-ecosystem: "github-actions"`

echo-branch (approved):

PR #1239, opened 2023-03-17
[WIP][minor][SDK-1800] Create BNCReferringURLUtility for URL query parameters

nsingh-branch:
## Reference
SDK-1800 -- [iOS] Update iOS SDK handling to extract GBRAID & GCLID from Branch referred uri/url

## Summary
Create a new class called `BNCReferringURLUtility` which makes it easier to parse query parameters from URL's and append them to requests. Currently it supports `gbraid` and `gclid`, but it was built to easily support new parameters without much more work. 

The class saves query parameters under `referringURLQueryParameters` in the PrefHelper in a dictionary that has the value name as a key and a `BNCUrlQueryParameter`(serialized to a dictionary) as a value. The `BNCUrlQueryParameter` contains a few basic fields like name, value, timestamp, validityWindow, and isDeepLink. 

The two main methods are `parseReferringURL()`, which parses referred Branch and non-Branch URLs for supported query parameters and saves/updates them. Then whenever a request is made, like v2/event or v1/open, `getURLQueryParamsForRequest()` is called to attach the respective parameters to the request body. This method just calls the other methods created for each query parameter and those determine which fields to add based on the request.

**WIP Notes** 
Currently `is_deeplink_gbraid` appears to always return true. Also `setReferrerGbraidValidityWindow()` may need to be updated further. Also needs further testing for upgrade scenarios. And old code and tests for Gbraid need to be removed or updated.

## Motivation

We already supported Gbraid as a URL query parameter, but needed to add Gclid support as well. So to make supporting future URL query parameters easier, we decided to make this new class to modularize things a bit more.

## Type Of Change

## Testing Instructions

Open a link with gbraid and gclid as query parameters and you should see them in the body of v2/event and v1/open requests. 
**Example Link** https://bnctestbed.app.link/?gbraid=12345&gclid=678910


BranchSDK/BNCReferringURLUtility.m
```diff
@@ -0,0 +1,205 @@
...
+            
+            result[param.name] = param;
+        }
+    }
+    
+    return result;
+}
+
+//Turns an old gbraid into a new BNCUrlQueryParameter and saves it for use, then clears it.
+-(void)checkForAndMigrateOldGbraid {
```
echo-branch, line 184: Let's either clear the old value, or check for a more recent value first. Don't want to migrate more than once.

echo-branch:
We also need to add this to the BranchSDK project used to generate frameworks and carthage compatibilty. Really should put a lifecheck on this.

nsingh-branch:
> We also need to add new files to the BranchSDK project used to generate frameworks and carthage compatibilty. Really should fix up the testbed situation soonish so this is easier.
Just added them to BranchSDK.xcodeproj. Is there anywhere else I may be missing them?

echo-branch (approved):

PR #1243, opened 2023-03-31
Qentelli test worklow

NidhiDixit09:
## Reference

SDK-1874-Run tests from qentelli-saas-sdk-testing-automation repo
https://branch.atlassian.net/browse/SDK-1874

## Summary
Updated file .github/workflows/automation-trigger-test.yml to
* Build and upload static lib which will be used by Qentelli Automation Testbed App (for linking with framework build with the current branch).
* Build Qentelli Automation Testbed App 
* Creat and upload IPA 
* Check out code from qentelli-saas-sdk-testing-automation and copy uploaded IPA to testdata/builds/iOS_TestbedApp_21-Jun-2022-1.ipa
* Run tests and upload report
Note : Last job `buildAndRunAutomationTestSuite:`  is a copy-paste from https://github.com/BranchMetrics/qentelli-saas-sdk-testing-automation/blob/master/.github/workflows/ios-maven.yml. I have used same steps for running tests using maven.

## Motivation
Integrate Qentelli Automation Tests with iOS SDK.

## Type Of Change

## Testing Instructions
Run workflow manually from the current Branch. Please see attached screen shot.
image


.github/workflows/automation-trigger-test.yml
```diff
@@ -1,18 +1,115 @@
...
+          path: |
+            ./Branch_static.zip
+            
+  # This job
+  # 1.Downloads static framework build in job build-Brach-Static-framework and copies it to framework directory of qentelli-saas-sdk-testing-ios-testbed repo
+  # 2.Builds testbed app.
+  # 3.Installs Certificates and Provisioning profile for signing App and creating IPA. Final IPA is uploaded as an artifact.
+  build-Automation-testbed-App:
+    runs-on: macos-latest
+    needs: [build-Brach-Static-framework]
```
echo-branch, line 28: Spelling error here. 
build-branch-static-framework

.github/workflows/automation-trigger-test.yml
```diff
@@ -1,18 +1,115 @@
-#on: [push]
+
 on:
  workflow_dispatch:
  
 jobs:
-  build:
-    runs-on: ubuntu-latest
+  # This job creates a static lib which will be used by Qentelli Automation Testbed App. Its uploaded as an artifact.
+  build-Brach-Static-framework:
```
echo-branch, line 7: spelling error here.
build-branch-static-framework

.github/workflows/automation-trigger-test.yml
```diff
@@ -1,18 +1,115 @@
...
+            unzip release_framework/Branch_static.zip -d ./Framework/
+            xcodebuild -workspace DeepLinkDemo.xcworkspace -scheme DeepLinkDemo  -sdk iphoneos archive -archivePath ./IPA/DeepLinkDemo.xcarchive
+            xcodebuild -exportArchive -archivePath ./IPA/DeepLinkDemo.xcarchive -exportOptionsPlist IPA/Info.plist -exportPath IPA/
+      - name: Upload build artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: DeepLinkDemo.ipa
+          path: |
+                ./IPA/DeepLinkDemo.ipa
+  buildAndRunAutomationTestSuite:
```
echo-branch, line 78: let's rename the phases in a consistent manner. The first two use `-` names while this one uses camel case. I personally have no preference

echo-branch:
I'd like to rename the whole GHA from `.github/workflows/automation-trigger-test.yml` to something more informative and less verbose. Not sure what our options are here, for unit tests the GHA displays the PR title.

.github/workflows/automation-trigger-test.yml
```diff
@@ -1,18 +1,117 @@
...
-            workflow_id: 'ios-maven.yml',
-            ref: 'master'
-          })
+      - name: Check out code
+        uses: actions/checkout@v3
+      - name: Build xcframework
+        run: |
+            ./scripts/prep_static_xcframework.sh
+      - name: Upload build artifacts
+        uses: actions/upload-artifact@v3
```
JagadeeshKaricherla-branch, line 17: For the purpose of sharing data between jobs - unless there's a need to save/persist data after a workflow run, we can use cache via `actions/cache@v3` in place of upload and downloading artifacts.Using cache is in general faster than uploading and downloading artifacts.

.github/workflows/automation-trigger-test.yml
```diff
@@ -1,18 +1,117 @@
...
+          # import certificate to keychain
+          security import $CERTIFICATE_PATH -P "$P12_PASSWORD" -A -t cert -f pkcs12 -k $KEYCHAIN_PATH
+          security list-keychain -d user -s $KEYCHAIN_PATH
+
+          # apply provisioning profile
+          mkdir -p ~/Library/MobileDevice/Provisioning\ Profiles
+          cp $PP_PATH ~/Library/MobileDevice/Provisioning\ Profiles
+          
+      - name: Download Branch SDK lib, build testbed app and create IPA File
+        uses: actions/download-artifact@v3
```
JagadeeshKaricherla-branch, line 67: same as above. Use cache in place of artifacts.

.github/workflows/automation-trigger-test.yml
```diff
@@ -1,18 +1,117 @@
...
+      - name : Unzip and copy folder
+        run: |
+            unzip release_framework/Branch_static.zip -d ./Framework/
+            xcodebuild -workspace DeepLinkDemo.xcworkspace -scheme DeepLinkDemo  -sdk iphoneos archive -archivePath ./IPA/DeepLinkDemo.xcarchive
+            xcodebuild -exportArchive -archivePath ./IPA/DeepLinkDemo.xcarchive -exportOptionsPlist IPA/Info.plist -exportPath IPA/
+      - name: Upload build artifacts
+        uses: actions/upload-artifact@v3
```
JagadeeshKaricherla-branch, line 74: Instead of uploading ipa as an artifact, we can do one of these : 
- use cache which is in general faster than uploading & downloading artifacts to share data between jobs
- upload the ipa to the browserstack cloud with a custom id or upload and save the returned `app_url`. This way we can break the browserstack test execution into independently executable / failable steps. We can fail fast in case of an issue.
https://www.browserstack.com/docs/app-automate/appium/upload-app-from-filesystem
https://www.browserstack.com/app-live/rest-api

.github/workflows/automation-trigger-test.yml
```diff
@@ -1,18 +1,117 @@
...
+      - name : Unzip and copy folder
+        run: |
+            ls -la ./
+            cp -R DeepLinkDemo.ipa testdata/builds/iOS_TestbedApp_21-Jun-2022-1.ipa
+      - name: Build with Maven
+        id: build_maven
+        run: |
+           mvn clean test -DsuiteXMLFile=testng.xml -e -DInputFormat="json" -DInputFilePath="TestSelection_BrowserStack_iOS.json" -DAccessKey="${{ secrets.S3_ACCESS_KEY }}" -DSecretKey="${{ secrets.S3_SECRET_KEY }}" --log-file output.txt
+      - name: Print output
+        if: always()        # run this step even if previous step failed
```
JagadeeshKaricherla-branch, line 107: use `success() || failure() `instead of `always()` as always causes the step to always execute, and returns true, even when `canceled`.
Also, always will execute even when any of the steps above in the job fail meaning not just the success or failure of `build_maven` step but steps above it and in that case, you won’t have output.txt so its better to check
`if: success() || steps. build_maven.outcome == 'failure'`

.github/workflows/automation-trigger-test.yml
NidhiDixit09: @JagadeeshKaricherla-branch As per my understanding, Cache will retrieve same files. Here files are changing with every run. Cache is used for re-using same files between different runs, so they are faster.  For example,  we are using cache here - https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/blob/master/.github/workflows/release.yml for faster building of pods. Here it will not have any impact because they are re-generated with every run. Correct me if I am wrong.

NidhiDixit09:
@echo-branch Since your last approval, I did following updates in this PR -
 
* Updated workflow to point to master branch of SaaS SDK Automation Tests repo.
* Browserstack and AWS creds are added as repo secrets and they are passes to Tests as environment variables. So they are removed from command.
* Addressed review comments

.github/workflows/automation-trigger-test.yml
```diff
@@ -1,18 +1,122 @@
...
+          distribution: 'temurin'
+          cache: maven
+      - name: Setup Node.js environment
+        uses: actions/setup-node@v3.3.0
+      - name: Download artifact
+        uses: actions/download-artifact@v3
+      - name : Unzip and copy folder
+        run: |
+            ls -la ./
+            cp -R DeepLinkDemo.ipa testdata/builds/iOS_TestbedApp_21-Jun-2022-1.ipa
```
gdeluna-branch, line 106: Let me know if you need help with uploading and setting the app_url env var

PR #1253, opened 2023-05-08
[SDK-1944][patch] Add Developer ID to v1/install

nsingh-branch:
## Reference
SDK-1944 -- [iOS] Add Developer ID to v1/install

## Summary
This PR adds the identity field to the `v1/install` endpoint by utilizing a temporary variable to store the identity set by the `setIdentity()` method. This is necessary because the `v1/profile` endpoint is called after the `v1/install` endpoint.

## Motivation
The addition of the identity field to the `v1/install` endpoint aims to achieve parity between the iOS and Android SDKs. This improvement has also been requested by clients.

## Type Of Change

## Testing Instructions

1. Call the `setIdentity()` method.
2. Run the app as a fresh install.
3. Observe the identity field in the `v1/install` request, as well as in subsequent `v1/open` requests as expected.
4. Ensure that the `v1/profile` endpoint is still called as usual after the `v1/install` endpoint.


BranchSDK/Branch.m
```diff
@@ -166,6 +166,7 @@ @implementation Branch
 #pragma mark - Public methods
 
 #pragma mark - GetInstance methods
+@synthesize installUserId;
```
echo-branch, line 169: This is likely unnecessary, properties should autosynthesize.

echo-branch:
Let's chat about this design. I'd prefer if this wasn't a new thing for clients to keep in mind, but rather something the SDK does internally when we realize the setIdentity method is called out of order.

echo-branch:
Though technically it's different, from the product point of view they should be the same.

echo-branch:
@nsingh-branch Reminder to add whatever tests that you can. This area of code probably needs a testability refactor, but I don't want to expand scope too much here.

echo-branch (approved):

PR #1259, opened 2023-05-30
Fix for GHA time out issue on tvOS 

NidhiDixit09:
## Reference
SDK-1964 -- iOS SDK GHAs sometimes waits indefinitely for tvOS simulators.
https://branch.atlassian.net/browse/SDK-1964

## Summary
Changed disable tracking to false. True was making integration tests to hang on GH runners for tvOS. Though its working fine locally. 

## Type Of Change

## Testing Instructions
Run GHA "Pre Release SDK Integration Tests". It should finish tests execution in  15-20 minutes
For fake/empty commits use following command - 
git commit --allow-empty -m "Trigger Build for testing"


echo-branch (approved):
Looks safe and fixes the test runner. I would like to add some basic tests and possible remove the disable tracking one since it's appears flakey when running in parallel anyway. But we can do that in a follow up ticket.

PR #1268, opened 2023-07-17
[SDK-1997] Update share sheet to use short links with tracking is disabled

nsingh-branch:
## Reference
SDK-1997 -- Update share sheet methods to still create a short link when tracking is disabled

## Summary
Update the share sheet methods to still generate short links when tracking is disabled instead of long links. By removing the initialization status check, the SDK will make a `v1/url` call instead of constructing a long link.

## Motivation
Currently, `getShortUrl()` already can be used to create a short link when tracking is disabled, but the share sheet methods would still create a long link. This brings parity between the two and overall improves the share sheet behavior.

## Type Of Change

## Testing Instructions

Trying to share a link via the share sheet and using BranchShareLink while tracking is disabled and enabled.

### DIFF
--- a/BranchSDK/Branch.m
+++ b/BranchSDK/Branch.m
@@ -1729,61 +1729,50 @@ static NSString *bnc_branchKey = nil;
                      andParams:(NSDictionary *)params
                 ignoreUAString:(NSString *)ignoreUAString
              forceLinkCreation:(BOOL)forceLinkCreation {
-
+    
     NSString *shortURL = nil;
-
+    
     BNCLinkData *linkData =
-        [self prepareLinkDataFor:tags
-            andAlias:alias
-             andType:type
-    andMatchDuration:duration
-          andChannel:channel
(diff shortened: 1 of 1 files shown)


echo-branch (commented):
I'd like to sync on offline behavior of this change. Ideally, we'd be able to keep the callee informed about what's going on, but the design is rather fire and forget which makes it awkward.

NidhiDixit09 (approved):

nsingh-branch:
> I'd like to sync on offline behavior of this change. Ideally, we'd be able to keep the callee informed about what's going on, but the design is rather fire and forget which makes it awkward.

@echo-branch When offline, the SDK will make a long link and still return it in the completion handler.

echo-branch:
Ok that sounds good! Let's go ahead and merge this as well.

echo-branch:
Ok that sounds good! Let's go ahead and merge this as well.

PR #1273, opened 2023-07-29
SDK 2005 iOS placeholder stop calling v1 profile and v1 logout

NidhiDixit09:
## Reference
SDK-2005 -- Stop calling /v1/profile and /v1/logout.

## Summary
Store and clear the identity locally without making requests to the endpoints.

## Type Of Change
- [ ] New feature (non-breaking change which adds functionality) --( Its not a breaking feature right now )

## Testing Instructions
-> All unit tests should pass
-> Following scenarios should work-
 *  User should be able to set identity and then retrieve it later.
 *  User should get back first referring params via callback.
 *  User should be able to logout. On logout only identity is cleared. First referring params and latest referring params are not cleared.
 * Ensure No network calls are made for endpoints /v1/profile and v1/logout. 


echo-branch:
@jf-branch Just want to get your thoughts on the edge case of calling the setIdentity prior to init. Is it ok to have no first referring params when called prior to init? In the past, we used caching to workaround server limitations which happened to make the first params available. However, that code is complex and we'd like to get rid of it.

nsingh-branch (approved):

echo-branch:
@NidhiDixit09 Let's go ahead and remove the caching behavior and simplify the code. Not returning first referring params if it's called prior to init is ok.

NidhiDixit09:
@echo-branch Caching behavior is removed.

echo-branch:
On install I'm seeing an issue with no developer id, think it's due to the old cache variable being checked directly in `BNCServerInterface.m`.

echo-branch:
@NidhiDixit09 The last change looks to have fixed the install issue. There are some merge conflicts that need to be resolved, but other than that it looks good.

PR #1282, opened 2023-08-24
[SDK-2068] Added ScCid support and tests

nsingh-branch:
## Reference
SDK-2068 -- Collect, Persist, and Forward 'ScCid' value

## Summary
Added support for collecting the Snapchat Click Identifier (ScCid) as a URL Query parameter. The parameter is collected and attached to each install, open, and v2/event request. Since ScCid is is session based, it is cleared whenever the app is backgrounded or terminated. 

## Motivation
To collect and forward the new ScCid URL query parameter.

## Type Of Change

## Testing Instructions

Open a link like `https://branchtestbed.app.link/testlink&ScCid=c22f092d-0961-4dff-9110-009df25e1328` and observe that the ScCid is persistent and attached to the proper requests, then cleared appropriately. 


echo-branch:
@nsingh-branch Did we ever hear back from product about our concerns about this behavior?

BranchSDK/BNCReferringURLUtility.m
```diff
@@ -25,10 +26,25 @@ - (instancetype)init {
         self.preferenceHelper = [BNCPreferenceHelper sharedInstance];
         self.urlQueryParameters = [self deserializeFromJson:self.preferenceHelper.referringURLQueryParameters];
         [self checkForAndMigrateOldGbraid];
+        
+        NSNotificationCenter *notificationCenter = [NSNotificationCenter defaultCenter];
```
echo-branch, line 30: Don't love this design, since it duplicates the session handling elsewhere. However, we can come back to it when we update session handling in general instead of making this more complex now.

echo-branch (approved):

BranchSDK/BNCReferringURLUtility.m
echo-branch: Maybe add a comment describing that it is session based instead of time window like the others.

echo-branch (approved):

PR #1286, opened 2023-08-29
[SDK-2084] Create new callback for NSData in BNCCallbacks

nsingh-branch:
## Reference
SDK-2084 -- Create new callback for NSData in BNCCallbacks

## Summary
A new callback, `callbackWithData`, has been added to BNCCallbacks to support QR codes in Unity. 

## Motivation
For the Unity SDK to use the QR code methods, a new callback with NSData is required. The changes have already been made one the Unity side, so now the iOS SDK just needs to be updated so it can be referenced. 

## Type Of Change

## Testing Instructions

Confirm there were no unintended changes from this.


rob-gioia-branch (approved):

echo-branch:
Is it necessary to do this as a typedef? Xcode doesn't resolve them nicely and it's usually better to just inline the block.

nsingh-branch:
> @nsingh-branch Is it necessary to do this as a typedef? Xcode doesn't resolve them nicely and it's usually better to just inline the block.

@echo-branch I'm not exactly sure, but it looks like Unity defines the callbacks in the `BranchiOSWrapper.mm files`, so i just went with the convention here and used a typedef.

echo-branch:
Ok. We can just go with this to be consistent with the existing code. We have a backlogged ticket to move away from typedef'd blocks as they're not autocompleted by Xcode, while inline versions are.

echo-branch (approved):

PR #1288, opened 2023-09-01
SDK-1982 SKAN Simplification DB Updates

NidhiDixit09:
## Reference
SDK-1982 -- [iOS] SKAN Simplification DB updates .
https://branch.atlassian.net/browse/SDK-1982 

## Summary
* Added parameter for current SKAN Window (skan_postback_index) in /v1/open and /v2/event requests
* Updated function 'shouldCallPostbackForDataResponse' to handle negative conversion values for second and third SKAN Windows.

## Motivation
Support SKAN 4.0 DB Simplication.

## Type Of Change

## Testing Instructions
With Every INSTALL/OPEN/EVENT requests, Verify using logs SDK is sending current SKAN window number. 
Verify in response , SDK is getting back conversion values.
Some of the samples are here -
https://www.notion.so/brancheng/SDK-SKAN-Testing-Results-5a74041cd815457292725673097f66c4


BranchSDK/BranchEvent.m
```diff
@@ -351,6 +351,19 @@ - (NSDictionary *)buildEventDictionary {
         eventDictionary[BRANCH_REQUEST_KEY_PARTNER_PARAMETERS] = partnerParameters;
     }
     
+    if (@available(iOS 16.1, *)){
```
echo-branch, line 354: Can we add a maccatalyst check as well.

BranchSDK/BNCSKAdNetwork.m
```diff
@@ -183,7 +183,11 @@ - (BOOL) shouldCallPostbackForDataResponse:(NSDictionary *) dataResponseDictiona
     }
     
     int highestConversionValue = (int)[BNCPreferenceHelper sharedInstance].highestConversionValueSent;
-    if( conversionValue.intValue <= highestConversionValue ){
+    if((currentWindow == BranchSkanWindowFirst) && (conversionValue.intValue <= highestConversionValue)){
+        BOOL ascendingOnly = [self getAscendingOnlyFromDataResponse:dataResponseDictionary];
+        if (!ascendingOnly)
+            shouldCallUpdatePostback = YES;
+    } else if ((currentWindow != BranchSkanWindowFirst) && (highestConversionValue != 0) && (conversionValue.intValue <= highestConversionValue)){
```
echo-branch, line 190: Is this second check just to prevent calling more than once with conversion value 0?

NidhiDixit09: @echo-branch  Now second and Third windows can have negative values (-1 (high), -2(medium), -3(low)) and default value for highestConversionValue (sent so for - used for ascendingOnly feature) is 0 for each window which is always higher than those negative conversion values. So in that scenario, if window is second or third and highest conversion value is 0 , it should go to last `else` condition and save that value in preferences.

BranchSDK/BranchEvent.m
NidhiDixit09: @echo-branch Done.

BranchSDK/BNCSKAdNetwork.m
echo-branch: So the second check is cause we default `highestConversionValueSent` to 0, which would prevent negative conversion values from being saved. What if we default it to `NSIntegerMin` instead?

echo-branch (approved):

echo-branch (approved):

PR #1295, opened 2023-09-18
SDK-2061 request payload refactor

echo-branch:
## Reference
SDK-2061 Request payload refactor

## Summary
Refactored server request generation out of BNCServerInterface to simplify data collection rules and enforcement. 

## Motivation
Existing data collection is split across several classes, this made it challenging to enforce data collection rules based on different privacy settings. 

* Remove deprecated code and public APIs
* Make request generation easier to test and update
* Assess and remove unnecessary data collection

## Type Of Change

## Testing Instructions
See testing notion doc and follow up JIRA tickets for details. 


BranchSDK/BNCServerAPI.m
```diff
@@ -24,53 +24,74 @@ + (BNCServerAPI *)sharedInstance {
...
+    self = [super init];
+    if (self) {
+        self.useTrackingDomain = NO;
+        self.useEUServers = NO;
+        self.automaticallyEnableTrackingDomain = YES;
+    }
+    return self;
+}
+
 - (NSURL *)installServiceURL{
```
NidhiDixit09, line 37: @echo Can we return only NSString from these functions instead of NSURL? For now these functions are only used by `[BNCServerInterface postRequest ]` API which expects a String. We are using absoluteString in those functions calls to get string. For example - https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/blob/69c84d0724133ed16ca15ff7bddc189b32d71bff/BranchSDK/BranchOpenRequest.m#L52

BranchSDK/BNCPreferenceHelper.m
```diff
@@ -817,23 +803,6 @@ - (void) setInvokeRegisterApp:(BOOL)invoke {
     }
 }
 
-- (BOOL) useEUServers {
```
NidhiDixit09, line 820: @echo-branch Is this property not persistent?

echo-branch: I was thinking it would be set on app launch, either in code or config file.

BranchSDK/BranchOpenRequest.m
```diff
@@ -240,6 +171,7 @@ - (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
     
     NSString *string = BNCStringFromWireFormat(data[BRANCH_RESPONSE_KEY_RANDOMIZED_BUNDLE_TOKEN]);
     if (!string) {
+        // fallback to deprecated name. The old name was easily confused with the setIdentity, hence the name change.
```
NidhiDixit09, line 174: Duplicate comment

NidhiDixit09 (approved):

PR #1317, opened 2023-11-02
SDK 2051 remove api variations that confuse swift

echo-branch:
## Reference
SDK-2051 remove Objective-C APIs that confuse Xcode's Swift compatibility

## Summary
NSError's within callback blocks are converted to exceptions in Swift. This causes unusual behavior when two methods differ only by a NSError in the callback. 

Removed problem public APIs, also added an NSError to one version that did not have one. This is a public API change.

## Motivation
Improve usability for Swift developers.

## Type Of Change

## Testing Instructions
Attempt to use these APIs in a Swift app, the NSErrors should be converted to exceptions correctly.

### DIFF
--- a/Branch-TestBed/Branch-SDK-Tests/BranchUniversalObject.Test.m
+++ b/Branch-TestBed/Branch-SDK-Tests/BranchUniversalObject.Test.m
@@ -172,45 +172,6 @@
     XCTAssertTrue([s bnc_isEqualToMaskedString:mask]);
 }
 
-- (void) testDeprecations {
-    #pragma clang diagnostic push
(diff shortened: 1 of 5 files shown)


NidhiDixit09:
@echo How can we make showsharesheet API to throw an error ?

echo-branch:
> @echo How can we make showsharesheet API to throw an error ?

I'm not entirely sure. It's a direct passthrough of this error from Apple's API. https://developer.apple.com/documentation/uikit/uiactivityviewcontroller/completionwithitemshandler

The expected error scenarios don't use that field.

NidhiDixit09 (approved):

PR #1324, opened 2023-11-28
the SDK state clear happened with the old close call

echo-branch:
## Reference
INTENG-19489 fix organic warm open

## Summary
SDK init status was cleared by the close call, with it's removal we need to clear it elsewhere to maintain previous behavior.

## Motivation
Restore previous behavior around organic opens.

## Type Of Change

## Testing Instructions
1. Run a test app with the debugger attached.
2. See install or open on app run.
3. Background the app.
4. Foreground the app. Notice no network call to init.

Repeat with this change, and step 4 will have a network call to init.

### DIFF
--- a/Branch-TestBed/Branch-TestBed/Branch-TestBed-Info.plist
+++ b/Branch-TestBed/Branch-TestBed/Branch-TestBed-Info.plist
@@ -54,7 +54,7 @@
 	<key>branch_key</key>
 	<dict>
 		<key>live</key>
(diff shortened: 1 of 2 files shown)


NidhiDixit09 (approved):

nsingh-branch:
The fix looks good to me. However, it looks like this branch is using the old testbed branch_key so the requests are failing.

nsingh-branch (approved):

NidhiDixit09:
Code looks good to me. I can see opens when app is foregrounded.

PR #1345, opened 2024-02-05
(WIP) - SDK 2223 DMA Consent Params Implementation

NidhiDixit09:
## Reference
SDK-2223 -- Design and implement key/values to represent consent param

## Summary
* Added following API in Branch Class.
` + (void) setDMAParamsForEEA:(BOOL) eeaRegion AdPersonalizationConsent:(BOOL) adPersonalizationConsent AdUserDataUsageConsent:(BOOL) adUserDataUsageConsent;
`
* Branch SDK will save DMA Compliance Parameters, set by developer using above API, and will send to server with install, open and event requests. BranchPreferenceHelper class is updated to store and read these params.
* Added unit test for testing above API

## Motivation
Support new params added in Google Conversion API for DMA Compliance.

## Type Of Change

## Testing Instructions
* All unit tests should pass.
* Call API setDMAParamsForEEA and check 
         -- SDK is sending value of `eeaRegion`, with every `/v1/install` , `/v1/open`, `/v2/event` , `v1/pageview` request.
         -- SDKis sending value of `adPersonalizationConsent` / `adUserDataUsageConsent` with every `/v1/install` , `/v1/open`, `/v2/event` , `v1/pageview` request only if -
                      - `eeaRegion` is `true` **OR** 
                      - `eeaRegion` is `false` and `adPersonalizationConsent` / `adUserDataUsageConsent` is also set to `false`
            Ref : https://www.notion.so/brancheng/Google-DMA-Compliance-c19ed28d2ea74b2b9fa4d2d77f8cdcbe 


BranchSDK/BNCPreferenceHelper.m
```diff
@@ -791,6 +798,60 @@ - (void) setInvokeRegisterApp:(BOOL)invoke {
     }
 }
 
+- (BOOL) eeaRegionInitialized {
+    @synchronized(self) {
+        if([self readObjectFromDefaults:BRANCH_PREFS_KEY_DMA_EEA])
+            return true;
+        return false;
```
echo-branch, line 805: Minor preference to use the ObjC YES/NO over the C true/false. Effectively the same, but more consistent with Apple's style.

BranchSDK/Branch.m
```diff
@@ -540,6 +540,14 @@ + (void)setReferrerGbraidValidityWindow:(NSTimeInterval)validityWindow{
     }
 }
 
++ (void) setDMAParamsForEEA:(BOOL)eeaRegion AdPersonalizationConsent:(BOOL)adPersonalizationConsent AdUserDataUsageConsent:(BOOL)adUserDataUsageConsent{
+    @synchronized(self) {
```
echo-branch, line 544: I know our old code uses this, but I doubt it does anything useful. It locks on the `[Branch class]` object. If anything, locking on the `[BNCPreferenceHelper sharedInstance]` would make more sense. But even that is likely unnecessary since the public methods in that class are synchronized on the instance.

PR #1346, opened 2024-02-06
[SDK-2163] Added new BranchLogger class

nsingh-branch:
## Reference
SDK-2163 -- Replace BNCLog with a simpler class based solution

## Summary
This PR introduces a new logging class, `BranchLogger`, designed to replace the existing `BNCLog` utility. The goal is to simplify logging within the SDK by providing a clear, straightforward way to log messages at different levels (Debug, Info, Warning, Error), support conditional logging based on the set log level threshold, and provide clients with a log callback. 

This is phase one of this change which is just creating the new class and implementing it in a few places. Once approved, the next phase will remove the existing `BNCLog` class and replace all of its usage with `BranchLogger`.

## Motivation

The motivation behind this change is to make it easier to log at different levels throughout the SDK and replace the old BNCLog class. Additionally, this change provide clients with new features like enabling logs only above a certain level and the ability to receive logs via a callbacks instead.

## Type Of Change

## Testing Instructions

1. Enable logging in the SDK using the existing `[[Branch getInstance] enableLogging]` or the new `[[Branch getInstance] enableLoggingAtLevel:BranchLogLevelVerbose]`;
2. Observe the logs to ensure that messages are correctly formatted, include the calling class and method where applicable, and that only logs at or above the set threshold are displayed.
3. Also try the logCallback like this:
```
  [BranchLogger shared].logCallback = ^(NSString *message, BranchLogLevel logLevel, NSError * _Nullable error) {
        // This block will be called for every log message emitted by the Branch SDK.
         if (error) {
             NSLog(@"[BranchLog] Level: %lu, Message: %@, Error: %@", (unsigned long)logLevel, message, error.localizedDescription);
         } else {
             NSLog(@"[BranchLog] Level: %lu, Message: %@", (unsigned long)logLevel, message);
         }     
    };
```


Branch-TestBed/Branch-SDK-Tests/BranchLoggerTests.m
```diff
@@ -0,0 +1,110 @@
...
+#import <XCTest/XCTest.h>
+#import "BranchLogger.h"
+#import "Branch.h"
+
+@interface BranchLoggerTests : XCTestCase
+@end
+
+@implementation BranchLoggerTests
+
+- (void)setUp {
```
echo-branch, line 18: Use a local copy instead of the singleton when doing unit tests. Just less likely to run into issues.

BranchSDK/BranchLogger.m
```diff
@@ -0,0 +1,107 @@
...
+//  Created by Nipun Singh on 2/1/24.
+//  Copyright © 2024 Branch, Inc. All rights reserved.
+//
+
+#import "BranchLogger.h"
+#import <os/log.h>
+
+@implementation BranchLogger
+
+static BranchLogLevel _logLevelThreshold = BranchLogLevelDebug;
```
echo-branch, line 14: Lets make this a local variable of the logger

echo-branch (approved):

PR #1347, opened 2024-02-10
SDK-2208 replace bnc field defines

echo-branch:
## Reference
SDK-2208 replace BNC field defines.

## Summary
`BNCFieldDefines.h` was replaced with regular Objective-C code. Mostly equivalent code was added to a category. 

## Motivation
This is to unblock a fix for non-Xcode versions of SPM. 

## Type Of Change

## Testing Instructions
Existing functionality and unit tests should work.


echo-branch:
I merged this into the PR for tuist as this is blocking cocoapods functionality. We only need to merge the other one, but this isolates the change for this.
https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/pull/1349

PR #1368, opened 2024-03-26
Add safety fallback when we are unable to unarchive requests

echo-branch:
## Reference
SDK-2263

## Summary
On failure to unarchive our request queue, we do not remove the incorrect requests. This leads to repeated failures to unarchive.

Believe this issue can occur when upgrading to version 3.3 from some older versions.

## Motivation
Safer fallback when we fail to unarchive.

## Type Of Change

## Testing Instructions
We do not have internal repro. However we can mimic a failure to unarchive, will demonstrate during team working meeting.


Sources/BranchSDK/BranchEvent.m
```diff
@@ -134,8 +134,11 @@ - (instancetype)initWithCoder:(NSCoder *)decoder {
     self = [super initWithCoder:decoder];
 	if (!self) return self;
 
-	self.serverURL = [decoder decodeObjectOfClass:NSString.class forKey:@"serverURL"];
-	self.eventDictionary = [decoder decodeObjectOfClass:NSDictionary.class forKey:@"eventDictionary"];
```
gdeluna-branch, line 138: What was the representation of this object without the data type enumeration?

Branch-TestBed/Branch-SDK-Tests/BNCClassSerializationTests.m
```diff
@@ -0,0 +1,117 @@
...
+//  Created by Ernest Cho on 3/28/24.
+//  Copyright © 2024 Branch, Inc. All rights reserved.
+//
+
+#import <XCTest/XCTest.h>
+#import "BranchEvent.h"
+#import "BranchOpenRequest.h"
+#import "BranchInstallRequest.h"
+
+@interface BranchEvent()
```
gdeluna-branch, line 14: Do we need to test the write to disk portion?

Sources/BranchSDK/BranchEvent.m
```diff
@@ -134,8 +134,11 @@ - (instancetype)initWithCoder:(NSCoder *)decoder {
...
+    self.serverURL = [decoder decodeObjectOfClass:NSURL.class forKey:@"serverURL"];
```
gdeluna-branch, line 137: Since this failed before, and will succeed now, will link replay be a significant behavior change?

Branch-TestBed/Branch-SDK-Tests/BNCClassSerializationTests.m
echo-branch: I don't think so. This test suite is focused on checking secure coding behavior on a few classes. Everything else is considered out of scope.

Also testing the disk writes would require refactoring and possibly a separate test suite since our tests are run in parallel so a global data source such as disk would cause issues.

Sources/BranchSDK/BranchEvent.m
echo-branch: Nil

Sources/BranchSDK/BranchEvent.m
echo-branch: It's an edge case error handling change. I've reached out to product about it.

echo-branch: product would prefer to fix/test it

Branch-TestBed/Branch-SDK-Tests/BNCClassSerializationTests.m
```diff
@@ -0,0 +1,120 @@
...
+    BranchEventRequest *request = [[BranchEventRequest alloc] initWithServerURL:url eventDictionary:eventDictionary completion:nil];
+    return request;
+}
+
+- (void)testBranchEventRequestArchive {
+    BranchEventRequest *request = [self buildBranchEventRequest];
+
+    // archive the event
+    NSError *error = nil;
+    NSData *data = [NSKeyedArchiver archivedDataWithRootObject:request requiringSecureCoding:YES error:&error];
```
gdeluna-branch, line 54: `archivedDataWithRootObject` is deprecated https://developer.apple.com/documentation/foundation/nskeyedarchiver/1413189-archiveddatawithrootobject

Is there a replacement or is this fine for this purpose?

Sources/BranchSDK/BNCServerRequestQueue.m
```diff
@@ -284,6 +284,19 @@ - (void)retrieve {
...
 
+// It's been reported that unarchive can fail in some situations. In that case, remove the queued requests file.
+- (void)removeSaveFile {
+    NSURL *fileURL = [BNCServerRequestQueue URLForQueueFile];
+    if (fileURL) {
+        NSError *error;
+        [NSFileManager.defaultManager removeItemAtURL:fileURL error:&error];
+        
+        if (error) {
+            [[BranchLogger shared] logError:@"Failed to remove archived queue" error:error];
```
gdeluna-branch, line 295: Can you add a test to capture the fail cases?

Branch-TestBed/Branch-SDK-Tests/BNCClassSerializationTests.m
```diff
@@ -0,0 +1,120 @@
...
+
+@interface BNCClassSerializationTests : XCTestCase
+
+@end
+
+// Test serialization of replayable requests
+@implementation BNCClassSerializationTests
+
+- (void)setUp {
+    // Put setup code here. This method is called before the invocation of each test method in the class.
```
gdeluna-branch, line 31: Are these empty functions required?

Sources/BranchSDK/BNCServerRequestQueue.m
echo-branch: Not easily for a couple reasons. The class itself needs to be redesigned for testability. Second this scenario is a disk write which cannot be run in parallel with other tests without potentially impacting them.

Branch-TestBed/Branch-SDK-Tests/BNCClassSerializationTests.m
echo-branch: Objective-C's odd function naming where arguments break up the function name gets people all the time.

https://developer.apple.com/documentation/foundation/nskeyedarchiver/2962880-archiveddatawithrootobject

Branch-TestBed/Branch-SDK-Tests/BNCClassSerializationTests.m
echo-branch: I don't think they are, but they're auto generated boilerplate that I usually just leave.

gdeluna-branch (approved):

nsingh-branch (approved):

PR #1369, opened 2024-03-27
SDK-2309 cache link data when initialization is deferred

echo-branch:
## Reference
SDK-2309 Cache link data when initialization is deferred

## Summary
On React Native we had a caching mechanism to store links until the JS layer was ready. This is not working properly when using the optional deferred init on iOS. This moves the caching to the native iOS layer.

## Motivation
Fix cold link open when using deferred init on iOS with React Native

## Type Of Change

## Testing Instructions
Enable deferred init with the `branch.json`
Add a delayed `notifyNativeToInit`
Install app, but make sure it is fully closed
Open the app with a universal link

Without this change the link will be ignored.
With this change the link will be used with the deferred init.

Also need to test this on React Native.


Sources/BranchSDK/Branch.m
```diff
@@ -1922,11 +1923,21 @@ - (void)initSafetyCheck {
...
+                [[BranchLogger shared] logDebug:@"Branch init is deferred, caching link"];
+                self.cachedURLString = urlString;
+            } else {
+                [[BranchLogger shared] logDebug:@"Branch init is deferred, ignoring lifecycle call without a link"];
+            }
             return;
+        } else {
+            if (!urlString && self.cachedURLString) {
+                urlString = self.cachedURLString;
+                self.cachedURLString = nil;
```
gdeluna-branch, line 1938: Move this line out of if so that the cached url will get cleared, in case of race with new url after deferred init is unlocked

Sources/BranchSDK/Branch.m
```diff
@@ -1922,11 +1923,21 @@ - (void)initSafetyCheck {
 
 - (void)initUserSessionAndCallCallback:(BOOL)callCallback sceneIdentifier:(NSString *)sceneIdentifier urlString:(NSString *)urlString {
     
-    // ignore lifecycle calls while waiting for a plugin runtime.
     @synchronized (self) {
         if (self.deferInitForPluginRuntime) {
-            [[BranchLogger shared] logDebug:@"Branch init is deferred, ignoring init call."];
+            if (urlString) {
```
gdeluna-branch, line 1928: Does this impact the ability to enqueue v2 events prior to init? Or will they be dropped still

echo-branch: No change from existing behavior. Though we may wish to evaluate if the existing behavior is acceptable in another ticket. SDK-2340

nsingh-branch:
What does the latest [Allow init calls to come in late.](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/pull/1369/commits/2009c84ae1b603d963a5e064c2f5eb7e97f6f242) commit do?

Sources/BranchSDK/Branch.m
```diff
@@ -624,27 +625,20 @@ - (void)initSessionWithLaunchOptions:(NSDictionary *)options
...
         id branchUrlFromPush = [options objectForKey:UIApplicationLaunchOptionsRemoteNotificationKey][BRANCH_PUSH_NOTIFICATION_PAYLOAD_KEY];
         if ([branchUrlFromPush isKindOfClass:[NSString class]]) {
             self.preferenceHelper.universalLinkUrl = branchUrlFromPush;
             self.preferenceHelper.referringURL = branchUrlFromPush;
+            pushURL = (NSString *)branchUrlFromPush;
         }
     }
     #endif
 
-    // Handle case where there's no URI scheme or Universal Link.
```
echo-branch, line 644: This code assumed that there would be a follow up link lifecycle call, but with delayed init that call may have already happened and the associated link saved off. 

Currently researching if this change has side effects.

echo-branch:
> What does the latest [Allow init calls to come in late.](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/pull/1369/commits/2009c84ae1b603d963a5e064c2f5eb7e97f6f242) commit do?

Added comment to code view. Still researching if this change is correct.

Sources/BranchSDK/Branch.m
echo-branch: After some testing this appears to be a legacy check that hasn't been required in a long time. It did negatively impact apps are are using just an `AppDelegate` on cold launch with a delayed init. Removing the check fixes the issue.

Sources/BranchSDK/Branch.m
```diff
@@ -624,27 +625,20 @@ - (void)initSessionWithLaunchOptions:(NSDictionary *)options
...
-    if (![options.allKeys containsObject:UIApplicationLaunchOptionsURLKey] && ![options.allKeys containsObject:UIApplicationLaunchOptionsUserActivityDictionaryKey]) {
-        [self initUserSessionAndCallCallback:YES sceneIdentifier:nil urlString:nil];
-    }
+    [self initUserSessionAndCallCallback:YES sceneIdentifier:nil urlString:pushURL];
```
gdeluna-branch, line 641: Can you add a comment if there are any assumptions about from which lifecycle call(s) this must go through?

gdeluna-branch (approved):

Sources/BranchSDK/Branch.m
echo-branch: Previously we assumed that initSession always came first, and other lifecycle calls would follow. With the delay feature, we cannot assume that anymore and cache away links arriving from other lifecycle calls. So now there are no assumptions about where the initSession is coming from.

PR #1372, opened 2024-03-28
SDK-2239 clean up logs

echo-branch:
## Reference
SDK-2239 clean up logs

## Summary
1. Reduce log noise by adjusting log levels of log message and simplifying the log message.
2. Allow NSError to be passed at any log level. This handles situations where NSError may be common, such as no network.
3. Make enableLogging a static method to capture SDK init issues. Deprecate the older instance versions, they still work though.
5. Unwrap network NSErrors so it's easier to determine the root cause. 
7. Add a helper method to check if a Network NSError looks like an Ad blocker.
8. Remove commented out test cases.

## Motivation
With switch to os_log, we noticed that logs were very noisy with inappropriate log levels for some logs. This is mainly reduce noise and make SDK debugging easier.

## Type Of Change

## Testing Instructions
All changes should be compatible with existing logging.

For testing with a DNS ad blocker, the device should route calls through Pi-hole or a similar application. The SDK should log a warning indicating that some calls are likely blocked.


Branch-TestBed/Branch-SDK-Tests/BranchLoggerTests.m
```diff
@@ -11,99 +11,251 @@
...
+    logger.logCallback = ^(NSString * _Nonnull message, BranchLogLevel logLevel, NSError * _Nullable error) {
+        count = count + 1;
+    };
+    
+    [logger logError:@"msg" error:nil];
+    XCTAssertTrue(count == 1);
+    [logger logWarning:@"msg" error:nil];
+    XCTAssertTrue(count == 2);
+    
+    // this should be ignored
```
gdeluna-branch, line 158: Can this comment be clarified? Ignored by who, can/should this be removed later?

Branch-TestBed/Branch-SDK-Tests/BranchLoggerTests.m
```diff
@@ -11,99 +11,251 @@
...
 
-- (void)testLogLevelThresholdBlocksLowerLevels {
+- (void)testLoggingEnabled_NOByDefault {
     BranchLogger *logger = [BranchLogger new];
-    logger.loggingEnabled = true;
-    logger.logLevelThreshold = BranchLogLevelDebug;
     
-    XCTestExpectation *expectation = [self expectationWithDescription:@"Log callback expectation for message that should pass the threshold"];
-
+    __block int count = 0;
```
gdeluna-branch, line 28: Is this thread safe? i.e when doing assertions, are there other tests that can run interference?

Branch-TestBed/Branch-SDK-Tests/BranchLoggerTests.m
```diff
@@ -11,99 +11,251 @@
...
+    BranchLogger *logger = [BranchLogger new];
+    logger.loggingEnabled = YES;
+    
+    __block int count = 0;
     logger.logCallback = ^(NSString * _Nonnull message, BranchLogLevel logLevel, NSError * _Nullable error) {
-        XCTAssertEqualObjects(message, expectedMessage, "Logged message does not match expected message.");
-        XCTAssertEqual(logLevel, expectedLevel, "Logged level does not match expected level.");
-        XCTAssertNil(error, "Error should be nil.");
-        [expectation fulfill];
+        count = count + 1;
```
gdeluna-branch, line 85: Some callbacks use `count++;`, others `count = count +1;` is there a preferred style?

Sources/BranchSDK/BNCApplication.m
```diff
@@ -69,64 +69,71 @@ + (BNCApplication*) createCurrentApplication {
...
     NSURL *appURL = nil;
     NSURL *bundleURL = [NSBundle mainBundle].bundleURL;
     NSDictionary *info = [NSBundle mainBundle].infoDictionary;
-    NSString *appName = info[(__bridge NSString*)kCFBundleExecutableKey];
+    NSString *appName = info[(__bridge NSString *)kCFBundleExecutableKey];
     if (appName.length > 0 && bundleURL) {
+        // path to the app on device. file:///private/var/containers/Bundle/Application/GUID
         appURL = [bundleURL URLByAppendingPathComponent:appName];
     } else {
+        // TODO: Why is this fallback necessary? The NSBundle approach has been available since iOS 2.0
```
gdeluna-branch, line 81: Is there a ticket tracking this investigation?

Sources/BranchSDK/BNCEncodingUtils.m
```diff
@@ -177,38 +176,30 @@ + (NSString *)encodeDictionaryToJsonString:(NSDictionary *)dictionary {
         id obj = dictionary[key];
         if ([obj isKindOfClass:[NSString class]]) {
             value = [BNCEncodingUtils sanitizedStringFromString:obj];
-        }
-        else if ([obj isKindOfClass:[NSURL class]]) {
+        } else if ([obj isKindOfClass:[NSURL class]]) {
```
gdeluna-branch, line 179: Nit: Personally i prefer if/else statements beginning on new lines since the collapse buttons on certain IDEs pushes the subsequent statement to the end of the line of the collapsed function. (not blocking) 

Screenshot 2024-04-01 at 7 18 33 PM
Screenshot 2024-04-01 at 7 18 41 PM

Branch-TestBed/Branch-SDK-Tests/BranchLoggerTests.m
echo-branch: Not really. When equivalent, `count++` is most commonly used in C. Google's style guide suggests `++count` as being more readable when prefix/postfix doesn't matter. I personally prefer being explicit in tests.

Sources/BranchSDK/BNCEncodingUtils.m
```diff
@@ -468,9 +458,12 @@ + (NSData *) dataFromHexString:(NSString*)string {
     }
 
     // If highValue != -1 then we got an odd number of hex values, which is an error.
-    if (highValue == -1)
+    if (highValue == -1) {
         data = [NSData dataWithBytes:bytes length:b-bytes];
+    }
 
+    // Error handling in C code is one case where goto can improve readability.
+    // https://www.kernel.org/doc/html/v4.19/process/coding-style.html
```
gdeluna-branch, line 466: (Out of scope) goto/label usage aside, the same label name is used throughout with different routines. For readability, the name of the label should be more specific than `exit`.

Branch-TestBed/Branch-SDK-Tests/BranchLoggerTests.m
echo-branch: Yep

Sources/BranchSDK/BNCPreferenceHelper.m
```diff
@@ -985,7 +985,7 @@ - (NSMutableDictionary *)deserializePrefDictFromData:(NSData *)data {
 
         dict = [NSKeyedUnarchiver unarchivedObjectOfClasses:classes fromData:data error:&error];
         if (error) {
-            [[BranchLogger shared] logWarning:@"Failed to load preferences from storage."];
+            [[BranchLogger shared] logWarning:@"Failed to load preferences from storage." error:error];
```
gdeluna-branch, line 988: Just for confirmation, can the log function work without passing in the error parameter? This hasn't been widely adopted yet right?

Sources/BranchSDK/BNCApplication.m
echo-branch: Theres a ticket in backlog.

gdeluna-branch (approved):
non-blocking feedback. looks good.

Sources/BranchSDK/BNCEncodingUtils.m
echo-branch: The Objective-C style guide is to put it on the same line. Similar to K&R and Linux kernel style. https://google.github.io/styleguide/objcguide.html#conditionals

Sources/BranchSDK/BNCEncodingUtils.m
echo-branch: Yea I wanted to do it, but felt like I was expanding scope of the ticket from clarifying logs to clarifying/fixing all kinds of little things. I stopped after adding those comments. Really should come back and make that change.

echo-branch: I'll make a follow up for this.

Sources/BranchSDK/BNCEncodingUtils.m
echo-branch: Xcode also doesn't have that neat feature to fold code nicely.

Sources/BranchSDK/BNCPreferenceHelper.m
echo-branch: Should have called this out in the PR. The reason I didn't leave a version with and without NSError is that it breaks Swift. 

The bridge between ObjC and Swift converts those NSErrors into Exceptions. So if you have both versions, you'll end up with a definition that throws and exception and a definition that does not. Although not breaking, this is very confusing to devs as there's two conflicting definitions and related warnings.

echo-branch: And although `BranchLogger` is a public class, it's really intended for our use in the SDK and any wrappers. The actual public log API for clients has not changed as that always had an NSError on it.

gdeluna-branch: I see so you'd have to use a different name for the wrapping function.

PR #1381, opened 2024-04-05
SDK-1217 workaround tvOS SPM issue

echo-branch:
## Reference
SDK-1217 workaround tvOS SPM issue

## Summary
Added preprocessor guards on all files that are not valid on tvOS. With other integration methods we can conditionally exclude these files, but it seems SPM does not allow this.

## Motivation
Allow tvOS builds using SPM.

## Type Of Change

## Testing Instructions
Integrate Branch SDK into a tvOS app using SPM.
App should build properly.


nsingh-branch (approved):
Ran a test with a tvOS app with and without this change, looks good to me.

PR #1388, opened 2024-04-22
SDK 2302 update singleton creation to dispatch once

echo-branch:
## Reference
SDK-2302

## Summary
Update to modern singleton creation syntax

## Motivation
Code modernization

## Testing Instructions
There should be no functional change

### DIFF
--- a/Sources/BranchSDK/BNCDeviceInfo.m
+++ b/Sources/BranchSDK/BNCDeviceInfo.m
@@ -34,7 +34,7 @@
 @implementation BNCDeviceInfo
 
 + (BNCDeviceInfo *)getInstance {
-    static BNCDeviceInfo *bnc_deviceInfo = 0;
+    static BNCDeviceInfo *bnc_deviceInfo = nil;
     static dispatch_once_t onceToken = 0;
     dispatch_once(&onceToken, ^{
         bnc_deviceInfo = [BNCDeviceInfo new];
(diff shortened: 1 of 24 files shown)


Sources/BranchSDK/BNCDeviceInfo.m
```diff
@@ -34,7 +34,7 @@ @interface BNCDeviceInfo()
 @implementation BNCDeviceInfo
 
 + (BNCDeviceInfo *)getInstance {
-    static BNCDeviceInfo *bnc_deviceInfo = 0;
+    static BNCDeviceInfo *bnc_deviceInfo = nil;
```
gdeluna-branch, line 37: Any difference in parsing/storage?

Sources/BranchSDK/BranchContentDiscoverer.m
```diff
@@ -27,12 +27,11 @@ @implementation BranchContentDiscoverer
 
 + (BranchContentDiscoverer *)getInstance {
     static BranchContentDiscoverer *sharedInstance = nil;
-    @synchronized (self) {
-        if (!sharedInstance) {
-            sharedInstance = [[BranchContentDiscoverer alloc] init];
-        }
+    static dispatch_once_t onceToken = 0;
+    dispatch_once(&onceToken, ^{
```
gdeluna-branch, line 31: Is this by default on the main thread?

echo-branch: `dispatch_once` doesn't switch threads or anything like that. 

It's a shorthand for locking around Singleton creation.
https://developer.apple.com/documentation/dispatch/1447169-dispatch_once

gdeluna-branch: I see, I saw the synchronized keyword and questioned if there was a threading assumption. But since dispatch once happens .. once, then this makes sense.

gdeluna-branch (approved):

PR #1408, opened 2024-07-12
fix: update metadata type to string

ssriram-branch:
## Reference

No JIRA ticket, based on Slack discussions

## Summary

The `setRequestMetadataKey` function used to accept any object in the value parameter. The function is now modified to take in only a string or nil as Branch APIs cannot process non-string values for metadata.

## Motivation

The SDK exposes the `setRequestMetadataKey` function to allow developers to send in any value that can be attached with all event metadata. This includes IDs, custom parameters etc. 

As of today, the function takes in a string key with any value (string/bool/object/integer) without any errors. However, on passing a non-string value for metadata, one can observe that all Branch APIs start failing with a 400 error, citing invalid JSON. This is because the Branch APIs are designed to take in only string values for metadata. This can lead to confusion and difficulty in debugging the issue because it is not immediately apparent that the issue is caused because of a non-string metadata value.

This PR aims to eliminate this confusion by updating the signature of the `setRequestMetadataKey` function to accept only string or nil values (nil is accepted so that the value can be unset).

## Type Of Change

## Testing Instructions

On a sample project, use this version of the SDK and try to set a value for metadata that is not null and not a string. The function will throw an error.

When trying the same experiment with an older version of the SDK, there is no compile time error but Branch SDK API calls constantly fail at run-time with a 400 "Invalid JSON" error.

### DIFF
--- a/Sources/BranchSDK/Branch.m
+++ b/Sources/BranchSDK/Branch.m
@@ -491,7 +491,7 @@ static NSString *bnc_branchKey = nil;
 }
 
 
-- (void)setRequestMetadataKey:(NSString *)key value:(NSObject *)value {
+- (void)setRequestMetadataKey:(NSString *)key value:(NSString *)value {
     [self.preferenceHelper setRequestMetadataKey:key value:value];
 }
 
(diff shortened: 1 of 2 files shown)


nsingh-branch (approved):
LGTM. Thanks again for making this change.

PR #1421, opened 2024-08-02
Universal ink cold launch fix

NidhiDixit09:
Fix for Universal Link Cold Launch

## Summary

Issues -

--   If tracking is disabled and app launch is cold, Universal Links were not resolving intermittently because of race condition happening between `branch.initSession` and `[BranchScene scene:continueUserActivity:]` for accessing preference variables. It happens only if both functions are called at the very same time. 
   Following is the sequence of events which happens 
   1. Session initialization starts from function `DidFinishLaunching` using `branch.initSession`. It has no `universal_link` in request. First Open Request is sent to server.
   2. In the meantime  - `[BranchScene scene:continueUserActivity:]` is called and it has `universal_link` url . It sets this value in prefs. So when First open request is going out - it checks if its a linking request (by checking url in prefs, which is just set by second request), function returns - YES.
  3. After First request is processed - it clears this url from prefs. Now when second Open request is sent - although request params have` universal_link_url` but its not anymore in prefs - its considered a non-linking request.

**Fix** : `isLinkingRelatedRequest` function will read values of identifiers from post params of request instead of preferences.

-- SDK keeps sending archived opens from the previous launch even after link is resolved

**Fix** :  A property is added in `BranchOpenRequest` class to find out if the event is from the archived queue. Another property `processArchivedOpens` is added in `BNCServerRequestQueue` to signal if archived opens should be propecessed.
After SDK gets `referringURL` in  `BranchOpenRequest` response, it turns of this flag. 

`processNextQueueItem` function uses these properties to find out if the next request is from archive and it should not be processed. If YES, it removes it from the queue and processes next one.

--  SDK resets session using enum variable `initializationStatus`. It sets its value to `BNCInitStatusUninitialized`.  Its a global variable and its value is sometimes changed by other parallel session initializations happening, if any,  to  `BNCInitStatusInitializing` or `BNCInitStatusInitialized`. So function `initUserSessionAndCallCallback`, which checks this value for session initialization, simply returns wihtout resetting/initializing session.

**Fix** : Added param in function `initUserSessionAndCallCallback` for forced session re-initialization irrespective of current SDK initializationStatus

## Testing Instructions

Use a test app with Universal Link Enabled and tracking disabled. 

--> If Test App is running kill it. Cold Launch is required.
--> Copy universal links for testing in iMessage.
--> Click on Univesal Links in iMessage.
--> Test App should be launched.
--> Check logs to see if Open request sent with param "universal_link_url" is processed and other opens without this params are dropped.

Repeat this test multiple times, by killing app and re-launching app using links from iMessage.

I have used this app https://github.com/BranchMetrics/SpotifyTest 

Links used for testing - 

```
https://cq9pf.app.link/FoVf9M9bDLb

https://cq9pf.app.link/fSMdfpldDLb

https://cq9pf.app.link/TfySwxRdDLb

```
Repeat test by enabling tracking as well as delaying  function calls `initSession` and` BranchScene.shared().scene`(...) 

## Type Of Change

### DIFF
--- a/Sources/BranchSDK/BNCServerRequestQueue.m
+++ b/Sources/BranchSDK/BNCServerRequestQueue.m
@@ -386,7 +386,7 @@ static inline uint64_t BNCNanoSecondsFromTimeInterval(NSTimeInterval interval) {
     static dispatch_once_t onceToken = 0;
     dispatch_once(&onceToken, ^ {
         sharedQueue = [[BNCServerRequestQueue alloc] init];
-        [sharedQueue retrieve];
+        //[sharedQueue retrieve];
         [[BranchLogger shared] logVerbose:[NSString stringWithFormat:@"Retrieved from storage: %@.", sharedQueue] error:nil];
     });
     return sharedQueue;
--- a/Sources/BranchSDK/BNCServerRequestQueue.m
+++ b/Sources/BranchSDK/BNCServerRequestQueue.m
@@ -42,6 +42,7 @@ static inline uint64_t BNCNanoSecondsFromTimeInterval(NSTimeInterval interval) {
 
     self.queue = [NSMutableArray<BNCServerRequest *> new];
     self.asyncQueue = dispatch_queue_create("io.branch.persist_queue", DISPATCH_QUEUE_SERIAL);
+    self.processArchivedOpens = YES;
     return self;
 }
 
--- a/Sources/BranchSDK/BNCServerRequestQueue.m
+++ b/Sources/BranchSDK/BNCServerRequestQueue.m
@@ -167,7 +167,8 @@ static inline uint64_t BNCNanoSecondsFromTimeInterval(NSTimeInterval interval) {
             BNCServerRequest *request = [self.queue objectAtIndex:i];
 
             // Install subclasses open, so only need to check open
-            if ([request isKindOfClass:[BranchOpenRequest class]]) {
+            // Request should not be the one added from archived queue
+            if ([request isKindOfClass:[BranchOpenRequest class]] && !((BranchOpenRequest *)request).isFromArchivedQueue) {
                 return (BranchOpenRequest *)request;
             }
         }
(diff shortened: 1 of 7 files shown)


Sources/BranchSDK/BNCServerRequestQueue.m
```diff
@@ -279,6 +281,11 @@ - (void)retrieve {
         NSData *data = [NSData dataWithContentsOfURL:self.class.URLForQueueFile options:0 error:&error];
         if (!error && data) {
             NSMutableArray *decodedQueue = [self unarchiveQueueFromData:data];
+            for (id request in decodedQueue) {
+                if ([request isKindOfClass:[BranchOpenRequest class]] || [request isKindOfClass:[BranchInstallRequest class]]) {
+                    ((BranchOpenRequest *)request).isFromArchivedQueue = YES;
```
gdeluna-branch, line 286: Does this type cast work if it's BranchInstallRequest?

NidhiDixit09: Yes.  BranchOpenRequest is superclass of BranchInstallRequest.

Sources/BranchSDK/BranchOpenRequest.m
```diff
@@ -151,6 +152,10 @@ - (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
             }
         }
     }
+    
+    if (referringURL.length > 0) {
```
gdeluna-branch, line 156: Should we instead use the function `isLinkingRelatedRequest`?

NidhiDixit09: That function is not availbale here. Also we dont store post params inside this object.

Sources/BranchSDK/Branch.m
```diff
@@ -1971,8 +1981,9 @@ - (void)initUserSessionAndCallCallback:(BOOL)callCallback sceneIdentifier:(NSStr
     
     dispatch_async(self.isolationQueue, ^(){
 
+        
         // If the session is not yet initialized
```
gdeluna-branch, line 1985: Can you add a comment on this line that reset means we've got a new link we need to process

gdeluna-branch: Or whatever `reset` specifically indicates

gdeluna-branch (approved):

nsingh-branch (approved):

PR #1433, opened 2024-09-24
[IN PROGRESS] INTENG-21262 Fix for Duplicate purchase events (cached).

NidhiDixit09:
## Reference
https://branch.atlassian.net/browse/INTENG-21262 

## Summary

Issue : SDK sends Duplicate purchase events on cold launch of app if network request processing failed in previous launches.

SDK archives a copy of its Request Queue on Disk. Its a txt file (`BNCServerRequestQueue`)stored inside the `Application Support` folder of the data container of app. This archived copy is updated at regular intervals using a timer. 
When app is cold launched, SDK reads this file and adds requests left unprocessed from last run in its Request Queue. 
Because of a bug in SDK, this archived file was not deleted from disk if current request queue is empty. As a result, SDK keeps sending these cached requests on every cold launch of app.

Steps for reproducing bug :

1. Turn off WIFI / internet
2. Send any Standard or custom Event.
3. Quit App.
4. Turn On WIFI / Internet
5. Launch App. (Cold launch is required for reproducing this issue)
6. On Launch app will send all failed events. This is expected behavior.
7. Quit App 
8. Re-Launch again
9. On Launch app will send all failed events again. These events are not cleared from cache and SDK keeps sending them.

## Type Of Change

## Testing Instructions
-> Check all unit tests pass.
-> Verify fix -
       - Create `BNCServerRequestQueue` archive file using steps mentioned above (Steps 1-4). On my system its stored at path `/Users/nidhi.dixit/Library/Developer/CoreSimulator/Devices/9C999B8A-A12B-4DDC-9A7E-A7F9DE201024/data/Containers/Data/Application/FD451857-78A8-4C5A-88C3-7CC9932D931C/Library/Application\ Support/io.branch`. I am testing using simulator.
       - Cold Lauch App and check all failed requests in previous launch are processed. `BNCServerRequestQueue`  file is deleted / updated and it does not contain processed requests anymore.
       - Cold Launch App again and check  SDK is not sending duplicate purchase events.

### DIFF
--- a/Sources/BranchSDK/BNCServerRequestQueue.m
+++ b/Sources/BranchSDK/BNCServerRequestQueue.m
@@ -217,8 +217,6 @@ static inline uint64_t BNCNanoSecondsFromTimeInterval(NSTimeInterval interval) {
 - (void)persistImmediately {
     @synchronized (self) {
         if (!self.queue || self.queue.count == 0) {
-            //No more requests. Delete cached queue file.
-            [self removeSaveFile];
             return;
         }
         NSArray *queueCopy = [self.queue copy];
@@ -298,7 +296,7 @@ static inline uint64_t BNCNanoSecondsFromTimeInterval(NSTimeInterval interval) {
 // It's been reported that unarchive can fail in some situations. In that case, remove the queued requests file.
 - (void)removeSaveFile {
     NSURL *fileURL = [BNCServerRequestQueue URLForQueueFile];
-    if (fileURL && [NSFileManager.defaultManager fileExistsAtPath:[fileURL path]]) {
+    if (fileURL) {
         NSError *error;
         [NSFileManager.defaultManager removeItemAtURL:fileURL error:&error];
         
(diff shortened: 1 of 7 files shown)


nsingh-branch (approved):
After testing with the current SDK and this branch, i was able to see the issue before but now it appears to be fixed.

Sources/BranchSDK/BNCServerRequestQueue.m
```diff
@@ -287,6 +287,8 @@ - (void)retrieve {
                 }
             }
             self.queue = decodedQueue;
+            // Requests are loaded into queue now. Delete queue file stored on disk.
+            [self removeSaveFile];
```
gdeluna-branch, line 291: Is this line always executed at the end? In other words, can an error scenario prevent this line?

gdeluna-branch, line 291: Just confirming this operation is atomic

NidhiDixit09: In current implementation, retrive function is called only once (called using function dispatch_once - singleton class). So no threading issues I see here.
This line will not be executed only if `data` is empty or any error happened while reading the cache file. That is correct behavior.
Other than that, I dont see anything stopping this line execution after decoded queue is assigned to SDK request queue `self.queue = decodedQueue;`

NidhiDixit09: Yes it is.

gdeluna-branch (approved):

PR #1441, opened 2024-10-18
SDK-2506 Removed Events Caching

NidhiDixit09:
## Reference
SDK-2506  -- Remove Events Caching
https://branch.atlassian.net/browse/SDK-2506

## Summary

This PR removes on disk caching of events as well as related test cases.

NOTE : Encoding Functions (initWithCoder, encodeWithCoder, supportsSecureCoding) are not removed from request classes. They might be used by app developers if they serialize/archive these objects.

## Testing Instructions
Prior behavior:  SDK maintained a copy of Server Request Queue on disk. On every cold launch, SDK reads request objects from this disk queue and add then tries to send them again.

After: SDK will not read and write any events to disk.

Steps to test -

1. Try to fail network requests either by turning off network or force quitting app.
2. Cold launch app again. 
3. Verify previously failed events are not sent again.


Sources/BranchSDK/BNCServerRequestQueue.m
```diff
@@ -17,20 +17,13 @@
 
 #import "BranchLogger.h"
 
-
-static NSString * const BRANCH_QUEUE_FILE = @"BNCServerRequestQueue";
-static NSTimeInterval const BATCH_WRITE_TIMEOUT = 3.0;
-
-
 static inline uint64_t BNCNanoSecondsFromTimeInterval(NSTimeInterval interval) {
```
nsingh-branch, line 20: Can this be removed as well? It seems like its not used anywhere else either

nsingh-branch (approved):
Looks good to me! Just one small nit

PR #1451, opened 2024-11-26
[SDK-2473] Fix for early branch init on install

nsingh-branch:
## Summary
When a BranchEvent is logged on first open before the SDK is initialized, we try to initialize first to log it properly. However, the SDK was reaching a state where `processNextQueueItem` was stuck and unable to process the first install or the event and nothing would happen until the next session. It seems like this was due to the BranchEvent request basically causing the queue to get stuck and the SDK to stay in an initializing state since `processNextQueueItem` would be skipped `networkCount` wasn't equal to 0. 
This change updates the if statement to also set the networkCount to 0 which allows the install to be processed and then the event to be processed after that. Since this statement only applies to non-install or open events on the first session, this change shouldn't have an effect on other cases.

## Motivation

Fix for a bug, mainly for the Adobe Branch plugin, but other clients may be trying to log events early in some cases as well.

## Type Of Change

## Testing Instructions

Try out the updated testbed which logs an event from the AppDelegate before the SDK is initialized. 

### DIFF
--- a/Branch-TestBed/Branch-TestBed/AppDelegate.m
+++ b/Branch-TestBed/Branch-TestBed/AppDelegate.m
@@ -12,6 +12,7 @@
 #import "ViewController.h"
 #import "Branch.h"
 #import "BNCEncodingUtils.h"
+#import "BranchEvent.h"
 
 AppDelegate* appDelegate = nil;
 void APPLogHookFunction(NSDate*_Nonnull timestamp, BranchLogLevel level, NSString*_Nullable message);
@@ -33,7 +34,6 @@ didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
 
     // Branch.useTestBranchKey = YES;  // Make sure to comment this line out for production apps!!!
     Branch *branch = [Branch getInstance];
-    
 
     // Change the Branch base API URL
     //[Branch setAPIUrl:@"https://api3.branch.io"];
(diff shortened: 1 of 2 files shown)


Branch-TestBed/Branch-TestBed/AppDelegate.m
```diff
@@ -81,6 +81,12 @@ - (BOOL)application:(UIApplication *)application
         [self handleDeepLinkObject:universalObject linkProperties:linkProperties error:error];
     }];
 
+    
+    BranchEvent *earlyEvent = [BranchEvent standardEvent:BNCAddToCartEvent];
+    NSLog(@"Logging Early Event: %@", earlyEvent);
+    [earlyEvent logEvent];
```
gdeluna-branch, line 87: Is this supposed to queue an event before the sdk is initialized?

nsingh-branch: Yeah

Sources/BranchSDK/Branch.m
```diff
@@ -1920,10 +1922,11 @@ - (void)processNextQueueItem {
 
         if (req) {
 
-            // If tracking is disabled, then do not check for install event.  It won't exist.
+            // If tracking is disabled, then do not check for install event. It won't exist.
             if (!Branch.trackingDisabled) {
                 if (![req isKindOfClass:[BranchInstallRequest class]] && !self.preferenceHelper.randomizedBundleToken) {
                     [[BranchLogger shared] logError:@"User session has not been initialized!" error:nil];
+                    self.networkCount = 0;
```
gdeluna-branch, line 1929: Does this need to be a thread safe operation? Can other areas be checking this value at the same time, and executing/not executing an event because of it? cc @NidhiDixit09

NidhiDixit09: @gdeluna-branch `networkCount` is defined as a property and its protected using `@synchronized (self)` inside its accessor functions. So its thread safe. 

@nsingh-branch I think we should add this fix for opens also. Its a rare case to happen but technically `else if `  condition has the same issue.

ahmednawar:
@nsingh-branch what's the next step here? cc @gdeluna-branch

gdeluna-branch:
@NidhiDixit09 will take this pr @ahmednawar

NidhiDixit09 (approved):

PR #1458, opened 2024-12-09
[SDK-2527] Add request & response objects to the BranchLogCallback

nsingh-branch:
## Reference
SDK-2427 -- Add request & response objects to the BranchLogCallback

## Summary

Updates the BranchLogCallback to include the request and response objects themselves rather than just include the data in the log message.

## Motivation

Make it easier to parse networking logs.

## Type Of Change

## Testing Instructions

Try out the new BranchLogCallback with the request and response. This is a breaking change since it adds new fields to the BranchLogCallback

### DIFF
--- a/Branch-TestBed/Branch-TestBed/AppDelegate.m
+++ b/Branch-TestBed/Branch-TestBed/AppDelegate.m
@@ -41,13 +41,21 @@ didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
     // test pre init support
     //[self testDispatchToIsolationQueue:branch]
     
-    
-    [Branch enableLoggingAtLevel:BranchLogLevelVerbose withCallback:^(NSString * _Nonnull message, BranchLogLevel logLevel, NSError * _Nullable error) {
+    [Branch enableLoggingAtLevel:BranchLogLevelVerbose withCallback:^(NSString * _Nonnull message, BranchLogLevel logLevel, NSError * _Nullable error, NSMutableURLRequest * _Nullable request, BNCServerResponse * _Nullable response) {
         // Handle the log message and error here. For example, printing to the console:
         if (error) {
             NSLog(@"[BranchLog] Level: %lu, Message: %@, Error: %@", (unsigned long)logLevel, message, error.localizedDescription);
         } else {
             NSLog(@"[BranchLog] Level: %lu, Message: %@", (unsigned long)logLevel, message);
+        }
+        
+        if (request) {
+            NSString *jsonString = [[NSString alloc] initWithData:request.HTTPBody encoding:NSUTF8StringEncoding];
+            NSLog(@"[BranchLog] Got %@ Request: %@", request.URL , jsonString);
+        }
+        
+        if (response) {
+            NSLog(@"[BranchLog] Got %@ Response: %@", response, response.data);
         }
         
         NSString *logEntry = error ? [NSString stringWithFormat:@"Level: %lu, Message: %@, Error: %@", (unsigned long)logLevel, message, error.localizedDescription]
--- a/Branch-TestBed/Branch-TestBed/AppDelegate.m
+++ b/Branch-TestBed/Branch-TestBed/AppDelegate.m
@@ -41,7 +41,7 @@ didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
     // test pre init support
     //[self testDispatchToIsolationQueue:branch]
     
-    [Branch enableLoggingAtLevel:BranchLogLevelVerbose withCallback:^(NSString * _Nonnull message, BranchLogLevel logLevel, NSError * _Nullable error, NSMutableURLRequest * _Nullable request, BNCServerResponse * _Nullable response) {
+    [Branch enableLoggingAtLevel:BranchLogLevelVerbose withAdvancedCallback:^(NSString * _Nonnull message, BranchLogLevel logLevel, NSError * _Nullable error, NSMutableURLRequest * _Nullable request, BNCServerResponse * _Nullable response) {
         // Handle the log message and error here. For example, printing to the console:
         if (error) {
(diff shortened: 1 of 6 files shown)


Branch-TestBed/Branch-TestBed/AppDelegate.m
```diff
@@ -41,15 +41,23 @@ - (BOOL)application:(UIApplication *)application
...
+    [Branch enableLoggingAtLevel:BranchLogLevelVerbose withCallback:^(NSString * _Nonnull message, BranchLogLevel logLevel, NSError * _Nullable error, NSMutableURLRequest * _Nullable request, BNCServerResponse * _Nullable response) {
         // Handle the log message and error here. For example, printing to the console:
         if (error) {
             NSLog(@"[BranchLog] Level: %lu, Message: %@, Error: %@", (unsigned long)logLevel, message, error.localizedDescription);
         } else {
             NSLog(@"[BranchLog] Level: %lu, Message: %@", (unsigned long)logLevel, message);
         }
         
+        if (request) {
+            NSString *jsonString = [[NSString alloc] initWithData:request.HTTPBody encoding:NSUTF8StringEncoding];
```
bredmond5, line 53: this gets rid of the class logging?

Branch-TestBed/Branch-TestBed/AppDelegate.m
```diff
@@ -41,15 +41,23 @@ - (BOOL)application:(UIApplication *)application
     // test pre init support
     //[self testDispatchToIsolationQueue:branch]
     
-    
-    [Branch enableLoggingAtLevel:BranchLogLevelVerbose withCallback:^(NSString * _Nonnull message, BranchLogLevel logLevel, NSError * _Nullable error) {
+    [Branch enableLoggingAtLevel:BranchLogLevelVerbose withCallback:^(NSString * _Nonnull message, BranchLogLevel logLevel, NSError * _Nullable error, NSMutableURLRequest * _Nullable request, BNCServerResponse * _Nullable response) {
```
bredmond5, line 44: so this allows usage of the old function without request and response right?

bredmond5 (approved):
couple questions but seems good

Branch-TestBed/Branch-TestBed/AppDelegate.m
nsingh-branch: The class details are still included in the log message. This part of the code is just part testbed app using the new request object.

Branch-TestBed/Branch-TestBed/AppDelegate.m
nsingh-branch: Since they were added to the callback that the function uses itself, there is a slight code change when calling `enableLoggingAtLevel withCallback`

bredmond5: ok, does that make this a major change?

nsingh-branch: Since it's just a small callback syntax change, it can still be a minor version bump

gdeluna-branch: Does this force developers to change their function call and recompile?

nsingh-branch: @gdeluna-branch Yes, they would have to add `NSMutableURLRequest * _Nullable request, BNCServerResponse * _Nullable response)` to the end of the callback if they're using it, but it wouldn't break or require any changes in their logic.

gdeluna-branch: Is there not a way to instead keep the original function signature and have null/empty defaults? Otherwise this is a breaking change.

nsingh-branch: @NidhiDixit09 The change is in BranchLogCallback itself though as opposed to changing enableLogging, so overloading wouldn't work unless we also created another BranchLogCallback.

Overall, this may be a breaking change but I think it's still fine considering the fact that it's a more niche, advanced method/callback and the fix is simple enough to be handled by autocomplete.

NidhiDixit09: Yes, we need to define `BranchLogCallbackWithRequestObjects` also in order to overload `enableLogging` method.

gdeluna-branch: Yeah I'd rather we extend this until we really want to break it. We/our field teams will get questions if something breaks otherwise.

nsingh-branch: Understandable, but I'm not sure if there's an elegant way to extend it without creating a new function or callback, which may end up being more confusing.
So my vote is just breaking it, but we can try creating a new callback instead if everyone thinks that's better in this case.

Sources/BranchSDK/BranchLogger.m
```diff
@@ -77,8 +93,10 @@ - (void)logMessage:(NSString *)message withLevel:(BranchLogLevel)level error:(NS
     if (self.includeCallerDetails) {
         formattedMessage = [NSString stringWithFormat:@"%@ %@", [self callingClass], message];
     }
-
-    if (self.logCallback) {
+    
+    if  (self.advancedLogCallback) {
```
NidhiDixit09, line 97: @nsingh-branch  Since `advancedLogCallback` is always non-null, its initialized in `init`, `else if ` condition will never be executed.

nsingh-branch: I removed the default init since it shouldn't be necessary to have both. All of the tests appear to be passing now.

gdeluna-branch: Did the new automation succeed as well?

gdeluna-branch (approved):

NidhiDixit09 (approved):

PR #1469, opened 2025-01-09
[INTENG-21518] Fix for compilation issue on tvOS

nsingh-branch:
## Reference
INTENG-21518

## Summary
Removed a check for tvOS when including certain files in the umbrella header. Since the files themselves contain the tvOS checks, this one appears to be unnecessary and caused Xcode to be unable to locate these files when building an app with the SDK.

## Motivation
Fix for clients using tvOS apps.

## Type Of Change

## Testing Instructions
Compile and run the app to make sure it still builds on different integrations as well.

### DIFF
--- a/Framework/BranchSDK.h
+++ b/Framework/BranchSDK.h
@@ -30,14 +30,10 @@ FOUNDATION_EXPORT const unsigned char BranchSDKVersionString[];
 
 #import <BranchSDK/BranchDeepLinkingController.h>
 
-#if !TARGET_OS_TV
-// tvOS does not support these features
 #import <BranchSDK/BranchShareLink.h>
 #import <BranchSDK/BranchCSSearchableItemAttributeSet.h>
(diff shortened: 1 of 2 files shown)


Framework/BranchSDK.h
```diff
@@ -30,14 +30,10 @@ FOUNDATION_EXPORT const unsigned char BranchSDKVersionString[];
 
 #import <BranchSDK/BranchDeepLinkingController.h>
 
-#if !TARGET_OS_TV
-// tvOS does not support these features
```
gdeluna-branch, line 34: So we'll just have to manually place a check in each class/function where we don't support tvOS?

NidhiDixit09: Yes. Apps trying to use it will get errors similar to - API not available in this scope.

NidhiDixit09 (approved):

PR #1474, opened 2025-01-30
Removed 'nullable' attribute of BranchAdvancedLogCallback param

NidhiDixit09:
## Reference
SDK-2537 -- Fix enableLogging API error - Ambiguous use of 'enableLogging'
https://branch.atlassian.net/browse/SDK-2537

## Summary

Following are two overloaded versions of static API `enableLoggingAtLevel` and both had the second param `nullable` attribute. This resulted in two methods with same name and same param.

```
+ (void)enableLoggingAtLevel:(BranchLogLevel)logLevel withCallback:(nullable BranchLogCallback)callback;
+ (void)enableLoggingAtLevel:(BranchLogLevel)logLevel withAdvancedCallback:(nullable  BranchAdvancedLogCallback)callback;
```

## Type Of Change

## Testing Instructions
In a test app, call API ` Branch.enableLogging(at: .verbose)`. Compile app to see we no longer get Build error.

### DIFF
--- a/Sources/BranchSDK/Public/Branch.h
+++ b/Sources/BranchSDK/Public/Branch.h
@@ -568,7 +568,7 @@ extern NSString * __nonnull const BNCSpotlightFeature;
  */
 + (void)enableLogging;
 + (void)enableLoggingAtLevel:(BranchLogLevel)logLevel withCallback:(nullable BranchLogCallback)callback;
-+ (void)enableLoggingAtLevel:(BranchLogLevel)logLevel withAdvancedCallback:(nullable BranchAdvancedLogCallback)callback;
++ (void)enableLoggingAtLevel:(BranchLogLevel)logLevel withAdvancedCallback:(BranchAdvancedLogCallback)callback;
 
 // The new logging system is independent of the Branch singleton and can be called earlier.
 - (void)enableLogging __attribute__((deprecated(("This API is deprecated. Please use the static version."))));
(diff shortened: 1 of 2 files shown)


Sources/BranchSDK/Public/Branch.h
```diff
@@ -568,7 +568,7 @@ extern NSString * __nonnull const BNCSpotlightFeature;
  */
 + (void)enableLogging;
 + (void)enableLoggingAtLevel:(BranchLogLevel)logLevel withCallback:(nullable BranchLogCallback)callback;
-+ (void)enableLoggingAtLevel:(BranchLogLevel)logLevel withAdvancedCallback:(nullable BranchAdvancedLogCallback)callback;
++ (void)enableLoggingAtLevel:(BranchLogLevel)logLevel withAdvancedCallback:(BranchAdvancedLogCallback)callback;
```
gdeluna-branch, line 571: Since this is no longer nullable, do we expect any existing integrations to break as a result? What is the impact?

NidhiDixit09: @gdeluna-branch  ` + (void)enableLoggingAtLevel:(BranchLogLevel)logLevel withCallback:(nullable BranchLogCallback)callback; ` has `nullable` attribute. So this will take care of any code calling `enableLoggingAtLevel` without callback.
Both these API just differ from the callback param - one is older one and the other one is `BranchAdvancedLogCallback` which was introduced in last Release. So just changed the attiribute type of this new API. Nothing is changed for the older one.

gdeluna-branch (approved):

PR #1479, opened 2025-02-19
INTENG-21518 - SPM + Hybrid app compilation issue.

NidhiDixit09:
## Reference
INTENG-21518 : tvOS v3.5.1 and v3.6.4 are throwing compilation errors addressed in previous INTENG due to clients updated implementation
https://branch.atlassian.net/browse/INTENG-21518

## Summary

Branch SDK umbrella header file is common for all the platforms. Some files are for iOS platform only. Branch SDK handles platform specific code using preprocessor checks.  Module map and SPM manifest file has also configuration common for all the platforms. This works fine if app is iOS only or tvOS only. But it fails for a hybrid app.

This PR includes following workarounds :

SPM :  Removed a check for tvOS when including certain files in the umbrella header. Since the files themselves contain the tvOS checks, this one appears to be unnecessary and caused Xcode to be unable to locate these files when building an app with the SDK. 
Cocoa Pods: Removed explicit exclusion of files included in cocoa pod spec file.
Framework Binaries : Moved BranchPasteControl.h to Public folder for tvOS platform.

## Motivation
Fix for clients using tvOS apps.

## Type Of Change

## Testing Instructions
Compile and run the app to make sure it still builds on different integrations.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Run unit tests
         run: bundle exec fastlane unit_tests
       - name: Upload test results
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         if: always() # even if tests fail
         with:
           name: test-results
--- a/BranchSDK.podspec
+++ b/BranchSDK.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = "BranchSDK"
-  s.version          = "3.6.5"
+  s.version          = "3.9.0"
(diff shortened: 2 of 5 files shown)


gdeluna-branch:
Can you describe if there's any available workaround for cocoapods going forward or if this is a known issue?

NidhiDixit09:
@gdeluna-branch Its a known issue. Cocoapod does not provide this feature. It does not support multi platform frameworks.

gdeluna-branch (approved):

NidhiDixit09:
Added cocoa pod issue link for ref- https://github.com/CocoaPods/CocoaPods/issues/9566

PR #1570, opened 2026-02-16
Refactor SKAN related code and BranchOpenRequest Class

NidhiDixit09:
## Reference
[iOS] Refactor SKAN related code and BranchOpenRequest Class
https://branch.atlassian.net/browse/EMT-3068

## Summary

This PR contains no behavioral changes. It has purely structural refactoring to remove duplicate codes and simplify functions for easier testing and debugging.
ReFactored BranchOpenRequest and BranchEvent  class to move SKAN related code to BNCSKAdNetwork class. 
Moved code out of processResponse function into more focused helper functions.


Sources/BranchSDK/BranchOpenRequest.m
```diff
@@ -266,6 +157,84 @@ - (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
...
+    NSString *token = BNCStringFromWireFormat(data[BRANCH_RESPONSE_KEY_RANDOMIZED_BUNDLE_TOKEN]);
+    if (!token) {
+        // fallback to deprecated name. The old name was easily confused with the setIdentity, hence the name change.
+        token = BNCStringFromWireFormat(data[@"identity_id"]);
+    }
+    return token;
+}
+
+- (NSString *)randomizedDeviceTokenFromResponseData:(NSDictionary *)data {
+    NSString *token = data[BRANCH_RESPONSE_KEY_RANDOMIZED_DEVICE_TOKEN];
```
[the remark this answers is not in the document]
NidhiDixit09: Its internal function and Null is handled by other functions using it. Also didnt want to change existing logic. So keeping it as it is.

Sources/BranchSDK/BranchOpenRequest.m
```diff
@@ -266,6 +157,84 @@ - (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
...
+    } else
+    if ([sessionData isKindOfClass:[NSDictionary class]]) {
+        [[BranchLogger shared] logWarning:[NSString stringWithFormat:@"Received session data of type '%@' data is '%@'.", NSStringFromClass(sessionData.class), sessionData] error:nil];
+        sessionData = [BNCEncodingUtils encodeDictionaryToJsonString:(NSDictionary*)sessionData];
+    } else
+    if ([sessionData isKindOfClass:[NSArray class]]) {
+        [[BranchLogger shared] logWarning:[NSString stringWithFormat:@"Received session data of type '%@' data is '%@'.", NSStringFromClass(sessionData.class), sessionData] error:nil];
+        sessionData = [BNCEncodingUtils encodeArrayToJsonString:(NSArray*)sessionData];
+    } else {
+        [[BranchLogger shared] logError:[NSString stringWithFormat:@"Received session data of type '%@' data is '%@'.", NSStringFromClass(sessionData.class), sessionData] error:nil];
```
wpinho-branch: The original `logError` call here passed the error parameter from `processResponse:error:`, which provided HTTP/network error context alongside the type mismatch. The extracted helper lost that context and now passes `error:nil`.

Consider either passing the error as a parameter to this helper, or accepting this minor loss if this log path is rarely hit.

Sources/BranchSDK/BranchOpenRequest.m
```diff
@@ -159,100 +135,15 @@ - (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
...
-            } else {
-                [[BNCSKAdNetwork sharedInstance] updateConversionValue:conversionValue.integerValue];
-            }
-        }
-    }
-#endif
+
+    [self handleSKANWithResponseData:data];
+
+    [BranchOpenRequest releaseOpenResponseLock];
```
wpinho-branch, line 144: nit: `releaseOpenResponseLock` ordering change

In the original code, `releaseOpenResponseLock` was called **before** SKAN handling. With this refactoring it moved **after** `handleSKANWithResponseData:` (line 144), which means the open response lock is now held during SKAN operations.

Any request blocked on `waitForOpenResponseLock` will wait slightly longer. Since SKAN operations don't depend on the lock, would it make sense to keep the original ordering?

```objc
// Original order preserved:
[BranchOpenRequest releaseOpenResponseLock];

[self handleSKANWithResponseData:data];
```

Not a blocker, just flagging since the PR description says "no behavioral changes."

Sources/BranchSDK/BranchOpenRequest.m
```diff
@@ -266,6 +157,84 @@ - (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
...
+- (NSString *)randomizedBundleTokenFromResponseData:(NSDictionary *)data {
+    NSString *token = BNCStringFromWireFormat(data[BRANCH_RESPONSE_KEY_RANDOMIZED_BUNDLE_TOKEN]);
+    if (!token) {
+        // fallback to deprecated name. The old name was easily confused with the setIdentity, hence the name change.
+        token = BNCStringFromWireFormat(data[@"identity_id"]);
+    }
+    return token;
+}
+
+- (NSString *)randomizedDeviceTokenFromResponseData:(NSDictionary *)data {
```
wpinho-branch, line 169: The new version always attempts the `device_fingerprint_id` fallback when the primary value is `nil`, including when the key is entirely absent from the dictionary. This could be a net improvement (supporting very old server responses), but it's a change worth being aware of.

Also, `randomizedBundleTokenFromResponseData:` (line 160) uses `BNCStringFromWireFormat` for safe `NSNull` handling, but this method accesses the dictionary directly.

Sources/BranchSDK/BranchOpenRequest.m
```diff
@@ -266,6 +157,84 @@ - (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
...
+    NSString *token = data[BRANCH_RESPONSE_KEY_RANDOMIZED_DEVICE_TOKEN];
+    if (!token) {
+        // fallback to deprecated name. Fingerprinting was removed long ago, hence the name change.
+        token = data[@"device_fingerprint_id"];
+    }
+    return token;
+}
+
+- (NSString *)userIdentityFromResponseData:(NSDictionary *)data {
```
wpinho-branch, line 181: Suggestion: `BNCStringFromWireFormat` handles `NSNull` differently than the original

The original code only converted `NSNumber` to string and passed everything else through. `BNCStringFromWireFormat` calls `description` on any non-string object — for `NSNull`, that returns the literal string `"<null>"` instead of `nil`.

If the server ever sends `null` for developer identity, the old code stored `NSNull` (generally treated as nil-like downstream), while the new code would store `"<null>"` as an actual string identity.

If `null` identity from the server is not a realistic scenario, this is fine. Otherwise, an explicit nil-guard could help:

```objc
if (!userIdentity || [userIdentity isKindOfClass:[NSNull class]]) {
    return nil;
}
return BNCStringFromWireFormat(userIdentity);
```

wpinho-branch (approved):
Overall the refactoring direction looks great — consolidating SKAN logic into BNCSKAdNetwork and extracting helpers from processResponse makes the code much more readable and testable. Nice work!

A few minor things I noticed:

1. releaseOpenResponseLock ordering (BranchOpenRequest.m:146) — moved after SKAN handling, which extends the lock scope slightly. See inline comment.
2. randomizedDeviceTokenFromResponseData: fallback (BranchOpenRequest.m:169) — subtle change in when the deprecated key fallback is attempted. See inline comment.
3. Commit hygiene — there are duplicate commit messages and a "Test commit" (b4ef55e3). Consider squashing before merge for a cleaner history.

 None of these are blockers — the core refactoring is solid.

gdeluna-branch (approved):

Sources/BranchSDK/BranchOpenRequest.m
```diff
@@ -266,6 +157,87 @@ - (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
...
+        // fallback to deprecated name. The old name was easily confused with the setIdentity, hence the name change.
+        token = BNCStringFromWireFormat(data[@"identity_id"]);
+    }
+    return token;
+}
+
+- (NSString *)randomizedDeviceTokenFromResponseData:(NSDictionary *)data {
+    if (![data objectForKey:BRANCH_RESPONSE_KEY_RANDOMIZED_DEVICE_TOKEN]) {
+        return nil;
+    }
```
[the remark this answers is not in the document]
NidhiDixit09: Previous implementation was like this only. So keeping that logic.

Sources/BranchSDK/BranchOpenRequest.m
NidhiDixit09: Fixed.

Sources/BranchSDK/BranchOpenRequest.m
NidhiDixit09: Keeping it as it is to keep "No behavioral changes".

Sources/BranchSDK/BranchOpenRequest.m
NidhiDixit09: fixed.

Sources/BranchSDK/BranchOpenRequest.m
NidhiDixit09: Fixed. Reverted to original behavior.

PR #1587, opened 2026-04-16
test(gptdriver-ios): add hybrid MobileBoost test suite (parity with Android)

wpinho-branch:
## Summary

Ships the iOS hybrid test suite (`Branch-TestBed/TestBed-GPTDriverTests/`) — parity with the Android `Branch-SDK-GPTDriver` module documented in the [EMT-MobileBoost-iOS-Test-Plan](https://www.notion.so/33d793ae707780a8aeb4d0eae919bd4a).

- **16 hybrid test classes** (Deterministic + Hybrid + AI-only layers) covering link creation, QR code, share, session/logs, user identity, event logging, deep link cold/warm open, browser handoff, tracking, consumer protection, referring params, notifications, plugin notify
- **`BaseGptDriverTest.swift`** bootstraps the `gptd-swift` MobileBoost SDK and resolves `MOBILEBOOST_API_KEY` from `Info.plist` (xcconfig) or env var
- **`TestBedIdentifiers.h/.m`** + accessibility identifiers on all interactive controls in `ViewController.m`, `LogOutputViewController.m`, `PasteControlViewController.m` — single source of truth for UI automation selectors
- **Deep-link test hook** in `AppDelegate.m` (`#if DEBUG`) synthesizes `NSUserActivity` from a `-testDeepLinkURL` launch arg so universal-link flows run on an unsigned simulator build
- **Test Plans:** `Smoke.xctestplan` (~37s, 5 tests) and `Release.xctestplan` (full suite)
- **Docs:** `README.md` (setup + secrets) and `TESTING_GUIDE.md` (philosophy + decision matrix + pre-merge checklist)
- **Results formatter:** `scripts/format-test-results.sh` emits a markdown row from an xcodebuild log
- **Secret hygiene:** `.gitignore` now excludes `mobileboost.config.json` and `*.local.xcconfig`; `MobileBoost.local.xcconfig.example` is the committed template

## Test Plan

Latest local run (2026-04-15):

| Platform | Destination | Pass | Fail | Skip | Wall time |
|----------|-------------|------|------|------|-----------|
| iOS | iPhone 16 Simulator, latest iOS | 37 | 0 | 2 | ~15m 24s |

The 2 skips are `XCTSkip` placeholders (`NotificationHybridTest`, `PluginNotifyHybridTest`) waiting on TestBed host-app buttons — tracked in the test plan doc. They activate automatically once the buttons land.

**Reproduce locally:**

```bash
cd Branch-TestBed/TestBed-GPTDriverTests/Config
[...]

PR #1590, opened 2026-05-12
feat: Implement Layer 1 SDK Wire-Level Validation and MobileBoost Release Workflow

wpinho-branch:
This PR establishes parity with Android by implementing L1 wire-level validation and automating the MobileBoost release gate for iOS.

**Key Changes:**
- **L1 Validation Script:** Added `scripts/validate_l1_logs.py` tailored for Xcode console output parsing and JSON payload assertion.
- **PR Workflow:** Created `.github/workflows/layer1-logger-tests.yml` (using `macos-latest`) to gate PRs based on wire-level correctness.
- **Release Automation:** Added `.github/workflows/gptdriver-release.yml` to trigger full UI automation on MobileBoost Cloud for `Release-*` branches.
- **Parity:** Implemented critical event detection logic to prevent false positives in CI.

---

## Update — Field expansion & full body dump (2026-05-22)

Resolves reviewer feedback on the validator output: [full body dump] and [extended required-field set].

### What changed

- `scripts/validate_l1_logs.py` now validates **presence** of 18 fields on `/v1/*` endpoints and prints the **full payload** for every captured request — with `branch_key`, `hardware_id`, `idfa`, `idfv`, `anon_id`, randomized tokens, `apple_attribution_token`, `identity`, `developer_identity`, and ad IDs masked — alongside a per-field ✓ table showing the actual value that went over the wire. No more silent passes on wrong values.
- New `scripts/test_validate_l1_logs.py` — 24 unit tests including iOS-specific assertions (no `wifi`/`ui_mode` in required set; `connection_type` is required; `sdk` vs `sdk_version` split on v2). Wired into the workflow ahead of the real validation step.
- New `scripts/fixtures/` — 5 hand-tailored `branchlogs.txt` snippets used by the unit tests.
- Workflow update: artifact sanitization now also strips `device_fingerprint_id`, `identity`, `developer_identity`, and `apple_attribution_token`.
- New `scripts/README.md` documenting local-run, the contract, and how to add a new required field.

### Validator behaviour

- Required-field checks scoped to `/v1/*`. Non-v1 endpoints (e.g. `/v2/event/*`) get their payload printed but do not fail the run — that is the L1 contract. iOS in particular nests device fields under `user_data` on `/v2/event/*` and uses different identity fields there; enforcing the v1 contract on v2 would surface false positives.
- iOS does NOT emit `wifi` or `ui_mode` on the wire — those are Android-only by design. Cross-platform alignment of these device-context fields is tracked under the Event API Modernization workstream, not this gate.

### Commits

| SHA | What it does |
| --- | --- |
| `e4b27464` | Rewrite validator: 18-field presence checks + full body dump + masking |
| `da6e6c56` | Unit tests + fixtures |
| `257bb877` | Workflow: unit-test step + extended sanitization |
| `bb476421` | scripts/README.md |
| `c4ae17b0` | Scope required-field checks to `/v1/*` only |

### CI evidence (end-to-end verified)

| Run | SHA | Result | What it proves |
| --- | --- | --- | --- |
| [26315817587](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/actions/runs/26315817587) | `c4ae17b0` | ✅ green | Baseline: validator green on real captured `branchlogs.txt` |
| [26316106114](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/actions/runs/26316106114) | `5c87d4eb` | ❌ red | Synthetic break (intentional) — proves the gate fails on a missing field |
| [26316281485](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/actions/runs/26316281485) | `22b4da09` | ✅ green | Revert — gate green on clean state |


.github/workflows/layer1-logger-tests.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          xcode-version: '15.4'
+
+      - name: Run Unit Tests & Capture Logs
+        run: |
+          xcodebuild test \
+            -project BranchSDK.xcodeproj \
+            -scheme BranchSDKTests \
+            -destination "platform=iOS Simulator,name=iPhone 15,OS=latest" \
+            | tee ios_test_output.log
```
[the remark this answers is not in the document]
wpinho-branch: Auto-resolved by the workflow rewrite in commit `1121078a`. The current `layer1-logger-tests.yml` no longer pipes `xcodebuild test` through `tee`; instead it invokes `scripts/run_l1_instrumented.sh`, which uses `set -euo pipefail` from the top and captures the xcodebuild exit code explicitly via `|| TEST_EXIT_CODE=$?`. Closing this thread.

scripts/validate_l1_logs.py
```diff
@@ -0,0 +1,80 @@
...
+        # Matches the first balanced { } or until a new log tag starts
+        json_match = re.search(r'(\{.*?\})', block, re.DOTALL)
+        if json_match:
+            try:
+                # Clean up potential log prefixes in each line
+                json_str = json_match.group(1)
+                entry['request'] = json.loads(json_str)
+                entries.append(entry)
+            except json.JSONDecodeError:
+                continue
```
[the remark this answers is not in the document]
wpinho-branch: Auto-resolved by the validator rewrite in commit `5164da64`. The current `validate_l1_logs.py` parses one NSLog line at a time using `REQUEST_LINE_RE = r"\[BranchLog\]\s+Got\s+(?P<url>https?://[^\s]+)\s+Request:\s*(?P<body>\{.*\})\s*$"` — a greedy `\{.*\}` capture to end-of-line. Because the iOS TestBed AppDelegate emits one NSLog call per request (single logical line), greedy capture is correct here and handles nested objects (e.g. `user_data`, `metadata`). No `.*?` non-greedy fragment exists. Closing this thread.

wpinho-branch:
## MatterAI review triage

Reviewed all suggestions from the two summary reviews (12 May + 15 May) and the two line-level threads. Disposition below.

### Applied (1)

| Suggestion | File | Action |
|---|---|---|
| Pass `MOBILEBOOST_API_KEY` via `env:` (not argv) | `.github/workflows/gptdriver-release.yml` | ✅ Applied in `18b17f56` |

Rationale: argv-passed secrets are visible via `ps aux` / kernel audit on the runner even though GitHub masks them in logs. Mirrors the Android workflow's `ORG_GRADLE_PROJECT_MOBILEBOOST_API_KEY` pattern.

### Already implemented (3) — verified, no action needed

| Suggestion | Where | Why no action |
|---|---|---|
| `validate_l1_logs.py` should use `encoding="utf-8"` | `validate_l1_logs.py:69` | Already applied (with `errors="replace"` for extra robustness) in commit `5164da64`. |
| Workflow `xcodebuild | tee` needs `pipefail` (thread on `layer1-logger-tests.yml:28`) | n/a — outdated | Workflow rewrite in `1121078a` replaced the `tee` pattern with `scripts/run_l1_instrumented.sh`, which runs under `set -euo pipefail`. Thread resolved. |
| Regex `{.*?}` mis-parses nested JSON (thread on `validate_l1_logs.py:39`) | n/a — outdated | Current parser uses greedy `\{.*\}` to end-of-line in `REQUEST_LINE_RE`. AppDelegate emits one NSLog call per request, so greedy capture is correct and handles nested objects (`user_data`, `metadata`). Thread resolved. |

### Rejected (2) — declined with reasoning

| Suggestion | File | Why declined |
|---|---|---|
| Replace inline `python3` simctl parsing with `plutil + awk` | `scripts/run_l1_instrumented.sh:53-69` | Python 3 is a hard guarantee on GitHub `macos-15` runners (pre-installed + `actions/setup-python@v6` runs in the next step). More importantly, the suggested `plutil -extract … raw -o -` returns JSON, not plist-text, but the suggested `awk` regex (`/"name" = "…"/`) is written in plist-text syntax and would not match the JSON output. Keeping current. |
| Add `sleep 2` after `simctl bootstatus -b` | `scripts/run_l1_instrumented.sh:72-74` | `xcrun simctl bootstatus -b` already blocks until the device finishes booting (that is the explicit semantic of the `-b` flag). Adding `sleep` is cargo-cult — the race the suggestion describes does not exist. Keeping current. |

### Deferred (2) — defensible improvements, low ROI right now

| Suggestion | File | Why deferred |
|---|---|---|
| EXIT trap to always pull `branchlogs.txt` on early abort | `scripts/run_l1_instrumented.sh:72-74` | Useful hygiene for the corner case where `bootstatus` aborts unexpectedly — but in that scenario the app never ran, so there are no useful logs to pull. ~15 lines for low value. Will revisit if CI exhibits this flake. |
| Validate `DERIVED_DATA_DIR` exists before `test-without-building` | `scripts/run_l1_instrumented.sh:34-41` | `xcodebuild` already fails with a clear error referencing the missing `.xctestrun` path. The preceding `build-for-testing` workflow step also fails the run before we reach this script. Marginal gain. |

---

Net result: 1 commit (`18b17f56`) applied, 3 dismissed as already done, 2 rejected with rationale, 2 deferred. The L1 PR-gate workflow itself is unchanged (path filter excludes `gptdriver-release.yml`) and remained ✅ green from the previous run.

scripts/validate_l1_logs.py
```diff
@@ -0,0 +1,266 @@
...
+    if not isinstance(request, dict):
+        errors.append(f"Request {idx} ({uri}): payload is not a JSON object")
+        return errors
+
+    masked = mask_payload(request)
+    print("Full payload (sensitive values masked):")
+    print(json.dumps(masked, indent=2, sort_keys=True))
+    print()
+
+    if not uri.startswith("/v1/"):
```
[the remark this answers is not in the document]
wpinho-branch: This is theoretical, not reachable in practice. `uri` is assigned at `parse_branch_logs` from `urlparse(url).path or url`, where `url` itself comes from a regex match group that requires `https?://[^\s]+` (cannot be `None` or empty). The exception fallback also assigns `url` directly. So `uri` is always a non-empty string by the time it reaches `validate_request`.

The defensive `uri or ""` guard would be effectively free but the bug it prevents is not reachable on any code path that produces an entry. Leaving as-is.

scripts/validate_l1_logs.py
```diff
@@ -0,0 +1,269 @@
...
+    r"\[BranchLog\]\s+Got\s+(?P<url>https?://[^\s]+)\s+Request:\s*(?P<body>\{.*\})\s*$"
+)
+
+# Required on every captured /v1/* request. `wifi` and `ui_mode` are
+# intentionally absent — iOS does not emit them. See the v4 Conversion API
+# parity tracker for the future-alignment plan.
+REQUIRED_COMMON = [
+    # TEMPORARY: synthetic break to demonstrate the gate catches missing
+    # required fields against a real CI capture. Removed in the next commit.
+    "INTENTIONAL_CI_RED_TEST_remove_after_verification",
```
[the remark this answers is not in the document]
wpinho-branch: Intentional synthetic break used to verify end-to-end that the L1 gate catches missing required fields against a real captured branchlogs.txt (not only against hand-crafted fixtures).

Reverted in commit `22b4da09`. CI evidence:
- Red on the synthetic break: run [26316106114](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/actions/runs/26316106114)
- Green after revert: run [26316281485](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/actions/runs/26316281485)

.github/workflows/layer1-logger-tests.yml
```diff
@@ -0,0 +1,147 @@
...
+      # Validate first (raw log) so we can assert real value prefixes like
+      # branch_key starting with "key_test_"/"key_live_". The validator
+      # itself never echoes payload contents to stdout, so the only place
+      # raw values could surface is the artifact upload below — which is
+      # gated to the sanitized copy.
+      - name: Validate wire logs
+        id: validate
+        run: python3 scripts/validate_l1_logs.py branchlogs.txt
+
+      # Always sanitize before upload, even on success. We replace values
```
gdeluna-branch: This block is not necessary.

.github/workflows/layer1-logger-tests.yml
```diff
@@ -0,0 +1,147 @@
...
+        if: always()
+        run: |
+          if [ -f branchlogs.txt ]; then
+            sed -E -i.bak \
+              -e 's/"branch_key"[[:space:]]*:[[:space:]]*"[^"]*"/"branch_key": "***"/g' \
+              -e 's/"hardware_id"[[:space:]]*:[[:space:]]*"[^"]*"/"hardware_id": "***"/g' \
+              -e 's/"idfa"[[:space:]]*:[[:space:]]*"[^"]*"/"idfa": "***"/g' \
+              -e 's/"idfv"[[:space:]]*:[[:space:]]*"[^"]*"/"idfv": "***"/g' \
+              -e 's/"anon_id"[[:space:]]*:[[:space:]]*"[^"]*"/"anon_id": "***"/g' \
+              -e 's/"google_advertising_id"[[:space:]]*:[[:space:]]*"[^"]*"/"google_advertising_id": "***"/g' \
```
gdeluna-branch: This field does not appear in iOS

Branch-TestBed/TestBed-GPTDriverTests/Hybrid/NotificationHybridTest.swift
```diff
@@ -15,29 +15,55 @@ final class NotificationHybridTest: BaseGptDriverTest {
...
+        var tapped = false
+        if banner.waitForExistence(timeout: 15) {
+            // Coordinate-based center tap — more robust than .tap() because
+            // notification banners sometimes have non-tappable edges.
+            banner.coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0.5)).tap()
+            tapped = true
+        }
+
+        if !tapped {
+            // 2. AI FALLBACK:
```
gdeluna-branch: Was this change meant to be on the other PR?

scripts/test_validate_l1_logs.py
```diff
@@ -0,0 +1,194 @@
...
+        errors, output = _run_validation("v2_event_out_of_scope.txt")
+        self.assertEqual(
+            errors, [],
+            f"Mixed v1+v2 capture should not produce errors; got: {errors}",
+        )
+        self.assertIn("Non-v1 endpoint", output)
+        self.assertIn("required-field checks skipped per L1 scope", output)
+
+
+class MaskingHelperTests(unittest.TestCase):
```
gdeluna-branch: This can be removed.

scripts/validate_l1_logs.py
```diff
@@ -0,0 +1,266 @@
...
+}
+
+# Fields whose values must never appear unredacted in CI logs.
+SENSITIVE_FIELDS = {
+    "branch_key",
+    "hardware_id",
+    "randomized_device_token",
+    "randomized_bundle_token",
+    "device_fingerprint_id",
+    "google_advertising_id",
```
gdeluna-branch: This is not present in iOS

gdeluna-branch (approved):

wpinho-branch:
## Review triage — gdeluna-branch comments

All 5 review comments addressed in 4 atomic commits pushed in this update:

| Thread | File:line | Action | Commit |
|---|---|---|---|
| "This is not present in iOS" | `scripts/validate_l1_logs.py:76` | Removed `google_advertising_id` from `SENSITIVE_FIELDS` (Android-only field). | `10211f47` |
| "This can be removed." | `scripts/test_validate_l1_logs.py:141` | Dropped `MaskingHelperTests` class — already covered end-to-end by `HappyPathTests.test_masks_sensitive_values`. | `16d96934` |
| "This block is not necessary." | `.github/workflows/layer1-logger-tests.yml:107` | Removed the pre-upload `sed`-based sanitization step. Validator already masks every sensitive value before printing, and the artifact upload is gated to `failure()` + same-repo PRs. | `496945ec` |
| "This field does not appear in iOS" | `.github/workflows/layer1-logger-tests.yml:121` | Implicitly handled by removing the whole sanitize step (covered the `google_advertising_id` sed pattern). | `496945ec` |
| "Was this change meant to be on the other PR?" | `Branch-TestBed/TestBed-GPTDriverTests/Hybrid/NotificationHybridTest.swift:40` | Reverted commit `8ef0f4b7` (iOS 18.4 banner-query fix). It belongs on the hybrid-tests PR #1587; will be re-applied there. | `051ca7c3` |

Verified locally:
- Validator unit tests: 20/20 PASS
- Validator against real `branchlogs.txt`: PASS (2/2 requests valid)
- Workflow YAML: syntactically valid

PR #1592, opened 2026-05-22
Removed trackingDisabled code and replaced with BranchAttributionLeve…

bboothe-branch:
## Reference
EMT-3679 -- iOS - Remove trackingDisabled from 4.0.0-beta.0.

## Summary
trackingDisabled() is deprecated and will be removed from public references in favor of the setConsumerProtectionAttributionLevel() functions. 

All trackingDisabled() calls will be replaced with the new method.

## Type Of Change

## Testing Instructions

- Run app and set consumer protection attribution level to none either using AppDelegate or button on home page.
- Check to confirm logs.

### DIFF
--- a/Sources/BranchSDK/BNCPreferenceHelper.m
+++ b/Sources/BranchSDK/BNCPreferenceHelper.m
@@ -892,7 +892,7 @@ NSURL* /* _Nonnull */ BNCURLForBranchDirectory_Unthreaded(void);
 }
 
 - (void)setAttributionLevel:(BranchAttributionLevel)level {
-    if (level == BranchAttributionLevelNone) {
+    if ([level isEqualToString:BranchAttributionLevelNone]) {
         [self clearTrackingInformation];
     }
     [self writeObjectToDefaults:BRANCH_PREFS_KEY_ATTRIBUTION_LEVEL value:level];
--- a/Sources/BranchSDK/BNCPreferenceHelper.m
+++ b/Sources/BranchSDK/BNCPreferenceHelper.m
@@ -892,11 +892,14 @@ NSURL* /* _Nonnull */ BNCURLForBranchDirectory_Unthreaded(void);
 }
 
 - (void)setAttributionLevel:(BranchAttributionLevel)level {
+    if (level == BranchAttributionLevelNone) {
+        [self clearTrackingInformation];
+    }
     [self writeObjectToDefaults:BRANCH_PREFS_KEY_ATTRIBUTION_LEVEL value:level];
 }
 
 - (BOOL)attributionLevelNone {
-    return [_attributionLevel isEqualToString:@"NONE"];
+    return [[self attributionLevel] isEqualToString:BranchAttributionLevelNone];
 }
 
 - (NSString *) uxType {
--- a/Sources/BranchSDK/BNCPreferenceHelper.m
+++ b/Sources/BranchSDK/BNCPreferenceHelper.m
@@ -895,19 +895,7 @@ NSURL* /* _Nonnull */ BNCURLForBranchDirectory_Unthreaded(void);
     [self writeObjectToDefaults:BRANCH_PREFS_KEY_ATTRIBUTION_LEVEL value:level];
 }
 
-- (BOOL)attributionLevelFull:(BranchAttributionLevel)level {
-    return [_attributionLevel isEqualToString:@"FULL"];
-}
-
-- (BOOL)attributionLevelReduced:(BranchAttributionLevel)level {
-    return [_attributionLevel isEqualToString:@"REDUCED"];
-}
-
-- (BOOL)attributionLevelMinimal:(BranchAttributionLevel)level {
-    return [_attributionLevel isEqualToString:@"MINIMAL"];
-}
-
-- (BOOL)attributionLevelNone:(BranchAttributionLevel)level {
+- (BOOL)attributionLevelNone {
     return [_attributionLevel isEqualToString:@"NONE"];
 }
 
--- a/Sources/BranchSDK/BNCPreferenceHelper.m
+++ b/Sources/BranchSDK/BNCPreferenceHelper.m
@@ -514,7 +514,7 @@ NSURL* /* _Nonnull */ BNCURLForBranchDirectory_Unthreaded(void);
 
 - (NSMutableString*) sanitizedMutableBaseURL:(NSString*)baseUrl_ {
     NSMutableString *baseUrl = [baseUrl_ mutableCopy];
-    if (self.trackingDisabled) {
+    if (self.attributionLevelNone) {
         NSString *id_string = [NSString stringWithFormat:@"%%24randomized_bundle_token=%@", self.randomizedBundleToken];
         NSRange range = [baseUrl rangeOfString:id_string];
         if (range.location != NSNotFound) [baseUrl replaceCharactersInRange:range withString:@""];
@@ -643,22 +643,6 @@ NSURL* /* _Nonnull */ BNCURLForBranchDirectory_Unthreaded(void);
 - (void) setDropURLOpen:(BOOL)value {
     @synchronized(self) {
         [self writeBoolToDefaults:@"dropURLOpen" value:value];
-    }
-}
-
-- (BOOL) trackingDisabled {
-    @synchronized(self) {
-        NSNumber *b = (id) [self readObjectFromDefaults:@"trackingDisabled"];
-        if ([b isKindOfClass:NSNumber.class]) return [b boolValue];
-        return false;
-    }
-}
-
-- (void) setTrackingDisabled:(BOOL)disabled {
-    @synchronized(self) {
-        NSNumber *b = [NSNumber numberWithBool:disabled];
-        [self writeObjectToDefaults:@"trackingDisabled" value:b];
-        if (disabled) [self clearTrackingInformation];
     }
 }
 
@@ -909,6 +893,22 @@ NSURL* /* _Nonnull */ BNCURLForBranchDirectory_Unthreaded(void);
 
 - (void)setAttributionLevel:(BranchAttributionLevel)level {
     [self writeObjectToDefaults:BRANCH_PREFS_KEY_ATTRIBUTION_LEVEL value:level];
+}
+
+- (BOOL)attributionLevelFull:(BranchAttributionLevel)level {
+    return [_attributionLevel isEqualToString:@"FULL"];
+}
+
+- (BOOL)attributionLevelReduced:(BranchAttributionLevel)level {
+    return [_attributionLevel isEqualToString:@"REDUCED"];
+}
+
+- (BOOL)attributionLevelMinimal:(BranchAttributionLevel)level {
+    return [_attributionLevel isEqualToString:@"MINIMAL"];
+}
+
+- (BOOL)attributionLevelNone:(BranchAttributionLevel)level {
+    return [_attributionLevel isEqualToString:@"NONE"];
 }
 
 - (NSString *) uxType {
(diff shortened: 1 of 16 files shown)


Sources/BranchSDK/BNCPreferenceHelper.m
```diff
@@ -911,6 +895,22 @@ - (void)setAttributionLevel:(BranchAttributionLevel)level {
     [self writeObjectToDefaults:BRANCH_PREFS_KEY_ATTRIBUTION_LEVEL value:level];
 }
 
+- (BOOL)attributionLevelFull:(BranchAttributionLevel)level {
```
NidhiDixit09: @bboothe-branch  `attributionLevelFull`, `attributionLevelReduced`, `attributionLevelMinimal` These functions are not used in the code. We should remove it, if its a dead code.

Sources/BranchSDK/Branch.m
```diff
@@ -564,39 +564,27 @@ - (void)setRequestMetadataKey:(NSString *)key value:(NSString *)value {
     [self.preferenceHelper setRequestMetadataKey:key value:value];
 }
 
++ (BOOL)attributionLevelFull {
```
NidhiDixit09: `attributionLevelFull`, `attributionLevelReduced` and `attributionLevelMinimal` - these functions are also not used. If not required, we can remove them.

Sources/BranchSDK/NSError+Branch.m
```diff
@@ -39,7 +39,7 @@ + (NSString *)messageForCode:(BNCErrorCode)code {
         [messages setObject:@"Spotlight indexing requires a title." forKey:@(BNCSpotlightTitleError)];
         [messages setObject:@"The Spotlight identifier is required to remove indexing from spotlight." forKey:@(BNCSpotlightIdentifierError)];
         [messages setObject:@"Spotlight cannot remove publicly indexed content." forKey:@(BNCSpotlightPublicIndexError)];
-        [messages setObject:@"User tracking is disabled and the request is not allowed" forKey:@(BNCTrackingDisabledError)];
+        [messages setObject:@"Cannot process className. Attribution Level is set to \"NONE\". To process this className call setConsumerProtectionAttributionLevel(\"FULL\")." forKey:@(BNCAttributionLevelNoneError)];
```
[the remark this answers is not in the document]
bboothe-branch: Applied change.

Sources/BranchSDK/BNCPreferenceHelper.m
```diff
@@ -911,6 +895,22 @@ - (void)setAttributionLevel:(BranchAttributionLevel)level {
...
+    return [_attributionLevel isEqualToString:@"REDUCED"];
+}
+
+- (BOOL)attributionLevelMinimal:(BranchAttributionLevel)level {
+    return [_attributionLevel isEqualToString:@"MINIMAL"];
+}
+
+- (BOOL)attributionLevelNone:(BranchAttributionLevel)level {
+    return [_attributionLevel isEqualToString:@"NONE"];
+}
```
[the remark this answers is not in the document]
bboothe-branch: Applied changes

Sources/BranchSDK/Branch.m
```diff
@@ -1987,8 +1966,8 @@ - (void)applicationDidBecomeActive {
         
         [[BranchLogger shared] logVerbose:[NSString stringWithFormat:@"applicationDidBecomeActive installOrOpenInQueue %d", installOrOpenInQueue] error:nil];
 
-        if (!Branch.trackingDisabled && self.initializationStatus == BNCInitStatusUninitialized && !installOrOpenInQueue) {
-            [[BranchLogger shared] logVerbose:[NSString stringWithFormat:@"applicationDidBecomeActive trackingDisabled %d initializationStatus %d installOrOpenInQueue %d", Branch.trackingDisabled, self.initializationStatus, installOrOpenInQueue] error:nil];
+        if (!Branch.attributionLevelNone && self.initializationStatus == BNCInitStatusUninitialized && !installOrOpenInQueue) {
```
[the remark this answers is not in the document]
bboothe-branch: Applied changes.

Sources/BranchSDK/Branch.m
```diff
@@ -2000,7 +1979,7 @@ - (void)applicationWillResignActive {
     [[BranchLogger shared] logVerbose:@"applicationWillResignActive" error:nil];
 
     dispatch_async(self.isolationQueue, ^(){
-        if (!Branch.trackingDisabled) {
+        if (!Branch.attributionLevelNone) {
```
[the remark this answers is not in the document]
bboothe-branch: Applied change

Sources/BranchSDK/BNCPreferenceHelper.m
bboothe-branch: Removed unused functions.

Sources/BranchSDK/Branch.m
bboothe-branch: Removed Unused functions

Sources/BranchSDK/Public/Branch.h
```diff
@@ -966,6 +944,17 @@ extern BranchAttributionLevel const BranchAttributionLevelNone;
  */
 - (void)setConsumerProtectionAttributionLevel:(BranchAttributionLevel)level resetSession:(BOOL)resetSession;
 
+/// Returns a boolean based on if the current Attribution Level is set to "FULL".
++ (BOOL) attributionLevelFull;
```
NidhiDixit09: @bboothe-branch Are these properties anywhere? If not please remove them.

NidhiDixit09 (approved):
Added one more comment. Please check. 
Also please go through comments from matterAi and resolve them if they are correct.

Sources/BranchSDK/Public/Branch.h
bboothe-branch: Removed these properties

Sources/BranchSDK/Branch.m
```diff
@@ -1987,8 +1948,8 @@ - (void)applicationDidBecomeActive {
         
         [[BranchLogger shared] logVerbose:[NSString stringWithFormat:@"applicationDidBecomeActive installOrOpenInQueue %d", installOrOpenInQueue] error:nil];
 
-        if (!Branch.trackingDisabled && self.initializationStatus == BNCInitStatusUninitialized && !installOrOpenInQueue) {
-            [[BranchLogger shared] logVerbose:[NSString stringWithFormat:@"applicationDidBecomeActive trackingDisabled %d initializationStatus %d installOrOpenInQueue %d", Branch.trackingDisabled, self.initializationStatus, installOrOpenInQueue] error:nil];
+        if (!Branch.attributionLevelNone && self.initializationStatus == BNCInitStatusUninitialized && !installOrOpenInQueue) {
+            [[BranchLogger shared] logVerbose:[NSString stringWithFormat:@"applicationDidBecomeActive attributionLevelNone %d initializationStatus %d installOrOpenInQueue %d", Branch.attributionLevelNone, self.initializationStatus, installOrOpenInQueue] error:nil];
```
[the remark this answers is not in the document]
bboothe-branch: Modified all calls to [Branch attributionLevelNone]

Sources/BranchSDK/BNCPreferenceHelper.m
```diff
@@ -911,6 +895,10 @@ - (void)setAttributionLevel:(BranchAttributionLevel)level {
     [self writeObjectToDefaults:BRANCH_PREFS_KEY_ATTRIBUTION_LEVEL value:level];
 }
 
+- (BOOL)attributionLevelNone {
+    return [_attributionLevel isEqualToString:@"NONE"];
```
wpinho-branch: I think this helper has a subtle bug: it compares against the `_attributionLevel` ivar, but in this class the `attributionLevel` getter and setter both go straight to NSUserDefaults — neither of them ever assigns the ivar. So `_attributionLevel` stays nil and this method returns NO even when the stored level is actually NONE.

The practical impact is in `sanitizedMutableBaseURL:` — with the level set to NONE it would still keep the `$randomized_bundle_token` in the URL, which is exactly what we're supposed to strip in that mode.

Suggest:
```objc
- (BOOL)attributionLevelNone {
    return [[self attributionLevel] isEqualToString:BranchAttributionLevelNone];
}
```
(also swaps the `@"NONE"` literal for the constant). Probably worth a quick unit test that sets the level and asserts this returns YES, since nothing caught it.

wpinho-branch:
One behavioral difference vs. the old code that I want to make sure is intentional: `setTrackingDisabled:YES` used to clear the user's tracking info immediately, inside the setter. In the new flow, setting the level to NONE clears the queue and the link cache, but `clearTrackingInformation` only happens later, when a non-linking request gets dropped in BNCServerInterface. So there's a window where identity/tokens are still on disk even though the user opted out — and if the app never fires another request, they stay there.

If that's acceptable for the privacy semantics we want, fine — but it feels safer to clear the tracking info right away in `setConsumerProtectionAttributionLevel:` when the level becomes NONE, like the old setter did.

wpinho-branch:
Tiny thing: the testing instructions mention setting the attribution level via the button on the TestBed home page, but this PR removes that button from the storyboard. Might be worth updating the description so QA doesn't go looking for it.

Sources/BranchSDK/BNCPreferenceHelper.m
bboothe-branch: Modified to suggestion. I prefer this!

bboothe-branch:
> One behavioral difference vs. the old code that I want to make sure is intentional: `setTrackingDisabled:YES` used to clear the user's tracking info immediately, inside the setter. In the new flow, setting the level to NONE clears the queue and the link cache, but `clearTrackingInformation` only happens later, when a non-linking request gets dropped in BNCServerInterface. So there's a window where identity/tokens are still on disk even though the user opted out — and if the app never fires another request, they stay there.
> 
> If that's acceptable for the privacy semantics we want, fine — but it feels safer to clear the tracking info right away in `setConsumerProtectionAttributionLevel:` when the level becomes NONE, like the old setter did.

Hey @wpinho-branch I modified the preferencehelper to this. What do you think?

```
- (void)setAttributionLevel:(BranchAttributionLevel)level {
    if (level == BranchAttributionLevelNone) {
        [self clearTrackingInformation];
    }
    [self writeObjectToDefaults:BRANCH_PREFS_KEY_ATTRIBUTION_LEVEL value:level];
}
```

bboothe-branch:
> Tiny thing: the testing instructions mention setting the attribution level via the button on the TestBed home page, but this PR removes that button from the storyboard. Might be worth updating the description so QA doesn't go looking for it.

There were actually two buttons at first. I removed the second one. One is still there near the bottom titled Consumer Protection Preference

image

Sources/BranchSDK/BNCPreferenceHelper.m
```diff
@@ -908,9 +892,16 @@ - (BranchAttributionLevel)attributionLevel {
 }
 
 - (void)setAttributionLevel:(BranchAttributionLevel)level {
+    if (level == BranchAttributionLevelNone) {
```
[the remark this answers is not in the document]
bboothe-branch: changed to suggestion

wpinho-branch:
> Hey @wpinho-branch I modified the preferencehelper to this. What do you think?

That works — moving it into the setter means any path that flips the level to NONE clears immediately, which was exactly the gap I was worried about. Good catch from the bot on the string comparison too. LGTM on this one.

wpinho-branch:
> There were actually two buttons at first. I removed the second one. One is still there near the bottom titled Consumer Protection Preference

Ah, that explains it. I only saw the removal in the diff. Thanks for the screenshot, all good then.

wpinho-branch (approved):

PR #1593, opened 2026-05-28
[In Progress] Double opens in scenes issue fix

NidhiDixit09:
## Reference
SDK-XXXX -- <TITLE>.

## Type Of Change

## Testing Instructions


Sources/BranchSDK/BranchScene.m
```diff
@@ -35,6 +35,44 @@ - (void)initSessionWithLaunchOptions:(nullable NSDictionary *)options registerDe
...
+        // exists; the value is never read. The activity itself is delivered to Branch below
+        // via -scene:continueUserActivity:. We populate only the documented type key here —
+        // UIKit exposes no public constant for the NSUserActivity object within this dictionary.
+        launchOptions[UIApplicationLaunchOptionsUserActivityDictionaryKey] = @{
+            UIApplicationLaunchOptionsUserActivityTypeKey: activity.activityType ?: @""
+        };
+    }
+
+    if (connectionOptions.URLContexts.count) {
+        UIOpenURLContext *context = connectionOptions.URLContexts.allObjects.firstObject;
```
wpinho-branch, line 55: `userActivities` and `URLContexts` are both sets, so `allObjects.firstObject` picks an arbitrary element whenever there's more than one. I read this as non-deterministic rather than wrong, and I didn't build a multi-activity case to see whether it matters in practice.

Does a scene connection ever hand you more than one? If it can, picking the activity whose type is `NSUserActivityTypeBrowsingWeb` would be deterministic. If it can't, a line saying so would save the next reader the same question.

wpinho-branch (commented):
The core fix is right. Narrowing the guard so it only proceeds from the uninitialized state closes the window where a session already initializing still got through, which is the double open. Threading `sceneIdentifier` through instead of passing nil is the right follow-through.

What's holding it isn't the code. The title still says In Progress, the description is the untouched template, and the branch has conflicts. The fastlane bump, the Gemfile, the Fastfile and the workflow files are a separate change riding along, and those same workflows are touched by #1602, so one of the two will conflict with the other.

One design question that's bigger than this PR: this adds a scene init entry point on master while #1605 adds a different one on the 4.0 branch. Two scene APIs designed in parallel on two branches will have to be reconciled at some point, and it's cheaper to do that now than after both land. Worth a word with @bboothe-branch.

I didn't run this branch.

PR #1595, opened 2026-06-02
Audit duplicate code from BranchRequestOpen and BranchRequestDeepLink

bboothe-branch:
## Reference
EMT-3681-- Audit duplicate code from BranchRequestOpen and BranchRequestDeepLink

## Motivation

Reduce the unnecessary code to fields handled individually by the new separate classes.

## Type Of Change

## Testing Instructions

### DIFF
--- a/Sources/BranchSDK/BranchRequestDeepLink.m
+++ b/Sources/BranchSDK/BranchRequestDeepLink.m
@@ -251,8 +251,23 @@ static BOOL deepLinkRequestWaitQueueIsSuspended = NO;
         [[BranchLogger shared] logDebug:[NSString stringWithFormat:@"~referring_link found in response: %@, sending sendOpen network request." ,referringURL] error:nil];
         [[Branch getInstance] sendOpen:response.data skipCallback:skipCallback];
     } else {
-        [[BranchLogger shared] logDebug:@"No ~referring_link on deeplink data. Not sending sendOpen network request." error:nil];
+        [[BranchLogger shared] logDebug:@"No ~referring_link on deeplink data. Not sending sendOpen network request. Clearing link identifiers to prevent reuse." error:nil];
+        [self clearLinkIdentifiers:preferenceHelper];
     }
+}
+
+- (void)clearLinkIdentifiers:(BNCPreferenceHelper *)preferenceHelper {
+    // Clear link identifiers so they don't get reused on next open.
+    // This matches cleanup done in BranchRequestOpen.processResponse
+    preferenceHelper.linkClickIdentifier = nil;
+    preferenceHelper.spotlightIdentifier = nil;
+    preferenceHelper.universalLinkUrl = nil;
+    preferenceHelper.externalIntentURI = nil;
+    preferenceHelper.referringURL = nil;
+    preferenceHelper.initialReferrer = nil;
+    preferenceHelper.dropURLOpen = NO;
+    preferenceHelper.uxType = nil;
+    preferenceHelper.urlLoadMs = nil;
 }
 
 @end
--- a/Sources/BranchSDK/BranchRequestDeepLink.m
+++ b/Sources/BranchSDK/BranchRequestDeepLink.m
@@ -16,6 +16,11 @@
 #import "BNCServerAPI.h"
 #import "BNCInAppBrowser.h"
 
+// Forward declaration of private Branch method
+@interface Branch (PrivateMethods)
+- (void)sendOpen:(NSDictionary *)responseData skipCallback:(BOOL)skipCallback;
+@end
+
 @implementation BranchRequestDeepLink
 
 - (id)initWithCallback:(callbackWithStatus)callback {
--- a/Sources/BranchSDK/BranchRequestDeepLink.m
+++ b/Sources/BranchSDK/BranchRequestDeepLink.m
@@ -196,10 +196,9 @@ static BOOL deepLinkRequestWaitQueueIsSuspended = NO;
     if (self != [BranchRequestDeepLink self])
         return;
     deepLinkRequestWaitQueue =
-        dispatch_queue_create("io.branch.sdk.openqueue", DISPATCH_QUEUE_CONCURRENT);
+        dispatch_queue_create("io.branch.sdk.deeplinkqueue", DISPATCH_QUEUE_CONCURRENT);
 }
 
-// Need to Update names for all of these to deepLinkRespone Lock
 + (void) setWaitNeededForDeepLinkResponseLock {
     @synchronized (self) {
         if (!deepLinkRequestWaitQueueIsSuspended) {
@@ -210,7 +209,6 @@ static BOOL deepLinkRequestWaitQueueIsSuspended = NO;
     }
 }
 
-// Need to Update names for all of these to deepLinkRespone Lock
 + (void) waitForDeepLinkResponseLock {
     [[BranchLogger shared] logVerbose:@"Waiting for deepLinkRequestWaitQueue." error:nil];
     dispatch_sync(deepLinkRequestWaitQueue, ^ {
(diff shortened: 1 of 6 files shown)


Sources/BranchSDK/Private/BranchRequestDeepLink.h
```diff
@@ -19,9 +19,9 @@
...
 
-+ (void) waitForOpenResponseLock;
-+ (void) releaseOpenResponseLock;
-+ (void) setWaitNeededForOpenResponseLock;
++ (void) waitForDeepLinkResponseLock;
++ (void) releaseDeepLinkResponseLock;
++ (void) setWaitNeededForDeepLinkResponseLock;
 
 - (id)initWithCallback:(callbackWithStatus)callback;
 - (id)initWithCallback:(callbackWithStatus)callback isInstall:(BOOL)isInstall;
```
wpinho-branch, line 27: The header still declares `initWithCallback:isInstall:`, but this PR removes its implementation (along with the `isInstall` property). Anyone calling it will compile fine and crash at runtime with an unrecognized selector. Since the install concept now lives entirely in BranchRequestOpen, let's drop the declaration too.

Sources/BranchSDK/BranchRequestDeepLink.m
```diff
@@ -75,202 +56,31 @@ - (void)processResponse:(BNCServerResponse *)response error:(NSError *)error {
...
     }
 
-    if (preferenceHelper.referringURL != nil) {
-        [[BranchLogger shared] logDebug:[NSString stringWithFormat:@"~referring_link found in response: %@, sending sendOpen network request." ,preferenceHelper.referringURL] error:nil];
-        [[Branch getInstance] sendOpen:response.data];
-    } else {
-        [[BranchLogger shared] logDebug:[NSString stringWithFormat:@"No referring URL on deeplink data. Not sending sendOpen network request."] error:nil];
-    }
+    // Normal flow - send attribution and allow initialization callback
+    [self attemptToSendOpen:preferenceHelper response:response skipCallback:NO];
```
wpinho-branch, line 88: With this slimming, the deeplink request no longer persists session params / install params and no longer clears the link identifiers (linkClickIdentifier, spotlightIdentifier, universalLinkUrl, externalIntentURI...). That's all delegated to the open request now — which makes sense — but the open only fires when we find a `~referring_link`.

So in the path where a deeplink response comes back without a referring link: who clears those identifiers? If nothing does, the next open could reuse a stale link identifier from the previous deeplink. Same question for CPP = NONE, where the open is intentionally suppressed but a deeplink may have set identifiers. If there's a cleanup elsewhere that I'm missing, pointing me to it is enough — otherwise this probably needs handling and a test.

Sources/BranchSDK/Public/Branch.h
```diff
@@ -563,7 +563,7 @@ extern NSString * __nonnull const BNCSpotlightFeature;
  */
 - (void)sendOpen;
 
-- (void)sendOpen:(NSDictionary *)responseData;
+- (void)sendOpen:(NSDictionary *)responseData skipCallback:(BOOL)skipCallback;
```
wpinho-branch: Since this is public API on the beta, a short doc comment on `skipCallback` would help — it's not obvious from the signature that "skip" means "we're redirecting to web, send attribution but don't fire the init callback".

Sources/BranchSDK/BranchRequestDeepLink.m
```diff
@@ -378,45 +188,74 @@ + (BOOL)supportsSecureCoding {
...
+static dispatch_queue_t deepLinkRequestWaitQueue = NULL;
+static BOOL deepLinkRequestWaitQueueIsSuspended = NO;
 
 
 + (void) initialize {
     if (self != [BranchRequestDeepLink self])
         return;
-    openRequestWaitQueue =
+    deepLinkRequestWaitQueue =
         dispatch_queue_create("io.branch.sdk.openqueue", DISPATCH_QUEUE_CONCURRENT);
```
wpinho-branch: Nice that the lock got properly renamed. Two leftovers from the rename: the dispatch queue label is still `io.branch.sdk.openqueue` (handy to rename for debugging sessions, you'll thank yourself in Instruments), and the `// Need to Update names for all of these to deepLinkRespone Lock` comments can go now that it's done.

wpinho-branch:
Before diving into review: the `verify-manually-with-xcframework-iOS` job crashed (abort trap 6 during the iOSReleaseTest run, not a compile failure). Given this PR touches the suspend/resume logic of the deep link wait queue, I'd want to rule out that a suspended queue is being left hanging before we call it a flake. The xcresult is in the job artifacts — can you take a look or re-run and confirm?

wpinho-branch:
One more thing I noticed while reading the lock handling on the beta branch, since it's close to this PR's scope: we suspend the wait queues for both BranchRequestDeepLink and BranchRequestOpen in BNCServerRequestOperation, but nothing ever calls `waitFor...` on either of them — the only consumer of this mechanism is `getLatestReferringParamsSynchronous`, which still waits on the legacy BranchOpenRequest lock. So in the new flow that public API doesn't actually synchronize with the open anymore, and the suspend/resume cycle on the new classes is overhead with no reader. Feels like part of this audit: either wire the synchronous getter to the new locks or drop the suspends. Related: if an operation gets cancelled after the suspend, processResponse never runs and the queue stays suspended forever — harmless today, a deadlock the day someone adds a waiter.

Sources/BranchSDK/Private/BranchRequestDeepLink.h
bboothe-branch: Removed the declaration.

Sources/BranchSDK/Public/Branch.h
bboothe-branch: Thanks for this @wpinho-branch I'm going to check with Gabe on this. I think we decided no callback related parameters on the public API but want to confirm.

Sources/BranchSDK/BranchRequestDeepLink.m
bboothe-branch: Updated with new name for the logs

bboothe-branch:
> One more thing I noticed while reading the lock handling on the beta branch, since it's close to this PR's scope: we suspend the wait queues for both BranchRequestDeepLink and BranchRequestOpen in BNCServerRequestOperation, but nothing ever calls `waitFor...` on either of them — the only consumer of this mechanism is `getLatestReferringParamsSynchronous`, which still waits on the legacy BranchOpenRequest lock. So in the new flow that public API doesn't actually synchronize with the open anymore, and the suspend/resume cycle on the new classes is overhead with no reader. Feels like part of this audit: either wire the synchronous getter to the new locks or drop the suspends. Related: if an operation gets cancelled after the suspend, processResponse never runs and the queue stays suspended forever — harmless today, a deadlock the day someone adds a waiter.

I just added those waits to the `getLatestReferringParamsSynchronous` function in the latest commit. Thank you!

Sources/BranchSDK/BranchRequestDeepLink.m
bboothe-branch: Hey @wpinho-branch thanks for this! I updated the code to include clearing link identifiers. Now it should work in all of these scenarios.
1. ~referring link, CPP level NOT set to NONE
2.  ~  WITHOUT ~referring_link: BranchRequestDeepLink clears them (new)
3. ~  WITH ~referring_link, CPP = NONE: Branch.sendOpen clears them (new)

Sources/BranchSDK/Public/Branch.h
bboothe-branch: Went with suggestion to privatize this API!

wpinho-branch (approved):

NidhiDixit09 (approved):

PR #1599, opened 2026-06-12
Bump actions/checkout from 4 to 6 on 4.0.0-beta.0 (cherry-pick #1547)

wpinho-branch:
## Reference
No ticket. CI maintenance on the `4.0.0-beta.0` branch.

## Summary
Cherry-pick of #1547 (already merged on master): bumps `actions/checkout` from v4 to v6 across all workflows on the `4.0.0-beta.0` branch.

## Motivation
GitHub will force all actions to run on Node 24 starting June 16th, 2026 ([announcement](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/)). `checkout@v4` runs on Node 20. Since PR CI runs the workflows from the PR's branch, the open beta PRs (#1592, #1595) could start failing next week for reasons unrelated to their code. Master already received this bump via #1547; this brings the beta branch in line.

Note: `upload-artifact@v4`/`@v3` and `cache@v3` are also Node 20 and still present on both master and this branch. Dependabot should cover those next (#1597 and #1598 are open on master); keeping this PR scoped to the already-validated checkout bump.

## Type Of Change
- [x] CI/build maintenance (no SDK code changes)

## Testing Instructions
CI on this PR is the test: all verify jobs should run and pass with `checkout@v6`, same as they already do on master.

cc @BranchMetrics/saas-sdk-devs for visibility.

PR #1596, opened 2026-06-03
EMT-3754 - replaced inituserSessionAndCallCallback with sendOpen methods

bboothe-branch:
## Reference
EMT-3754 - replaced inituserSessionAndCallCallback with sendOpen methods

## Summary

Replacing old session initialization methods with new functions.

## Motivation

Without session initialization other Branch features would not work.

## Type Of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Testing Instructions

1. Start app
2. Click sendCommerceEvent to send event.
3. Click Create Branch Link to create link
4. Confirm 200 responses on each.

cc @BranchMetrics/saas-sdk-devs for visibility.

PR #1600, opened 2026-06-12
Fix dropped initSession callback when auto sendOpen is in flight (session-ready bridge)

wpinho-branch:
## Reference
No ticket yet (follow-up of EMT-3754 / #1596). Found while investigating the CI failure on #1599.

## Summary
Draft PR to discuss the fix direction for a race introduced on `4.0.0-beta.0`: a deep link handler registered via the legacy `initSession` while the automatic `sendOpen` is still in flight is never invoked.

## The race
1. App becomes active. Since #1596, `applicationDidBecomeActive` auto-fires `sendOpen` and the session status goes to `Initializing`.
2. The app calls `initSession(launchOptions:andRegisterDeepLinkHandler:)`. `initUserSessionAndCallCallback` only handles `Uninitialized` (starts an init) and `Initialized` (calls back immediately). In the `Initializing` state it silently does nothing.
3. The auto open completes with `handleInitSuccessAndCallCallback:NO`, intentionally skipping developer callbacks (opens should not return data to the app).
4. The handler registered in step 2 is never invoked. `testInitSessionAndSetCPPLevel` times out exactly here, which is why the integration suite has been flaky since #1596 merged (it passes or fails depending on which side of the in-flight window `initSession` lands).

The old flow did not have this gap because the auto path was `initUserSessionAndCallCallback:YES`, so the in-flight open invoked registered handlers on completion.

## Proposed fix (this PR)
Keep the principle that opens stay silent and do not return data to the app. Instead, separate "deep link data" (exclusive to `requestDeepLinkData`) from "the session has settled" (owned by the open flow), via a small internal primitive:

- `notifyWhenSessionReady:` queues blocks while an open is in flight (invokes immediately if the session is already initialized).
- `handleInitSuccessAndCallCallback:` / `handleInitFailure:` flush the queue on silent settles, and discard it on `callCallback:YES` settles (the single registered handler was already invoked directly, flushing would double-call it).
- The legacy `initSession` flow bridges through it: the previously-missing `Initializing` branch in `initUserSessionAndCallCallback` now queues the callback instead of dropping it.

This also moves session-state ownership into the `sendOpen` flow in the current build, and the primitive is independent of `initUserSessionAndCallCallback`, so it survives that function's planned removal.

## Alternative considered
Restoring `YES` in the auto `sendOpen` completion is a two-word fix that recreates the 3.x behavior, but it re-couples opens to developer callbacks, which is what the rewrite is moving away from. Listed here for completeness; happy to switch if preferred.

## Testing Instructions
- `testInitSessionAndSetCPPLevel` in the integration suite covers the race (it is the test that exposed it on #1599's CI). Re-running the suite a few times is the meaningful check, since the failure is timing dependent.
- Manual: launch TestBed, let the app go active (auto open fires), then call `initSession` and confirm the callback arrives.

Would appreciate a review of the direction before this leaves draft.

### DIFF
--- a/Branch-TestBed/Branch-SDK-Tests/BranchClassTests.m
+++ b/Branch-TestBed/Branch-SDK-Tests/BranchClassTests.m
@@ -18,6 +18,12 @@
 - (void)writeObjectToDefaults:(NSString *)key value:(NSObject *)value;
 @end
 
+
+// Expose private session-state internals used to reproduce the in-flight open window.
+@interface Branch (SessionReadySettleTest)
+- (void)handleInitSuccessAndCallCallback:(BOOL)callCallback sceneIdentifier:(NSString *)sceneIdentifier;
+@end
+
 @interface BranchClassTests : XCTestCase
 @property (nonatomic, strong) Branch *branch;
 @property (nonatomic, strong, readwrite) BNCPreferenceHelper *prefHelper;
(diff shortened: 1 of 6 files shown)


wpinho-branch:
Some verification notes beyond the green CI, since a race fix deserves more than one green run.

The second commit adds a deterministic regression test (`testInitSessionCallbackRegisteredWhileOpenInFlightIsInvoked`) that forces the exact state of the race: status set to Initializing, handler registered via legacy `initSession`, then a silent settle (`handleInitSuccessAndCallCallback:NO`, which is how the auto open completes). On the current beta branch this test fails every time with the same timeout we saw in CI; with this PR it passes. So the failure mode is reproducible on demand now, not just under CI timing.

I also stress tested with repeated fresh app launches of `testInitSessionAndSetCPPLevel` (the integration test that exposed this): 12 out of 12 launches pass on this branch. For honesty: the baseline also passed 12 local launches, the window is too narrow to hit on a fast local machine, which is consistent with it only manifesting on the slower CI runners (twice in the last two days).

CI is green across two full rounds, the second one running the new regression test in the verify job.

wpinho-branch:
We will no longer have `init session` in the beta; it has been changed to `handle deeplink` and `open/install`.

PR #1602, opened 2026-06-21
Removed synchronous user agent fetching on initialization and from v1 and v2 events.

NidhiDixit09:
## Reference
https://branch.atlassian.net/browse/EMT-3846

## Summary

User agent collection via WKWebView was blocking the serial isolationQueue during initSession, adding ~800–2000ms to cold start latency. The user_agent field is not required for v1/open, v1/install, or v2/event API requests.

Changes

1. Removed user_agent from all API requests — deleted from both v1 (updateDeviceInfoToMutableDictionary:) and v2 (v2dictionary) request builders in BNCRequestFactory.m.
2. Removed the blocking loadUserAgent call from init — previously, Branch.m dispatched user agent loading on the isolationQueue with a semaphore wait (blocking init). This is replaced with a fire-and-forget background warmup that never blocks initSession.
3. Made BNCUserAgentCollector.userAgent atomic — since the property is now written asynchronously and may be read concurrently (for share link cache-busting), it was changed from nonatomic to atomic.
4. Removed userAgentString from BNCDeviceInfo — the getter and its #import are deleted since no callers remain.
5. CI: Ruby 2.7 → 3.3 — all GitHub Actions workflows updated from deprecated Ruby 2.7 to 3.3.
6. Added concurrency tests — new tests verify: async warmup populates the property correctly, and concurrent reads during the background write don't crash.
7. Integration test app — AppDelegate.swift now includes Branch SDK initialization with timing instrumentation ([BranchPerf] logs).


Sources/BranchSDK/Branch.m
```diff
@@ -1062,12 +1062,12 @@ - (void)handlePushNotification:(NSDictionary *)userInfo {
 
 - (void)loadUserAgent {
     #if !TARGET_OS_TV
-    dispatch_async(self.isolationQueue, ^(){
-        dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
-        [[BNCUserAgentCollector instance] loadUserAgentWithCompletion:^(NSString * _Nullable userAgent) {
-            dispatch_semaphore_signal(semaphore);
-        }];
-        dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER);
+    // Fire-and-forget warmup so the user agent is cached for share link generation.
```
gdeluna-branch: Does share link await this until user agent is ready?

gdeluna-branch (approved):

bboothe-branch:
Tested and saw no issues with running build!

Sources/BranchSDK/Branch.m
wpinho-branch: Answering this since it's still open. No, share link generation doesn't wait on the user agent, and it doesn't go through the getter this PR removes either.

`BranchShareLink`, `BranchActivityItemProvider` and `BranchUniversalObject` all read `[BNCUserAgentCollector instance].userAgent` directly rather than through `deviceInfo.userAgentString`, so deleting that getter doesn't touch the `ignoreUAString` path. What they can get is nil, in the window the new comment in `loadUserAgent` describes, and then the link is created without the UA hint.

wpinho-branch (commented):
Not blocking, the change reads correct to me. I answered the open thread about share links, and there's one cross-platform note below.

The Ruby bump touches the same four workflow files as #1593, so whichever of the two lands second will conflict.

I didn't run this branch.

Sources/BranchSDK/BNCRequestFactory.m
```diff
@@ -642,7 +642,6 @@ - (NSDictionary *)v2dictionary {
         [dictionary bnc_safeSetObject:self.deviceInfo.country forKey:@"country"];
```
wpinho-branch, line 642: After this, iOS stops sending `user_agent` and Android keeps sending it. On the Android side `DeviceInfo` still writes it into the user data object in three places.

I can't see the backend from here, so I don't know whether that matters. If the field really is unused for v1/open, v1/install and v2/event, the asymmetry is fine, but it's worth writing down somewhere so nobody adds it back later. If it isn't, the two SDKs quietly start reporting different things, and that surfaces as a platform gap rather than as a bug in this PR.

PR #1603, opened 2026-06-22
EMT-3808 -- Removing Initialization States and Session ID

bboothe-branch:
## Reference
EMT-3808 -- Removing Initialization States and Session ID.

## Summary

- Removed all of the calls to sessionID in the Branch SDK
- Removed all of the checks for InitializationStatus in SDK

## Motivation

Session ID is not used in anything for Branch's backend, and it is only used for checking certain fields in the SDK. None of these help SDK functionality. Same for InitializationStatus. All of the calls simply make the SDK less effective and more cluttered. Removing these makes the SDK cleaner and more simple to use and run.

## Type Of Change

## Testing Instructions

1. Run the Branch testbed app
2. Verify none of the app breaks without previous session calls. Use buttons to test.


Sources/BranchSDK/Branch.m
```diff
@@ -1946,8 +1868,8 @@ - (void)applicationDidBecomeActive {
         
         [[BranchLogger shared] logVerbose:[NSString stringWithFormat:@"applicationDidBecomeActive installOrOpenInQueue %d", installOrOpenInQueue] error:nil];
 
-        if (![Branch attributionLevelNone] && self.initializationStatus == BNCInitStatusUninitialized && !installOrOpenInQueue) {
-            [[BranchLogger shared] logVerbose:[NSString stringWithFormat:@"applicationDidBecomeActive attributionLevelNone %d initializationStatus %d installOrOpenInQueue %d", [Branch attributionLevelNone], self.initializationStatus, installOrOpenInQueue] error:nil];
+        if (![Branch attributionLevelNone] && !installOrOpenInQueue) {
+            [[BranchLogger shared] logVerbose:[NSString stringWithFormat:@"applicationDidBecomeActive attributionLevelNone %d initializationStatus %d installOrOpenInQueue %d", [Branch attributionLevelNone], installOrOpenInQueue] error:nil];
```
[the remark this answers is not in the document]
bboothe-branch: Deleted the extra %d in the logger code.

Sources/BranchSDK/Branch.m
```diff
@@ -1685,18 +1506,6 @@ - (void)generateShortUrl:(NSArray *)tags
              andCampaign:campaign andParams:(NSDictionary *)params
              andCallback:(callbackWithUrl)callback {
 
-    @synchronized (self) {
```
NidhiDixit09, line 1688: @bboothe-branch If These APIs(v1/url, v2/event) are called before sending OPENs, these APIs will have missing `randomized_device_token` . Is that fine ?

Sources/BranchSDK/BNCInAppBrowser.m
```diff
@@ -76,10 +76,6 @@ - (void)openURLInSafariVC:(NSURL *)url overViewController:(UIViewController *)to
     [topVC presentViewController:safariVC animated:YES completion:nil];
 }
 
-- (void)safariViewControllerDidFinish:(SFSafariViewController *)controller {
```
NidhiDixit09, line 79: @bboothe-branch This should not be deleted. It should be replaced with appropriate request.

Sources/BranchSDK/Branch.m
```diff
@@ -866,8 +726,7 @@ - (BOOL)handleDeepLink:(NSURL *)url sceneIdentifier:(NSString *)sceneIdentifier
         NSString *urlString = [url absoluteString];
         self.preferenceHelper.externalIntentURI = urlString;
         self.preferenceHelper.referringURL = urlString;
-
-        [self initUserSessionAndCallCallback:YES sceneIdentifier:sceneIdentifier urlString:nil reset:YES];
```
NidhiDixit09, line 870: @bboothe-branch It should be replaced with corresponding new API.

Sources/BranchSDK/Branch.m
```diff
@@ -915,7 +774,7 @@ - (BOOL)handleSchemeDeepLink_private:(NSURL*)url sceneIdentifier:(NSString *)sce
             self.preferenceHelper.linkClickIdentifier = params[@"link_click_id"];
         }
     }
-    [self initUserSessionAndCallCallback:YES sceneIdentifier:sceneIdentifier urlString:url.absoluteString reset:YES];
```
NidhiDixit09, line 918: @bboothe-branch It should be replaced with corresponding new API.

Sources/BranchSDK/Branch.m
```diff
@@ -991,7 +847,6 @@ - (BOOL)continueUserActivity:(NSUserActivity *)userActivity sceneIdentifier:(NSS
     }
     #endif
 
-    [self initUserSessionAndCallCallback:YES sceneIdentifier:sceneIdentifier urlString:userActivity.webpageURL.absoluteString reset:YES];
```
NidhiDixit09, line 994: @bboothe-branch It should be replaced with corresponding new API.

Sources/BranchSDK/Branch.m
```diff
@@ -1175,11 +1030,8 @@ - (void)logout {
...
 
 - (void)logoutWithCallback:(callbackWithStatus)callback {
-    if (self.initializationStatus == BNCInitStatusUninitialized) {
-        NSError *error =
-            ([Branch attributionLevelNone])
-            ? [NSError branchErrorWithCode:BNCAttributionLevelNoneError]
-            : [NSError branchErrorWithCode:BNCInitError];
+    if ([Branch attributionLevelNone]) {
+        NSError *error = [NSError branchErrorWithCode:BNCAttributionLevelNoneError];
         [[BranchLogger shared] logWarning:@"Branch is not initialized, cannot logout." error:error];
```
NidhiDixit09: @bboothe-branch `Branch is not initialized` warning message does not match with the check `if ([Branch attributionLevelNone])`

NidhiDixit09:
@bboothe-branch What about previous deep linking related methods like as `continueUserActivity` , `- (BOOL)application:(UIApplication *)application
           openURL:(NSURL *)url
  sourceApplication:(NSString *)sourceApplication
         annotation:(id)annotation ` ... etc.  Since these functions are replaced with `requestDeepLinkData` , should we delete them also ?

Sources/BranchSDK/BNCInAppBrowser.m
bboothe-branch: Updated function with new function for requestDeepLinkData

Sources/BranchSDK/Branch.m
bboothe-branch: Updated function with new function for requestDeepLinkData

Sources/BranchSDK/Branch.m
bboothe-branch: Connected with @wpinho-branch on this and he put together this PR to merge into it to resolve this issue: https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/pull/1617

Sources/BranchSDK/Branch.m
bboothe-branch: Updated with requestDeepLink data API

Sources/BranchSDK/Branch.m
bboothe-branch: Replaced with requestDeepLinkData API

Sources/BranchSDK/Branch.m
bboothe-branch: Updated error message

bboothe-branch:
> @bboothe-branch What about previous deep linking related methods like as `continueUserActivity` , `- (BOOL)application:(UIApplication *)application openURL:(NSURL *)url sourceApplication:(NSString *)sourceApplication annotation:(id)annotation ` ... etc. Since these functions are replaced with `requestDeepLinkData` , should we delete them also ?

Hey @NidhiDixit09. We technically could replace them, but it comes with some issues. 

1. The legacy methods (`continueUserActivity`, `handleDeepLink`, `openURL`, return `(BOOL)`, whereas the `requestDeepLinkData`* functions return `(void)`. This is mostly due to internal callers, but it's worth flagging.
2. Some of these were used in BranchScene.m. in `scene:scene continueUserActivity:useractivity` and `scene:scene OpenURLContexts:URLContexts` also. There is also a call in passPasteItemProviders to handleDeepLink. Currently, all they do is processing of the URLs. Swapping these to `requestDeepLinkData`* would change them from "store the URL" to "store + enqueue a request + require a callback." Is this okay?
3. I could instead swap with the processing functions themselves, but that would require some header changes to make them accessible to ViewControllers and BranchScene.

If we do this I'd make the change on this PR to be concise: https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/pull/1605/changes#top

Sources/BranchSDK/Branch.m
wpinho-branch: Checking in on this one since it's the thread still open. #1617, the PR pointed to as the resolution, is still open, and its base is `bboothe-branch/EMT-3808`, so it targets this branch rather than the beta.

That means the answer to the missing `randomized_device_token` question isn't in this diff and can't be reviewed here. Would it be worth saying so in the description, so whoever picks this up next knows the order is #1617 into this branch first, then this branch into 4.0.0-beta.0? Right now it reads as self-contained.

NidhiDixit09 (approved):

PR #1605, opened 2026-06-29
EMT-3876: Added Convenience Functions to call requestDeepLinkData in AppDelegate

bboothe-branch:
## Reference
EMT-3876 -- Convenience Functions for AppDelegate and SceneDelegate.

## Type Of Change

## Testing Instructions

### DIFF
--- a/Sources/BranchSDK/Branch.m
+++ b/Sources/BranchSDK/Branch.m
@@ -1499,7 +1499,7 @@ static NSString *bnc_branchKey = nil;
                 else if ([Branch isBranchLink:url.absoluteString]) {
                     [self.preferenceHelper setLocalUrl:[url absoluteString]];
                     // 3. Send Open Event
-                    [[Branch getInstance] handleDeepLink:url];
+                    [[Branch getInstance] requestDeepLinkDataWithURL:url];
                 }
             }];
         }
--- a/Sources/BranchSDK/Branch.m
+++ b/Sources/BranchSDK/Branch.m
@@ -2045,7 +2045,12 @@ static inline void BNCPerformBlockOnMainThreadSync(dispatch_block_t block) {
 // Called from application:openURL:options:
 // Supports URI Schemes
 - (void)requestDeepLinkDataWithOpenURL:(NSURL *)url {
-    [[Branch getInstance] requestDeepLinkData:[url absoluteString] callback:^(NSDictionary *params, NSError *error) {
+    
+    NSString *urlStr = url.absoluteString;
+    
+    if (!urlStr.length) return;
+    
+    [[Branch getInstance] requestDeepLinkData:urlStr callback:^(NSDictionary *params, NSError *error) {
         if (error == nil) {
             if (params != nil) {
                 [[BranchLogger shared] logDebug:[NSString stringWithFormat:@"Deep Link Params: %@", params] error:nil];
@@ -2059,7 +2064,11 @@ static inline void BNCPerformBlockOnMainThreadSync(dispatch_block_t block) {
 // Called from application:continueUserActivity:restorationHandler:
 // Supports Universal Links
 - (void)requestDeepLinkDataWithContinueUserActivity:(NSUserActivity *)userActivity {
-    [[Branch getInstance] requestDeepLinkData:userActivity.webpageURL.absoluteString callback:^(NSDictionary *params, NSError *error) {
+    NSString *urlStr = userActivity.webpageURL.absoluteString;
+    
+    if (!urlStr) return;
+    
+    [[Branch getInstance] requestDeepLinkData:urlStr callback:^(NSDictionary *params, NSError *error) {
         if (error == nil) {
             if (params != nil) {
                 [[BranchLogger shared] logDebug:[NSString stringWithFormat:@"Deep Link Params: %@", params] error:nil];
@@ -2073,19 +2082,20 @@ static inline void BNCPerformBlockOnMainThreadSync(dispatch_block_t block) {
 // Called from application:didReceiveRemoteNotification:fetchCompletionHandler:
 // Supports Push Notifications
 - (void)requestDeepLinkDataFromDidReceiveRemoteNotification:(NSDictionary *)userInfo {
+    
     NSString *urlStr = [userInfo objectForKey:BRANCH_PUSH_NOTIFICATION_PAYLOAD_KEY];
 
-    if (urlStr.length) {
-        [[Branch getInstance] requestDeepLinkData:urlStr callback:^(NSDictionary *params, NSError *error) {
-            if (error == nil) {
-                if (params != nil) {
-                    [[BranchLogger shared] logDebug:[NSString stringWithFormat:@"Deep Link Params: %@", params] error:nil];
-                }
-            } else {
-                [[BranchLogger shared] logDebug:[NSString stringWithFormat:@"requestDeepLinkData failed with error: %@", error] error:error];
+    if (!urlStr.length) return;
+    
+    [[Branch getInstance] requestDeepLinkData:urlStr callback:^(NSDictionary *params, NSError *error) {
+        if (error == nil) {
+            if (params != nil) {
+                [[BranchLogger shared] logDebug:[NSString stringWithFormat:@"Deep Link Params: %@", params] error:nil];
             }
-        }];
-    }
+        } else {
+            [[BranchLogger shared] logDebug:[NSString stringWithFormat:@"requestDeepLinkData failed with error: %@", error] error:error];
+        }
+    }];
 }
 
 #pragma mark - Branch SceneDelegate Deep Linking Convenience Methods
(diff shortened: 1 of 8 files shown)


Sources/BranchSDK/Branch.m
```diff
@@ -2084,8 +2125,40 @@ - (void)requestDeepLinkDataWithSceneOptions:(nullable UISceneConnectionOptions *
...
+            } else {
+                [[BranchLogger shared] logDebug:[NSString stringWithFormat:@"requestDeepLinkData failed with error: %@", error] error:error];
+            }
+        }];
+    }
+}
+
+// Called from scene:continueUserActivity:
+- (void)requestDeepLinkDataWithScene:(UIScene *)scene continueUserActivity:(NSUserActivity *)userActivity {
+    [[Branch getInstance] requestDeepLinkData:userActivity.webpageURL.absoluteString callback:^(NSDictionary *params, NSError *error) {
```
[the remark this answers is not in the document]
bboothe-branch: Added handling fix

Sources/BranchSDK/Branch.m
```diff
@@ -2050,7 +2042,56 @@ - (void)requestDeepLinkDataWithLaunchOptions:(NSDictionary *)options
...
+        } else {
+            [[BranchLogger shared] logDebug:[NSString stringWithFormat:@"requestDeepLinkData failed with error: %@", error] error:error];
+        }
+    }];
+}
+
+// Called from application:continueUserActivity:restorationHandler:
+// Supports Universal Links
+- (void)requestDeepLinkDataWithContinueUserActivity:(NSUserActivity *)userActivity {
+    [[Branch getInstance] requestDeepLinkData:userActivity.webpageURL.absoluteString callback:^(NSDictionary *params, NSError *error) {
```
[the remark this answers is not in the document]
bboothe-branch: Added handling fix

Sources/BranchSDK/Branch.m
```diff
@@ -2050,7 +2042,56 @@ - (void)requestDeepLinkDataWithLaunchOptions:(NSDictionary *)options
     [self requestDeepLinkData:pushURL callback:callback];
 }
 
+// Called from application:openURL:options:
+// Supports URI Schemes
+- (void)requestDeepLinkDataWithOpenURL:(NSURL *)url {
+    [[Branch getInstance] requestDeepLinkData:[url absoluteString] callback:^(NSDictionary *params, NSError *error) {
```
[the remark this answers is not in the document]
bboothe-branch: Added handling fix

Sources/BranchSDK/Branch.m
```diff
@@ -2086,6 +2075,79 @@ - (void)requestDeepLinkDataWithSceneOptions:(nullable UISceneConnectionOptions *
...
+// AppDelegate Remote Notification deep linking
+// Used to support Push Notifications
+
+- (void)requestDeepLinkDataFromDidReceiveRemoteNotification:(NSDictionary *)userInfo {
+    NSString *urlStr = [userInfo objectForKey:BRANCH_PUSH_NOTIFICATION_PAYLOAD_KEY];
+    
+    if (urlStr.length) {
+        NSURL *url = [NSURL URLWithString:urlStr];
+        if (url)  {
+            [self requestDeepLinkData:url callback:^(NSDictionary *params, NSError *error) {
```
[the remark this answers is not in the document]
bboothe-branch: updated to urlStr

Sources/BranchSDK/Branch.m
```diff
@@ -2050,7 +2042,66 @@ - (void)requestDeepLinkDataWithLaunchOptions:(NSDictionary *)options
     [self requestDeepLinkData:pushURL callback:callback];
 }
 
+// Called from application:openURL:options:
+// Supports URI Schemes
+- (void)requestDeepLinkDataWithURL:(NSURL *)url {
```
gdeluna-branch, line 2067: function names should stay simple, the parameters give enough context.

`- (void)requestDeepLinkData:(NSURL *)url`
`- (void)requestDeepLinkData:(NSUserActivity *)userActivity`
`- (void)requestDeepLinkData:(NSDictionary *)userInfo`
etc..

Ensure that the parameters are well explained in the function comments.

gdeluna-branch (changes requested):

Sources/BranchSDK/Branch.m
bboothe-branch: Hey @gdeluna-branch I updated the descriptions to explain the parameters, but I can't write the method signatures like that. Obj-C doesn't check parameter types when declaring functions of the same name. It sees all three of those as the same function.

It works for things like `requestDeepLinkDataWithScene` because the `UIScene` parameter is required and adds another parameter to the declaration, so the full functions have `continueUserActivity` and `openURLContexts` in their use.

`- (void)requestDeepLinkDataWithScene:(UIScene *)scene continueUserActivity:(NSUserActivity *)userActivity`
`- (void)requestDeepLinkDataWithScene:(UIScene *)scene openURLContexts:(NSSet<UIOpenURLContext *> *)urlContexts`

When it's just one parameter the names have to be different unless we added a dummy value to them, but I don't think that's a good idea to add to our public APIs.

I did however change `requestDeepLinkDataFromRemoteNotification` to `requestDeepLinkDataWithUserInfo` as per this comments by Justin and Nidhi. I posted a comment in Slack and I can update if they disagree with this change.

NidhiDixit09:
@bboothe-branch 
* These convenience methods are bypassing many checks which were present in previous implementations. For example in `continueUserActivity` was parsing and initializing `initialReferrer` ,  depending upon the activity type, it was intializing `spotlightIdentifier` also, etc. 

* Also ensure checks in `- (BOOL)handleDeepLink: ...` are not skipped by new deep linking APIs.
* we should also add convenience methods for handling URI -- corresponding to original ```
- (BOOL)application:(UIApplication *)application
            openURL:(NSURL *)url
  sourceApplication:(NSString *)sourceApplication
         annotation:(id)annotation {
```. etc.

**Comment by bboothe-branch** [comment]
> @bboothe-branch
> 
> * These convenience methods are bypassing many checks which were present in previous implementations. For example in `continueUserActivity` was parsing and initializing `initialReferrer` ,  depending upon the activity type, it was intializing `spotlightIdentifier` also, etc.
> * Also ensure checks in `- (BOOL)handleDeepLink: ...` are not skipped by new deep linking APIs.
> * we should also add convenience methods for handling URI -- corresponding to original ```
> 
> * (BOOL)application:(UIApplication *)application
>   openURL:(NSURL *)url
>   sourceApplication:(NSString *)sourceApplication
>   annotation:(id)annotation {
> 
> ```
> 
> ```

Hey @NidhiDixit09 I added preprocessing functions that took the previous logic from `handleDeepLink` and `continueUserActivity` to process the preferenceHelper changes before the link is resolved with `requestDeepLinkData`.

`preprocessDeepLinkURL:sceneIdentifier:` uses the `handleDeepLink` processing
`preprocessUserActivity:sceneIdentifier:` uses the `continueUserActivity` processing

I figured we could remove handleDeepLink at some point so this would aid in that. I also added some tests to verify their function.

File: Sources/BranchSDK/Branch.m (hunk 5 above)
Comment (wpinho-branch) [reply Sources/BranchSDK/Branch.m:2130]: @bboothe-branch is right that these can't share the `requestDeepLinkData:` selector. Objective-C dispatches on the selector and not on the parameter type, so all three would collide.

There's a middle path if the short call site is what @gdeluna-branch is after: keep the explicit Objective-C names and add `NS_SWIFT_NAME` so Swift sees the clean ones. Something like `NS_SWIFT_NAME(requestDeepLinkData(launchOptions:callback:))` on the launch options variant gives Swift `requestDeepLinkData(launchOptions:)` while Objective-C keeps unambiguous selectors.

I haven't compiled that against the branch, so treat it as a direction rather than a patch. Worth checking how it interacts with the `NS_SWIFT_ASYNC_NAME` annotations coming in #1623, since both rename the same methods for Swift.

File: Sources/BranchSDK/Branch.m [hunk 6]
```diff
@@ -2077,7 +2085,94 @@ - (void)requestDeepLinkDataWithLaunchOptions:(NSDictionary *)options
...
+    NSString *urlStr = url.absoluteString;
+
+    if (!urlStr.length) return;
+
+    // Run the same preprocessing as the legacy handleDeepLink: path so referring-URL query params,
+    // the URL skiplist / dropURLOpen check, allowed-scheme filtering and link_click_id extraction are
+    // applied before the request is enqueued.
+    [self processDeepLinkURL:url sceneIdentifier:nil];
+
+    [[Branch getInstance] requestDeepLinkData:urlStr callback:^(NSDictionary *params, NSError *error) {
```
gdeluna-branch, line 2145: There may be a duplicate enqueue bug here. 

handleSchemeDeepLink_private: (line 791, reached via processDeepLinkURL: above) already calls requestDeepLinkData:nil. This line enqueues a second request with the URL. The second call cancels the first via cancelPendingDeepLinkRequests, but that only cancels requests not yet executing.

Sources/BranchSDK/Branch.m
```diff
@@ -2077,7 +2085,94 @@ - (void)requestDeepLinkDataWithLaunchOptions:(NSDictionary *)options
...
+    // The legacy application:openURL:sourceApplication:annotation: did not use them either.
+    [self requestDeepLinkDataWithURL:url];
+}
+
+// Called from application:continueUserActivity:restorationHandler:
+// Supports Universal Links
+- (void)requestDeepLinkDataWithUserActivity:(NSUserActivity *)userActivity {
+    NSString *urlStr = userActivity.webpageURL.absoluteString;
+
+    if (!urlStr) return;
```
gdeluna-branch: Looks like this early returns for Spotlight since it doesn't have webpageURL.

bboothe-branch:
@gdeluna-branch 
1. I added a filtered parameter that checks if a URL matches the skip list. If it does, requestDeepLinkData doesn't call. 
2. I moved the processUserActivity to run before webpageURL is read and removed the if(!url.str) so that it will always run. requestDeepLinkData   
  is still called with a nil value to recognize spotlight URLs that we set on preferenceHelper. 
3. I moved the requestDeepLinkData calls to the legacy functions to separate the process functions from enqueuing a request at all.

I also updated previous calls to `handleDeepLink` and `continueUserActivity` in `passPasteItemProviders` in **BranchScene** `scene:scene openURLContexts:URLContexts`. The **BranchScene** change still uses legacy functions since the `#if !TARGET_OS_TV` check is required for Scene. I wanted to remove these legacy functions but I don't think it's possible in this context unless I can use `requestDeepLinkDataWithScene:scene` without that check.

Sources/BranchSDK/Branch.m
bboothe-branch: Removed if(!urlStr) check and moved processing the userActivity to before the webpageURL check.

Sources/BranchSDK/Branch.m
bboothe-branch: Removed requestDeepLinkData from` handleSchemeDeepLink_private` (and `process_userActivity` since it had the same issue). Both only populate the `preferenceHelper` now.

requestDeepLinkData was moved up to handleDeepLink and continueUserActivity for the legacy APIs and requestDeepLinkData convenience methods have their own calls. So now they don't depend on cancelPendingDeepLinkRequests.

cancelPendingDeepLink requests is unchanged.

Sources/BranchSDK/Branch.m
bboothe-branch: Hey @wpinho-branch I really liked this suggestion. I added in some changes to this PR using `NS_SWIFT_NAME` and `NS_ASYNC_SWIFT_NAME`.

When looking at the original ideas, I thought that `requestDeepLinkData(url)` and `requestDeepLinkData(branchLink)` were too similar, so I modified them, along with the **openURLContext** functions to say **openURL** instead. I felt like developers would be confused as to which to use. 

I think this is a good compromise for what's possible with Swift vs Obj-C.

NidhiDixit09 (approved):

PR #1607, opened 2026-06-30
add Pact tests

hana-branch:
## Reference
SDK-XXXX -- <TITLE>.

## Type Of Change

## Testing Instructions

### DIFF
--- a/Branch-TestBed/Branch-TestBed.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Branch-TestBed/Branch-TestBed.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "bb99ce17c563a1c551806f1d8b2d31aef8e86c9f5237904897126d663a729026",
+  "originHash" : "bf3fe443d2349d81f7be27114be6380a364f2f6144593ceae9eac25dd8e8d72a",
   "pins" : [
     {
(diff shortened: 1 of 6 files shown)


Branch-TestBed/Branch-TestBed.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -36,6 +36,24 @@
         "revision" : "b7e1104502eca3a213b46303391ca4d3bc8ddec1",
         "version" : "2.30910.0"
       }
+    },
+    {
+      "identity" : "pactswift",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/surpher/PactSwift",
+      "state" : {
+        "branch" : "main",
```
wpinho-branch, line 45: PactSwift is pinned to a moving branch rather than a version. The server package right below it is pinned to a release, which is the shape you want. With `main`, a resolve six months from now doesn't build the same thing as today, and a test dependency that drifts on its own eventually fails in a way nobody can reproduce. Is there a tagged release that works?

wpinho-branch (commented):
Reviewing this as a draft, so treating it as direction rather than something ready. Contract tests against the Branch API are a good idea and I'd like to see this land.

Before it leaves draft: the description is still the template, and I couldn't trace the ticket from the branch name. Is this still active? It's been sitting since the end of June without a reviewer.

PR #1608, opened 2026-06-30
ci: remove gptdriver-release workflow, constrain layer1 to Xcode ^16.4 and gate the beta line

wpinho-branch:
## What

- **Remove `.github/workflows/gptdriver-release.yml`** — no longer needed.
- **Pin `layer1-logger-tests.yml` to exact Xcode `16.4`** (was floating `"16"`).

## Why

The `gptdriver-release` workflow pinned `xcode-version: "15.4"`, which GitHub has removed from the `macos-15`/`macos-latest` runner images. Available Xcode versions are now only 16.0–16.4 and 26.x, so any `Release-*` push failed at the "Set up Xcode" step (see [runner-images#14167](https://github.com/actions/runner-images/issues/14167)).

Since the workflow is being retired anyway, it is removed rather than rebumped.

For the remaining `layer1-logger-tests.yml` we move from a floating major (`"16"`, resolves to the highest 16.x available) to an exact GM (`"16.4"`). Exact pinning keeps the build deterministic and avoids silent drift across runner image revisions, while staying clear of Xcode 26 (where the `iPhone 15` simulator the suite uses no longer exists). The job already pins `runs-on: macos-15`.

Note: the iOS 12 minimum deployment target is independent of the compiler version — Xcode 16 builds for iOS 12 deployment targets without issue, so no downstream impact.

## Scope

CI configuration only. No SDK source changes.


.github/workflows/layer1-logger-tests.yml
```diff
@@ -47,10 +47,10 @@ jobs:
 
       # setup-xcode is more robust than `sudo xcode-select -s` because the
       # exact Xcode_16.x.app path varies between runner image revisions.
-      - name: Select Xcode 16
```
NidhiDixit09, line 50: @wpinho-branch Instead of hardcoding the version, can we use something like latest or latest-stable so that it does not break on next update.

Also for target branches, can we include Beta SDK also ? cc: @gdeluna-branch

wpinho-branch: Both done in 049f2530 and 5744f8f5.

Went with `^16.4` rather than `latest`/`latest-stable` — a floating version would walk onto Xcode 17 and change the toolchain under a wire-validation gate. `^16.4` still takes 16.x updates so it will not break when the image drops the exact build.

Added `4.0.0-beta.*` to the trigger list. That entry alone does not enable it: the workflow file has to exist in the PR merge tree and is currently absent on `4.0.0-beta.0`, so it needs a companion PR onto the beta line.

gdeluna-branch (approved):

PR #1609, opened 2026-07-09
Fixed crash for malformed or nil strings coming from pasteboard or sp…

NidhiDixit09:
…otlight.

## Reference
SDK-XXXX -- <TITLE>.

## Type Of Change

## Testing Instructions

### DIFF
--- a/Branch-TestBed/Branch-SDK-Tests/BranchClassTests.m
+++ b/Branch-TestBed/Branch-SDK-Tests/BranchClassTests.m
@@ -7,6 +7,7 @@
 //
 
 #import <XCTest/XCTest.h>
+#import <CoreSpotlight/CoreSpotlight.h>
 #import "Branch.h"
 #import "BranchConstants.h"
 #import "BNCPasteboard.h"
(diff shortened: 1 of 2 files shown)


Branch-TestBed/Branch-SDK-Tests/BranchClassTests.m
```diff
@@ -479,4 +480,51 @@ - (void)testHandleDeepLink_URISchemeWithBranchReferrer_ReturnsYes {
...
+}
+
+- (void)testIsBranchLink_BranchReferrerSubstringParam_ReturnsNo {
+    // "branch_referrer" is not "_branch_referrer".
+    XCTAssertFalse([Branch isBranchLink:@"https://example.com/products/42?branch_referrer=H4sIAAAA"]);
+}
+
+- (void)testContinueUserActivity_MissingSpotlightIdentifier_DoesNotThrow {
+    NSUserActivity *activity = [[NSUserActivity alloc] initWithActivityType:@"io.branch.testActivity"];
+    activity.userInfo = @{};
```
gdeluna-branch, line 526: What happens if the dictionary is not empty but does not contain a string value? like
```activity.userInfo = @{CSSearchableItemActivityIdentifier: @10};```

Is this case possible?

NidhiDixit09: It will crash if its a number. But its only possible to be number if user manually creates it. iOS always passes String in this. This is pre-existing bug and app will crash even before hitting the "isBranchLink" function.
Also X-code will show warning - `Incompatible pointer types sending 'NSNumber *' to parameter of type 'NSString * _Nonnull'` @gdeluna-branch

gdeluna-branch (approved):

bboothe-branch (approved):

PR #1612, opened 2026-07-15
EMT-3892: stop the duplicate open in the deep-link open lifecycle

wpinho-branch:
Fixes the duplicate `/v3/events/open` on a Branch deep-link launch. The launch open is deferred until `requestDeepLinkData` resolves, so `BranchRequestDeepLink.processResponse` sends a single attributed open, instead of one unattributed open firing immediately and a second attributed one landing after resolution. A bounded fallback (`preferenceHelper.timeout`) still sends the open unattributed if the host app never calls `requestDeepLinkData`, so it is never lost.

## Scope change

This was opened as EMT-3892 + EMT-3893. **EMT-3893 (link stickiness) is no longer part of the fix.** `2bd5e040` on `4.0.0-beta.0` already clears the link identifiers on the attributed open path — `BranchRequestOpen.processResponse` nils `linkClickIdentifier`, `spotlightIdentifier`, `universalLinkUrl`, `externalIntentURI` and sets `referringURL` from the response, which is nil when no link came back.

What this PR still contributes for 3893 is a **regression test** guarding that behaviour, since `2bd5e040` landed without one. No production change.

## Ordering fix after catching up with beta

The guard marked `lastAttributedDeepLinkURL` by reading `preferenceHelper.referringURL` back inside `sendOpen:skipCallback:`. That worked while `BranchRequestDeepLink` wrote `referringURL` into the preference helper, but `a03b72a4` made it a local variable. In the resolution-first ordering the mark then read nil, the guard in `handleUniversalDeepLink_private:` never matched, the fallback re-armed, and a second open went out — the exact duplicate this ticket exists to fix.

`BranchRequestDeepLink` now passes the resolved link to `markDeepLinkAttributed:` explicitly. The read in `sendOpen:` is kept for the reverse ordering (universal link first, which does populate `referringURL`) but no longer overwrites an existing mark with nil.

## Verification

`xcodebuild test -project BranchSDK.xcodeproj -scheme BranchSDKTests` — **491 tests, 0 failures.**

The suite aborts intermittently on this machine before completing, with WebKit/simulator process churn in the log; the QR-code tests reach `api3.branch.io` over the real network. Every run that ran to completion reported 491/0, and no run recorded a test failure. Pre-existing, not introduced here.

Synced with `4.0.0-beta.0` by merge, per the convention on this line. No force push, so the review threads above are intact.

### DIFF
--- a/Sources/BranchSDK/Branch.m
+++ b/Sources/BranchSDK/Branch.m
@@ -161,6 +161,13 @@ typedef NS_ENUM(NSInteger, BNCInitStatus) {
 @property (assign, nonatomic) BOOL deepLinkOpenPending;
 // Bounded fallback timer that fires the deferred open (unattributed) if resolution never happens.
 @property (strong, nonatomic) dispatch_source_t deepLinkOpenFallbackTimer;
+// EMT-3892 ordering fix: the link URL an ATTRIBUTED open was last sent for. Lets
+// handleUniversalDeepLink_private: recognize "this link was already resolved and
+// attributed" when requestDeepLinkData wins a race and resolves it first — otherwise the
+// late-arriving handleUniversalDeepLink_private: call would re-arm the deferred-open
+// fallback and emit a second, unattributed open for the same link. Cleared alongside
+// referringURL once the attributed open's own session settles (sendOpenNotificationWithLinkParameters:).
+@property (copy, nonatomic) NSString *lastAttributedDeepLinkURL;
 @property (assign, nonatomic) BOOL shouldAutomaticallyDeepLink;
 @property (strong, nonatomic) BNCLinkCache *linkCache;
 @property (strong, nonatomic) BNCPreferenceHelper *preferenceHelper;
@@ -1003,6 +1010,16 @@ static NSString *bnc_branchKey = nil;
     }
 
     // [self initUserSessionAndCallCallback:YES sceneIdentifier:sceneIdentifier urlString:urlString reset:YES];
+    // EMT-3892 ordering fix: requestDeepLinkData can win the race and resolve (and attribute)
+    // this exact link before this method runs (e.g. UIKit delivering continueUserActivity:
+    // after a scene-connection-options-driven resolution). If so, the single attributed open
+    // has already been sent — don't re-arm the deferred-open fallback, or it would fire a
+    // second, unattributed open for the same link.
+    if (urlString.length && [self.lastAttributedDeepLinkURL isEqualToString:urlString]) {
+        [[BranchLogger shared] logVerbose:[NSString stringWithFormat:@"handleUniversalDeepLink_private: %@ was already attributed via an earlier resolution; not re-arming the deferred open.", urlString] error:nil];
+        return [Branch isBranchLink:urlString];
+    }
+
     // EMT-3892: defer the launch open until requestDeepLinkData resolves this link (see
     // deferLaunchOpenPendingDeepLinkResolution) instead of firing an immediate, unattributed
     // sendOpen here — which previously raced with the attributed sendOpen: fired after
@@ -2425,6 +2442,9 @@ static inline void BNCPerformBlockOnMainThreadSync(dispatch_block_t block) {
         userInfo:userInfo];
 
     self.preferenceHelper.referringURL = nil;
+    // EMT-3892 ordering fix: the session this attribution belonged to has now settled;
+    // a later resolution of the same link URL should be treated as fresh again.
+    self.lastAttributedDeepLinkURL = nil;
 }
 
 - (void)removeViewControllerFromRootNavigationController:(UIViewController*)branchSharingController {
@@ -2799,6 +2819,12 @@ static inline void BNCPerformBlockOnMainThreadSync(dispatch_block_t block) {
         [[BranchLogger shared] logDebug: @"Branch Attribution Level set to NONE. Branch sendOpen network request prevented." error:nil];
         return;
     }
+
+    // EMT-3892 ordering fix: remember which link this ATTRIBUTED open was for, so a
+    // handleUniversalDeepLink_private: call arriving afterward for the SAME link (already
+    // resolved here) does not re-arm the deferred-open fallback. See
+    // handleUniversalDeepLink_private: and lastAttributedDeepLinkURL.
+    self.lastAttributedDeepLinkURL = self.preferenceHelper.referringURL;
 
     // Prepare callback block
     callbackWithStatus openCallback = ^(BOOL success, NSError *error) {
--- a/Sources/BranchSDK/Branch.m
+++ b/Sources/BranchSDK/Branch.m
@@ -155,6 +155,12 @@ typedef NS_ENUM(NSInteger, BNCInitStatus) {
 - (void)notifyWhenSessionReady:(void (^)(NSError * _Nullable error))block;
 - (void)deliverSessionReadyBlocksWithError:(nullable NSError *)error;
 - (void)discardSessionReadyBlocks;
+
+// EMT-3892: true while a Branch deep-link open is deferred, waiting for requestDeepLinkData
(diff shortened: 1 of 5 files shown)


Sources/BranchSDK/Branch.m
```diff
@@ -997,7 +1010,21 @@ - (BOOL)handleUniversalDeepLink_private:(NSString*)urlString sceneIdentifier:(NS
...
 
     // [self initUserSessionAndCallCallback:YES sceneIdentifier:sceneIdentifier urlString:urlString reset:YES];
-    [self sendOpen];
+    // EMT-3892 ordering fix: requestDeepLinkData can win the race and resolve (and attribute)
+    // this exact link before this method runs (e.g. UIKit delivering continueUserActivity:
+    // after a scene-connection-options-driven resolution). If so, the single attributed open
+    // has already been sent — don't re-arm the deferred-open fallback, or it would fire a
+    // second, unattributed open for the same link.
+    if (urlString.length && [self.lastAttributedDeepLinkURL isEqualToString:urlString]) {
+        [[BranchLogger shared] logVerbose:[NSString stringWithFormat:@"handleUniversalDeepLink_private: %@ was already attributed via an earlier resolution; not re-arming the deferred open.", urlString] error:nil];
```
gdeluna-branch, line 1010: Can you make the comments in this change to be more concise? it's difficult to read the code

wpinho-branch: Condensed in afffc417. Comments only, no code changed.

Sources/BranchSDK/Branch.m
```diff
@@ -1993,9 +2009,11 @@ - (void)applicationDidBecomeActive {
         
         [[BranchLogger shared] logVerbose:[NSString stringWithFormat:@"applicationDidBecomeActive installOrOpenInQueue %d", installOrOpenInQueue] error:nil];
 
-        if (!Branch.trackingDisabled && self.initializationStatus == BNCInitStatusUninitialized && !installOrOpenInQueue) {
-            [[BranchLogger shared] logVerbose:[NSString stringWithFormat:@"applicationDidBecomeActive trackingDisabled %d initializationStatus %d installOrOpenInQueue %d", Branch.trackingDisabled, self.initializationStatus, installOrOpenInQueue] error:nil];
-            
+        // EMT-3892: a deferred deep-link open is still pending; firing here would be a
+        // second, unattributed open.
+        if (!Branch.trackingDisabled && self.initializationStatus == BNCInitStatusUninitialized && !installOrOpenInQueue && !self.deepLinkOpenPending) {
```
bboothe-branch, line 2014:
```suggestion
        if (![Branch attributionLevelNone] && !installOrOpenInQueue && !self.deepLinkOpenPending) {
```

Instead of trackingDisabled lets check for BranchAttributionLevelNone since we're removing trackingDisabled. Let's also remove initializationStatus check since that initialization states will be removed.

Sources/BranchSDK/Branch.m
```diff
@@ -1993,9 +2009,11 @@ - (void)applicationDidBecomeActive {
...
+            [[BranchLogger shared] logVerbose:[NSString stringWithFormat:@"applicationDidBecomeActive trackingDisabled %d initializationStatus %d installOrOpenInQueue %d deepLinkOpenPending %d", Branch.trackingDisabled, self.initializationStatus, installOrOpenInQueue, self.deepLinkOpenPending] error:nil];
```
bboothe-branch, line 2015:
```suggestion
            [[BranchLogger shared] logVerbose:[NSString stringWithFormat:@"applicationDidBecomeActive attributionLevelNone %d installOrOpenInQueue %d deepLinkOpenPending %d", [Branch attributionLevelNone], installOrOpenInQueue, self.deepLinkOpenPending] error:nil];
```

Log attributionLevelNone check

bboothe-branch (commented):
Code looks good just added some changes to checks

Sources/BranchSDK/Branch.m
wpinho-branch: Agreed on `attributionLevelNone`, and it is more urgent than a style call: `trackingDisabled` has zero occurrences under `Sources/BranchSDK` on `4.0.0-beta.0` as of 8cb14a29, so this guard would not compile once the stack lands there.

Keeping the `initializationStatus` check though — beta still defines `BNCInitStatus` and still uses it in this same guard (`Branch.m:1969`). Will drop it with #1603, not before.

It cannot be applied as a line edit here: `attributionLevelNone` does not exist on this branch or its base yet. The base (`fix/beta-auto-open-session-ready`, #1600) is 17 commits behind beta, so this is a rebase.

Sources/BranchSDK/Branch.m
wpinho-branch: Taken with the guard change above — the log moves to `attributionLevelNone` in the same edit.

Sources/BranchSDK/Branch.m
wpinho-branch: Done in 88f7b271 — merged `4.0.0-beta.0` in rather than rebasing, so these threads survive. The guard now reads:

```objc
if (![Branch attributionLevelNone] && self.initializationStatus == BNCInitStatusUninitialized && !installOrOpenInQueue && !self.deepLinkOpenPending) {
```

Kept `initializationStatus` for now, since beta still uses it in this same guard and #1603 has not landed. Ready to drop it the moment it does.

Catching up with beta also surfaced a real defect the old base was hiding: `a03b72a4` stopped `BranchRequestDeepLink` from writing `preferenceHelper.referringURL`, which is what the ordering guard read to mark the attributed link. Two of the EMT-3892 tests failed on top of current beta — a second open really was going out. Fixed in a0888a59 by passing the link explicitly. 491 tests, 0 failures.

Sources/BranchSDK/Branch.m
wpinho-branch: Log updated with the guard in 88f7b271 — now prints `attributionLevelNone` and `deepLinkOpenPending`.

Sources/BranchSDK/Branch.m
wpinho-branch: Missed the initializationStatus half of this the first time round. Both are in e5f92843: the guard is down to attributionLevelNone, installOrOpenInQueue and deepLinkOpenPending, and the verbose log alongside it drops the same field.

Sources/BranchSDK/Branch.m
```diff
@@ -2651,6 +2676,71 @@ - (void)discardSessionReadyBlocks {
...
+}
+
+// Defers the launch open so BranchRequestDeepLink.processResponse sends the single
+// attributed open. Bounded fallback (preferenceHelper.timeout, default 5.5s) fires an
+// unattributed open if the host app never calls requestDeepLinkData. EMT-3892.
+- (void)deferLaunchOpenPendingDeepLinkResolution {
+    self.deepLinkOpenPending = YES;
+
+    @synchronized (self) {
+        [self cancelDeepLinkOpenFallbackTimerLocked];
```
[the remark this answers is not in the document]
wpinho-branch: Fixed in e5f92843. The assignment moved inside the `@synchronized (self)` block that already followed it, so the flag is only touched under the same monitor the timer handler uses.

Sources/BranchSDK/Branch.m
```diff
@@ -2651,6 +2676,71 @@ - (void)discardSessionReadyBlocks {
...
+    }
+}
+
+// Called once resolution is underway, so the fallback does not fire a redundant open.
+- (void)cancelDeepLinkOpenFallback {
+    self.deepLinkOpenPending = NO;
+    @synchronized (self) {
+        [self cancelDeepLinkOpenFallbackTimerLocked];
+    }
+}
```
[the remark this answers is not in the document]
wpinho-branch: Fixed in e5f92843, same change on this side: the assignment moved inside the `@synchronized (self)` block.

bboothe-branch (approved):

PR #1614, opened 2026-07-21
EMT-3977 - Adding BranchConfiguration class and Initialize method

bboothe-branch:
## Reference
EMT-3977 -- Adding BranchConfiguration class and Initialize method

## Summary

BranchConfiguration.m and BranchConfiguration.h added to the Branch SDK. Now pre-init settings are added to a "configuration" object to pass into the new SDK initialization function.

Old SDK functions remain as is and the new Branch Configuration class functions act as wrappers on old deprecated functions before a full transition to BranchConfiguration using those functions.

## Type Of Change

## Testing Instructions

1. Open app
2. View BranchConfiguration example in AppDelegate.m 
3. Run the app, set different attributes in the Branch Configuration class.
4. Send a Branch Event to view changes in SDK performance
5. View changes in SDK


Branch-TestBed/Branch-TestBed/AppDelegate.m
```diff
@@ -19,21 +19,88 @@ @implementation AppDelegate
...
+    // NOTE: iOS uses SECONDS (NSTimeInterval), unlike Android's milliseconds.
+    config.networkTimeout        = 10.0;                 // 10s   (Android 10_000ms)
+    config.retryCount            = 5;
+    config.retryInterval         = 1.0;                  // 1s    (Android 1_000ms)
+    config.thirdPartyAPIsWaitTime = 2.0;                 // wait up to 2s for ODM / Apple attribution token
+    // config.remoteInterface = [MyCustomNetworkService class]; // must conform to BNCNetworkServiceProtocol
+
+    // Privacy & attribution  (defaults: attributionLevel nil, limitFacebookAttribution NO,
+    //                          adNetworkCalloutsDisabled NO, DMA params unset)
+    config.attributionLevel = BranchAttributionLevelFull;
```
NidhiDixit09, line 76: @bboothe-branch This will send an extra open, even before deep linking call. Is this fine / expected behavior ?  [cc: @gdeluna-branch ]

Sources/BranchSDK/Branch.m
```diff
@@ -200,6 +201,152 @@ + (Branch *)getInstance:(NSString *)branchKey {
...
+// Applies every configuration value that depends on the singleton already existing. Settings that must
+// precede singleton creation (test key, network-service class, branch key) are handled inline in
+// +initialize: before the instance is created and are intentionally not repeated here.
+// This mirrors the Android SDK's `applyConfiguration(Branch, BranchConfiguration)`.
++ (void)applyConfiguration:(BranchConfiguration *)configuration toBranch:(Branch *)branch {
+    // Logging: the caller's logLevel/callback own the logger state, overriding any branch.json toggle.
+    if (configuration.loggingCallback) {
+        [Branch enableLoggingAtLevel:configuration.logLevel withCallback:configuration.loggingCallback];
+    } else {
+        BranchLogger *logger = [BranchLogger shared];
```
NidhiDixit09, line 270: @bboothe-branch This will always enable logging.

Sources/BranchSDK/BranchConstants.m
```diff
@@ -190,6 +190,7 @@
 NSString * const BRANCH_KEY_SOURCE_INFO_PLIST = @"info_plist";
 NSString * const BRANCH_KEY_SOURCE_GET_INSTANCE_API = @"getInstance_api";
```
NidhiDixit09, line 191: @bboothe-branch BRANCH_KEY_SOURCE_GET_INSTANCE_API is not used anymore. So it should be deleted.

wpinho-branch:
`BranchConfiguration` here exposes `networkTimeout`, `retryCount` and `retryInterval`. The Android builder in https://github.com/BranchMetrics/android-branch-deep-linking-attribution/pull/1381 exposes those three plus `networkConnectTimeout` and `noConnectionRetryMax`. Since this PR already mirrors Android deliberately, flagging the two that did not come across — they are different cases.

`networkConnectTimeout`: iOS already has the dimension. `BNCNetworkService.m:122-123` sets `timeoutIntervalForRequest` and `timeoutIntervalForResource` from the same `defaultTimeoutInterval`, so the two are collapsed onto one knob rather than missing. Exposing them separately is a config decision, not new capability.

`noConnectionRetryMax`: I found no iOS equivalent in `Sources/`, so this one looks Android-only by capability.

Worth settling either way while the PR is open — both are additive to a public config object today and a public API change after merge.

Branch-TestBed/Branch-TestBed/AppDelegate.m
bboothe-branch: Hey @NidhiDixit09 I am going to update this to add resetSession NO to this.

Sources/BranchSDK/Branch.m
bboothe-branch: Hey @NidhiDixit09 I will update this to an `else if` that checks for log level too.

Sources/BranchSDK/BranchConstants.m
bboothe-branch: Hey @NidhiDixit09 I can remove this in this later PR since it's removing getInstance. I think It makes it more clear why it's no longer used. https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/pull/1616

bboothe-branch:
> `BranchConfiguration` here exposes `networkTimeout`, `retryCount` and `retryInterval`. The Android builder in [BranchMetrics/android-branch-deep-linking-attribution#1381](https://github.com/BranchMetrics/android-branch-deep-linking-attribution/pull/1381) exposes those three plus `networkConnectTimeout` and `noConnectionRetryMax`. Since this PR already mirrors Android deliberately, flagging the two that did not come across — they are different cases.
> 
> `networkConnectTimeout`: iOS already has the dimension. `BNCNetworkService.m:122-123` sets `timeoutIntervalForRequest` and `timeoutIntervalForResource` from the same `defaultTimeoutInterval`, so the two are collapsed onto one knob rather than missing. Exposing them separately is a config decision, not new capability.
> 
> `noConnectionRetryMax`: I found no iOS equivalent in `Sources/`, so this one looks Android-only by capability.
> 
> Worth settling either way while the PR is open — both are additive to a public config object today and a public API change after merge.

Hey @wpinho-branch 

After reviewing Android's networkConnectTimeout, I found it is a fundamentally different function from iOS's defaultTimeoutInterval. 

Android has two distinct network phases: 
setConnectTimeout, which sets how long establishing the connection may take 
networkTimeout (setReadTimeout), which sets waiting time to read data. Both read through PrefHelper. 

iOS has no connection-establishment timeout. NSURLSession exposes only:
timeoutIntervalForRequest (an idle timeout — max time with no new data received)
timeoutIntervalForResource (total time budget for the whole resource load). 

Both are session-wide, and neither sets connection establishment the way Android's setConnectTimeout does.

Separately, iOS already supports the networkTimeout field, which governs the per-request timeout customers already set. So the user-facing timeout is already covered.

Since adding this wouldn’t be true parity, I decided not to add it.
[findings.md](https://github.com/user-attachments/files/30669738/findings.md)

Sources/BranchSDK/Branch.m
wpinho-branch: Following up here since it's the same `else if`.

The guard fixed the always-on case, but it inverted things for Verbose. `BranchLogLevelVerbose` is the first case of the enum in BranchLogger.h, so it's zero, and the check reads zero as false. I ran the three cases against c8aad1ab: with `[BranchConfiguration debug:]` logging comes back disabled, while the default Error level and an explicit Debug both enable it. So `debug:`, the preset that exists to give you verbose output, is the one that produces nothing.

The other half of that is worth a decision too. The default level is Error, which is non-zero, so `initialize:` now enables logging for every app that calls it. That used to take an explicit `[Branch enableLogging]`.

Since `logLevel` is a non-optional enum, testing its truthiness can't tell "not set" apart from "Verbose". A separate flag on the config, or comparing against the default, would cover both.

Sources/BranchSDK/Public/BranchConfiguration.h
```diff
@@ -0,0 +1,187 @@
...
+/// `+[Branch initialize:]` is called; to change consent at runtime, reinitialize with a new configuration.
+@property (nonatomic, copy, nullable) BranchDMAParameters *dmaParameters;
+
+#pragma mark - URL collection
+
+/// URI schemes allowed to be tracked by Branch. Empty means all schemes (the default).
+@property (nonatomic, copy, readonly) NSArray<NSString *> *allowedSchemes;
+
+/// Regex patterns matching URLs that Branch should never transmit (e.g. URLs with sensitive tokens).
+@property (nonatomic, copy, readonly) NSArray<NSString *> *urlPatternsToIgnore;
```
wpinho-branch: There's no way to set this from outside. It's readonly here and there's no `addUrlPatternToIgnore:` next to `addAllowedScheme:` and `addMetadataWithKey:` further down, so a caller can't reach it. The implementation already has the setter, it just never got declared. EMT-3977 lists it as a deliverable, mapped to `mutableUrlPatternsToIgnore`.

```suggestion
@property (nonatomic, copy) NSArray<NSString *> *urlPatternsToIgnore;
```

While you're in here: `apiUrl` and `safeTrackAPIUrl` are both documented as needing an `http://` or `https://` prefix, but `validate` only checks that the key isn't empty and that the numbers are in range. Either check the scheme or soften the doc.

Sources/BranchSDK/Branch.m
```diff
@@ -200,6 +201,152 @@ + (Branch *)getInstance:(NSString *)branchKey {
...
+    // Test key must be resolved before the branch key is read.
+    [Branch setUseTestBranchKey:configuration.testMode];
+
+    // Custom network service class is set-once and must precede singleton creation.
+    if (configuration.remoteInterface) {
+        [Branch setNetworkServiceClass:configuration.remoteInterface];
+    }
+
+    // Set the branch key explicitly so it takes precedence over Info.plist / branch.json.
+    self.branchKey = configuration.branchKey;
```
wpinho-branch: The guard closes a few lines above, but the key only gets set here. A second thread calling `initialize:` in that window takes the early return and asks for the singleton with `self.class.branchKey`, which hasn't been set from the configuration yet, so the getter falls back to branch.json or Info.plist.

If that happens first, `setBranchKey:` on this line finds a different key already in place, logs that the key can only be set once, and the one from the configuration is quietly dropped.

I worked this out from reading the ordering rather than reproducing it, and a two-thread race isn't something I can make fire on demand. Is `initialize:` meant to be called only from `application:didFinishLaunchingWithOptions:` on the main thread? If so, saying that in the header doc settles it. If not, holding the lock through the singleton creation would.

Sources/BranchSDK/Public/BranchDMAParameters.h
```diff
@@ -0,0 +1,54 @@
+//
```
wpinho-branch, line 1: Confirming this header ships with the PR, since the three umbrella imports depend on it. Both it and its `.m` are in the tree at this commit, and I built the branch and ran the unit target on a simulator with everything passing, so all three umbrellas resolve.

wpinho-branch (changes requested):
The config surface is a real improvement over the scattered setters. Two things before it lands: the `else if` on the logging path inverts for Verbose, which I followed up on in the existing thread, and `urlPatternsToIgnore` can't be set from the public header.

One in passing: `resetInitializationGuardForTesting` lives in `Branch.m` rather than the test target, so it ships in the release binary.

I built this branch and ran the unit target on a simulator and everything passed. The QR cases that fail in CI pass locally, so that red looks like network reachability rather than anything here.

Sources/BranchSDK/Public/BranchConfiguration.h
bboothe-branch: I updated the `urlPatternsToIgnore `field by removing readonly and also added checkers for the prefix in `apiUrl `and `safeTrackAPIUrl`. The API url checks are a good catch because the setter functions have this check but it's better it fails directly from the `BranchConfiguration` instead

Sources/BranchSDK/Branch.m
bboothe-branch: Thanks for this Willian. Added an update that holds the lock through singleton creation for more flexibility.

Sources/BranchSDK/Branch.m
NidhiDixit09: +1 for Willian's comment.

Sources/BranchSDK/Branch.m
```diff
@@ -190,6 +191,147 @@ + (Branch *)getInstance:(NSString *)branchKey {
...
+// again. Not declared in the public header; exposed to tests via a category.
++ (void)resetInitializationGuardForTesting {
+    @synchronized ([Branch class]) {
+        bnc_didInitializeWithConfiguration = NO;
+    }
+}
+
++ (Branch *)initialize:(BranchConfiguration *)configuration {
+    if (!configuration) {
+        [NSException raise:NSInvalidArgumentException
```
NidhiDixit09: In current design of Branch SDK, it reports failure through NSError. This is the first ever PR which raises an exception in SDK. Please check with team if we should use exceptions like this.
Also if these are not handled or called via swift code, they can result in App crashes.

Sources/BranchSDK/Branch.m
```diff
@@ -190,6 +191,147 @@ + (Branch *)getInstance:(NSString *)branchKey {
...
+
+        self.branchKey = configuration.branchKey;
+        [BranchConfigurationController sharedInstance].branchKeySource = BRANCH_KEY_SOURCE_INIT_FUNCTION;
+
+        // --- Create (or fetch) the singleton ---
+        branch = [Branch getInstanceInternal:self.branchKey];
+    }
+
+    // --- Settings applied to the instance / shared preference helper (caller wins) ---
+    [Branch applyConfiguration:configuration toBranch:branch];
```
NidhiDixit09: This is outside syncronized block, so if second initalize is called at the same time, while the first one is applying the configuration, second one will return `[Branch getInstanceInternal:self.class.branchKey];` without applying the rest of the settings in following code.

Branch-TestBed/Branch-TestBed/AppDelegate.m
```diff
@@ -24,50 +30,89 @@ @implementation AppDelegate
...
+    // Compliance Branch Init example
+    // BranchConfiguration *config = [BranchConfiguration compliance:@"key_live_xxx"];
+
+    // Example: full BranchConfiguration setup, with every settable field changed from its default.
+    // This is the iOS counterpart of the Android `BranchConfiguration.Builder` chain. Objective-C has no
+    // chained builder, so create the config with the key, set the properties you care about, then pass it
+    // to `+[Branch initialize:]`.
+
+    // ── Build the configuration ──────────────────────────────────────────
+    BranchConfiguration *config =
```
NidhiDixit09, line 65: Gaurd this code with MACROs. Certain settings in this configuration can cause the Integration tests to fail.  This app itself will not work as expected.  For example - `config.euEndpoint      = YES;    `

Sources/BranchSDK/Branch.m
```diff
@@ -190,6 +191,147 @@ + (Branch *)getInstance:(NSString *)branchKey {
...
+        logger.logLevelThreshold = configuration.logLevel;
+    }
+
+    if (configuration.requestTracingCallback) {
+        [Branch setCallbackForTracingRequests:configuration.requestTracingCallback];
+    }
+
+    // Identity & environment. These mutate the BNCServerAPI / BNCPreferenceHelper singletons, whose
+    // values are read lazily at request time, so they don't need to precede singleton creation.
+    if (configuration.euEndpoint) {
```
NidhiDixit09: euEndpoint = NO cannot turn EU routing off if something else set it. Same for every "only if YES" boolean.

Sources/BranchSDK/BranchConfiguration.m
```diff
@@ -0,0 +1,177 @@
...
+
+- (void)setUrlPatternsToIgnore:(NSArray<NSString *> *)urlPatternsToIgnore {
+    self.mutableUrlPatternsToIgnore = [urlPatternsToIgnore mutableCopy] ?: [NSMutableArray array];
+}
+
+#pragma mark - Validation
+
+- (void)validate {
+    if (self.branchKey.length == 0) {
+        [NSException raise:NSInvalidArgumentException
```
NidhiDixit09: In current design of Branch SDK, it reports failure through NSError. This is the first ever PR which raises an exception in SDK. Please check with team if we should use exceptions like this.
Also if these are not handled or called via swift code, they can result in App crashes.
This comment applies to all the following code rasing exceptions.

Sources/BranchSDK/Branch.m
bboothe-branch: Added a private flag for if this with a logLevelWasSet field to distinguish default from a set value.

Sources/BranchSDK/Branch.m
bboothe-branch: Modified to use NSError.

Sources/BranchSDK/BranchConfiguration.m
bboothe-branch: Modified to use NSError

Sources/BranchSDK/Branch.m
bboothe-branch: Moved within the synchronized block

Branch-TestBed/Branch-TestBed/AppDelegate.m
bboothe-branch: Added Macro guards

Sources/BranchSDK/Branch.m
bboothe-branch: Added a setEUEndpoint to catch this and turn off routing if set elsewhere.

bboothe-branch:
@NidhiDixit09 made all requested updates. If this is merged will pull through to all other Branches and resolve conflicts once those are approved.

Sources/BranchSDK/BranchConfiguration.m
```diff
@@ -0,0 +1,214 @@
...
+
++ (instancetype)compliance:(NSString *)branchKey {
+    BranchConfiguration *config = [[self alloc] initWithKey:branchKey];
+    config.attributionLevel = BranchAttributionLevelMinimal;
+    return config;
+}
+
+#pragma mark - Collection accessors
+
+- (NSArray<NSString *> *)allowedSchemes {
```
NidhiDixit09, line 132: Provide cusotm setter for this property also. Otherwise objective-c will create its own.

Sources/BranchSDK/BranchConfiguration.m
```diff
@@ -0,0 +1,214 @@
...
+    return [self.mutableAllowedSchemes copy];
+}
+
+- (NSArray<NSString *> *)urlPatternsToIgnore {
+    return [self.mutableUrlPatternsToIgnore copy];
+}
+
+- (NSDictionary<NSString *, NSString *> *)requestMetadata {
```
NidhiDixit09, line 140: Provide cusotm setter for this property also. Otherwise objective-c will create its own. 
Or may be you can make property readonly.

Sources/BranchSDK/Branch.m
```diff
@@ -190,6 +192,166 @@ + (Branch *)getInstance:(NSString *)branchKey {
...
+        NSString *errorMessage = @"A BranchConfiguration is required to initialize Branch.";
+        NSError *configurationError = [NSError branchErrorWithCode:BNCInvalidConfigurationError localizedMessage:errorMessage];
+        [[BranchLogger shared] logError:errorMessage error:configurationError];
+        [[BranchLogger shared] logError:@"Failed to initialize Branch" error:nil];
+        return nil;
+    }
+
+    // Logging is applied before validation: a rejected configuration is reported only through the
+    // logger, so the caller's logLevel/loggingCallback must be live before anything can be rejected.
+    [Branch applyLoggingConfiguration:configuration];
```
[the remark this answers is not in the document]
bboothe-branch: This was intentional so that a user can update these logging values after initial set.

NidhiDixit09 (approved):
@bboothe-branch Added two more comments. Other than that ... it looks good to me.

Sources/BranchSDK/BranchConfiguration.m
bboothe-branch: Added a custom setter

Sources/BranchSDK/BranchConfiguration.m
bboothe-branch: Added a custom setter

PR #1615, opened 2026-07-23
EMT-3982 - Add Deprecation Messages for Pre-Init Functions

bboothe-branch:
## Reference
EMT-3982 -- Add Deprecation Messages for Pre-Init Functions

## Summary

With the addition of the new BranchConfiguration classes, we no longer need these pre-init functions. Users will set the information on the BranchConfiguration class. Deprecating these for now until we remove in the future.

This PR adds deprecation messages for users to see that BranchConfiguration is a new class that they will need to set this information on going forward, and they should refrain from calling these directly from now on.

## Type Of Change

## Testing Instructions

1. Open Branch.h
2. CTRL + F and search "deprecated"
3. Pick any of the deprecated functions and type them into didFinishLaunchingWithOptions in AppDelegate.m
4. Confirm deprecation log message appearing


wpinho-branch (commented):
The deprecation messages themselves are clear and point at the right replacements. One thing to clean before merge.

`Branch-TestBed/Branch-TestBed/Main.storyboard` is in the diff, with the width and height constraints on the testbed controls replaced by empty `<constraints/>`. It looks like Xcode churn from opening the file rather than part of this work, and it will change how the testbed lays out. Could it come out of the PR?

The blanket `#pragma clang diagnostic ignored "-Wdeprecated-declarations"` around whole method bodies works, but it also silences future deprecated calls added inside those methods. Keeping a non-deprecated internal entry point and having the deprecated public one forward to it would keep the warning meaningful. Not blocking.

I did not run this branch.

bboothe-branch:
@wpinho-branch I updated the SDK to use a helper method for deprecated functions instead of wrapping all the methods in the the deprecated messages.

Also, I removed the constraint changes. They weren't Xcode churn. I added those to reduce some of the warning messages from Xcode but I agree they don't belong in this PR.

NidhiDixit09 (changes requested):
In Branch.m following use of following APIs is giving depcrecation warning - 

> Branch.m:305  'setNetworkTimeout:' is deprecated
> Branch.m:306  'setMaxRetries:' is deprecated
> Branch.m:307  'setRetryInterval:' is deprecated
> Branch.m:308  'setSDKWaitTimeForThirdPartyAPIs:' is deprecated
> Branch.m:328  'setAllowedSchemes:' is deprecated
> Branch.m:331  'setUrlPatternsToIgnore:' is deprecated
> Branch.m:341  'setAppClipAppGroup:' is deprecated
> Branch.m:346  'setDeepLinkDebugMode:' is deprecated

Wrap them in #pragma push / pop.

NidhiDixit09 (approved):

PR #1616, opened 2026-07-24
EMT-3987 - Removed Branch getInstance and replace with sharedInstance

bboothe-branch:
## Reference
EMT-3987 -- Removed Branch getInstance and replace with sharedInstance.

## Summary

Replaced all calls to Branch getInstance with sharedInstance.

## Type Of Change

## Testing Instructions


wpinho-branch (commented):
The rename looks complete. `getInstance` is gone from the public header and `getInstanceInternal:` stays private, which is what you'd want.

One structural note that isn't yours to fix here, just worth naming since it keeps coming up. A good chunk of the files in this diff come in pairs, the same test under `Branch-TestBed/Branch-SDK-Tests/` and again under `BranchSDKTests/`, edited together. The same thing shows up in #1619 and #1603. While both trees exist, every test change has to be made twice, and they'll drift the first time someone forgets one. Is there a plan to collapse them before 4.0 ships?

I didn't run the test target on this branch, so this is a read of the diff rather than a green run.

bboothe-branch:
@wpinho-branch we can connect with the team on collapsing them. I think it was like this before so not sure if there's a specific reason for it.

NidhiDixit09 (approved):

NidhiDixit09 (approved):

PR #1617, opened 2026-07-27
EMT-3808 -- Hold session-dependent requests until the session is established

wpinho-branch:
## Reference

EMT-3808 — follow-up to #1603, targeting `bboothe-branch/EMT-3808`.

Addresses @NidhiDixit09's review question on `Sources/BranchSDK/Branch.m`:

> If These APIs(v1/url, v2/event) are called before sending OPENs, these APIs will have missing `randomized_device_token`. Is that fine?

## What the investigation found

The payload itself is safe. `BNCServerRequestOperation.m:84` still drops any request that needs a session when `randomizedDeviceToken` or `randomizedBundleToken` is missing, so nothing goes out token-less. #1603 only removed `sessionID` from that condition.

The real gap is ordering, and it is a direct consequence of removing `initializationStatus`:

- `initializationStatus` was **in-memory** and reset every launch.
- `randomized_device_token` is **persisted** in `NSUserDefaults` (`BNCPreferenceHelper.m:33`).

So from the second launch onward, a `logEvent` or `generateShortUrl` called before the open is enqueued passes the token check (using the previous session's token) and reaches the server ahead of that session's open. `BNCServerRequestQueue.addInitDependencyIfNeeded` does not cover it, because it only adds a dependency when an init operation already exists — in this window `currentInitOperation` is `nil`.

That is an attribution-ordering problem, not a missing-field one.

## The change

`BNCServerRequestQueue` now holds requests that need a session until the session establishing request finishes (success or failure), then releases them in enqueue order. Specifics:

- `+[BNCServerRequestOperation requestRequiresSession:]` factors out the install/open/deep-link classification that `BNCServerRequestOperation.start` already had, so the queue and the operation cannot drift apart.
- A 15s timeout releases held requests if the host app never initializes a session. Held requests then hit the existing token check and fail with `BNCInitError`, exactly as they do today — a misintegrated app is not left hanging.
- `clearQueue` cancels held requests; `queueDepth` and `description` account for them.
- The drop log now names the fix ("Initialize the Branch session before calling this API"), recovering the guidance the removed guard used to give.

## Deliberate non-changes

- **v1/url still requires a session on iOS.** Android's `requestNeedsSession()` returns `false` for `ServerRequestCreateUrl` (`ServerRequestQueue.java:382-384`), so Android allows short-link creation without a session. iOS requires the device token for short URLs today, and loosening that is a separate product decision — flagging the divergence rather than resolving it here.
- The synchronous path (`BranchShortUrlSyncRequest`) bypasses the queue entirely and is untouched.

## Testing

`BranchSDKTests/BNCServerRequestQueueSessionWaitTests.m` — 9 tests, all passing on iPhone 16 (iOS 18.4):

- ordering: the open reaches the network before the request that needs the session, with tokens already persisted (the exact reported scenario)
- multiple held requests release in enqueue order
- session establishing requests are never held
- no holding once the session is established
- the timeout releases held requests when no session is ever started
- `queueDepth` counts held requests; `clearQueue` drops them
- classification for `BranchOpenRequest` / `BranchInstallRequest` / `BranchRequestOpen` / `BranchRequestDeepLink`

**Negative control:** with the hold disabled, 5 of these fail, including the ordering test — the tests capture the regression, not the implementation.

**Full suite:** the only failures are `BranchConfigurationControllerTests.testGetConfiguration` and `testInstallRequestParams`, which fail identically on `a98feb87` without this change. Verified pre-existing.

tvOS was not built locally (platform not installed on this machine); CI covers it.

## Temporary commit in this branch

The last commit (`Stop a slow user agent load from failing unrelated device info tests`) is cherry-picked from #1618 and is **not part of this change**. It is here only so `verify` can go green before review: without it this branch inherits the `testAdvertiserId` WKWebView timeout flake from its base, which failed three consecutive runs while every one of the 9 tests added here passed.

Drop it when rebasing, once #1618 has landed on `4.0.0-beta.0`.

## Type Of Change

### DIFF
--- a/Sources/BranchSDK/BNCServerRequestOperation.m
+++ b/Sources/BranchSDK/BNCServerRequestOperation.m
@@ -105,7 +105,7 @@
     }
     
     if ([self.request isKindOfClass:[BranchRequestDeepLink class]]) {
-        [BranchRequestDeepLink setWaitNeededForOpenResponseLock];
+        [BranchRequestDeepLink setWaitNeededForDeepLinkResponseLock];
     }
     
     if ([self.request isKindOfClass:[BranchRequestOpen class]]) {
--- a/Sources/BranchSDK/BNCServerRequestOperation.m
+++ b/Sources/BranchSDK/BNCServerRequestOperation.m
@@ -41,6 +41,17 @@
     return YES;
 }
 
++ (BOOL)requestRequiresSession:(BNCServerRequest *)request {
+    // BranchInstallRequest is a subclass of BranchOpenRequest, listed for clarity.
+    if ([request isKindOfClass:[BranchInstallRequest class]] ||
+        [request isKindOfClass:[BranchOpenRequest class]] ||
+        [request isKindOfClass:[BranchRequestOpen class]] ||
+        [request isKindOfClass:[BranchRequestDeepLink class]]) {
+        return NO;
+    }
+    return YES;
+}
+
 - (void)setExecuting:(BOOL)executing {
     [self willChangeValueForKey:@"isExecuting"];
     _executing = executing;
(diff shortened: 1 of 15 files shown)


Sources/BranchSDK/BNCServerRequestOperation.m
```diff
@@ -74,14 +85,12 @@ - (void)start {
...
-       // If we do not have a randomized bundle token, we should receive one from the service
-        // We will receive from callback in v3/deeplink
-    } else {
+    // Install, open and deep link requests establish the session, so they carry no session
+    // tokens of their own. If we do not have a randomized bundle token, we receive one from
+    // the service (via the v3/deeplink callback).
+    if ([[self class] requestRequiresSession:self.request]) {
         if (!preferenceHelper.randomizedDeviceToken || !preferenceHelper.randomizedBundleToken) {
-            [[BranchLogger shared] logError:[NSString stringWithFormat:@"Missing session items (device token or bundle token). Dropping request: %@", self.request.requestUUID] error:nil];
+            [[BranchLogger shared] logError:[NSString stringWithFormat:@"Missing session items (device token or bundle token). Dropping request: %@. Initialize the Branch session before calling this API.", self.request.requestUUID] error:nil];
```
bboothe-branch:
```suggestion
            [[BranchLogger shared] logError:[NSString stringWithFormat:@"Missing session items (device token or bundle token). Dropping request: %@. Initialize Branch before calling this API.", self.request.requestUUID] error:nil];
```

I think with us dropping Session verbiage we can change this to "Initialize Branch before calling this API." Can check with Nidhi on this.

bboothe-branch (approved):

bboothe-branch (approved):

bboothe-branch:
I pulled in some changes on 3808 prior should be good to merge after conflicts resolved

PR #1620, opened 2026-07-29
Create CLAUDE.md

wpinho-branch:
Adds repo context for Claude Code on `master`, matching what was done for Android in [Create CLAUDE.md](https://github.com/BranchMetrics/android-branch-deep-linking-attribution/pull/1384).

This is the prerequisite for the team adopting RPIC (Research → Plan → Implement → Code Review): with the context on disk, we stop paying tokens to re-derive the repo layout every session, and everyone prompts from the same starting point.

## What's in it

- **Products and distribution** — SPM / CocoaPods / Carthage / XCFramework built from one source tree; the `-noidfa` variants are the same sources under `BRANCH_EXCLUDE_IDFA_CODE=1`; version lives in both `BNCConfig.m` and the podspec and is only ever changed via `scripts/version.sh`.
- **Source layout** — `Public/` vs `Private/` header split (adding to `Public/` is a public API change), and the `BNC*` / `Branch*` naming convention.
- **Where to make changes** — a task → file table covering session init, request types, wire format, persisted state, queue behavior, transport, links, events, consent, SKAdNetwork, config, and diagnostics.
- **Build, test, lint** — the fastlane loop CI actually runs, the plain `xcodebuild` equivalent, and where the tests live (`Branch-TestBed/Branch-SDK-Tests/`, not next to `Sources/`). Notes that `Branch-TestBed-CI.xctestplan` runs only the `Branch-SDK-Tests` target, so a new test target that isn't added to the plan will never run in CI.
- **CI** — `verify.yml`, the Layer-1 wire validation gate and how to reproduce it locally, pre/post-release QA, release.
- **Architecture** — singleton wiring, the isolation queue as the concurrency model, the session init flow end to end (including `initSafetyCheck`, plugin deferral, and the index-0 / index-1 insertion ordering), the queue-vs-driver split, and `BNCPreferenceHelper`'s archive-file (not `NSUserDefaults`) persistence plus the keychain values that survive reinstall.
- **Non-obvious invariants** — the ones that bite. Ordering is a guarantee, not a side effect; one in-flight request is enforced by `networkCount` + semaphore rather than by the queue; only `BranchEventRequest` is replayable on network failure and only when the client registered no callback; setting attribution level `NONE` flips `trackingDisabled` as a side effect; and **a new ObjC category needs a `BNCForce…CategoryToLoad` hook or it works in SPM and crashes at runtime in the static XCFramework**.

## How it was written

Read from this branch's source, workflows, scripts, podspec, `Package.swift`, and test plans — not from memory. Claims that are cheap to check (test plan contents, the designated initializer signature, source file counts, endpoint constants) were verified against the files.

## Review ask

Accuracy over completeness. The architecture and invariants sections are the parts worth scrutinising — if something is described as intended behavior when it's actually just current behavior, that's the failure mode to catch, because agents will treat this file as authoritative.


CLAUDE.md
```diff
@@ -0,0 +1,313 @@
...
+- **`getLatestReferringParams` reads `sessionParams`; `getFirstReferringParams` reads
+  `installParams`.** Install params are written once, on the first-ever install response.
+- **Tracking disabled ≠ attribution level NONE, but setting NONE forces tracking off.**
+  `setConsumerProtectionAttributionLevel:BranchAttributionLevelNone` flips
+  `trackingDisabled = YES` as a side effect; the levels are `FULL`, `REDUCED`, `MINIMAL`, `NONE`.
+- **Categories must be force-loaded, but not by the function that looks like it does it.** Static
+  linking strips ObjC categories, so each category exposes a no-op `BNCForce…CategoryToLoad`
+  symbol. What actually registers them is `__attribute__((constructor))` on the declaration in the
+  private header — present on `NSError+Branch`, `NSString+Branch`, `NSMutableDictionary+Branch`
+  and `UIViewController+Branch`, and **absent on `Branch+Validator`**.
```
gdeluna-branch, line 284: Branch+Validator.m:22 does define BNCForceBranchValidatorCategoryToLoad(void) — the symbol is present. What's actually absent is the
  __attribute__((constructor)) decoration on it; that's why it needs a manual call from ForceCategoriesToLoad() instead of self-registering. As written, an agent could
  grep for the symbol, find it, and wrongly conclude this file doesn't need the fix.

  Suggested fix:

  `-` and **absent on `Branch+Validator`**.
  `+` and **not `__attribute__((constructor))`-decorated on `Branch+Validator`** (it defines
  `+` `BNCForceBranchValidatorCategoryToLoad`, but only `ForceCategoriesToLoad()` calls it manually).

CLAUDE.md
```diff
@@ -0,0 +1,313 @@
...
+| ------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Session/init flow, init status, deep-link callbacks     | `Branch.m` — `initUserSessionAndCallCallback:`, `initializeSessionAndCallCallback:`, `handleInitSuccessAndCallCallback:`, `handleInitFailure:`                                                                                                                                                                                                                          |
```
NidhiDixit09, line 63: For the UIScene lifecycle, the app-facing entry is BranchScene.m initSessionWithLaunchOptions:registerDeepLinkHandler:

CLAUDE.md
```diff
@@ -0,0 +1,313 @@
...
+| Universal link / scheme / push / user-activity entry    | `Branch.m` — `handleDeepLink:sceneIdentifier:`, `handleSchemeDeepLink_private:`, `handleUniversalDeepLink_private:`, `continueUserActivity:`, `handlePushNotification:`                                                                                                                                                                                                 |
```
NidhiDixit09, line 64: Scene-based apps enter through BranchScene.m scene:continueUserActivity: and scene:openURLContexts:,

NidhiDixit09 (approved):

CLAUDE.md
wpinho-branch: Agreed, fixed in a8af58fc.

CLAUDE.md
wpinho-branch: Agreed, added a UIScene row to the table in 8d34c526.

CLAUDE.md
wpinho-branch: Covered by the same row in 8d34c526.

PR #1621, opened 2026-07-29
Create CLAUDE.md for 4.0.0-beta branch

wpinho-branch:
Adds repo context for Claude Code on the `4.0.0-beta.*` line. Companion to #1620 (same for `master`), mirroring the Android pair — [Create CLAUDE.md](https://github.com/BranchMetrics/android-branch-deep-linking-attribution/pull/1384) and [Create CLAUDE.md for beta branch](https://github.com/BranchMetrics/android-branch-deep-linking-attribution/pull/1385).

Each long-lived branch needs its own file: the `master` one describes a queue and an init path that no longer exist here, so pointing an agent at it on this branch is worse than having nothing.

## Why this is a separate document, not a diff

The doc opens with the four things that break 3.x knowledge carried onto this branch:

1. **The queue is an `NSOperationQueue`.** `BNCServerRequestQueue` no longer holds an array that `Branch.m` drains — `processNextQueueItem` does not exist here. Requests are wrapped in `BNCServerRequestOperation`.
2. **`/v3` endpoints and a two-request open.** `openServiceURL` is `/v3/events/open`, plus a new `deepLinkServiceURL` → `/v3/deeplink`.
3. **`requestDeepLinkData:callback:` and friends** are the new entry point; `initUserSessionAndCallCallback:` is deprecated (still present, still used by `continueUserActivity:`).
4. **`+setTrackingDisabled:` / `+trackingDisabled` are gone** from both the public header and `Branch.m`. Code referencing them will not compile here.

## What's in it

Same shape as #1620 — products, source layout, task → file table, build/test, CI, architecture, invariants — with this branch's specifics:

- **Tests moved.** `BranchSDKTests/` is a first-class target in `BranchSDK.xcodeproj`; `verify.yml` runs `xcodebuild test -scheme BranchSDKTests -testPlan BranchSDKTests`, no fastlane. `Branch-TestBed/Branch-SDK-Tests/` still exists but **CI does not run it on this branch** — a unit test added there is invisible.
- **No Layer-1 gate.** `layer1-logger-tests.yml` and `gptdriver-release.yml` are not on this branch, so there is no automated wire-field assertion; wire-format changes are reviewed by hand.
- **Ordering is now expressed two ways** — `NSOperationQueuePriority` (session-critical work goes in at `High`) *and* operation dependencies via `addInitDependencyIfNeeded:`.
- **The `/v3` flow documented end to end** — cancel pending deep-link ops → `BranchRequestDeepLink` (`/v3/deeplink`) → on `invoke_features` redirect, send attribution but skip the app callback → otherwise fire the callback then `sendOpen:skipCallback:` → `BranchRequestOpen` (`/v3/events/open`) carrying the resolved `link_data`. One user-visible open, two requests.
- **`sendServerRequest:` changed behavior** — on `master` it self-initializes via `initSafetyCheck`; here it refuses and fails the callback with `BNCInitError`.
- **The queue must be configured.** `configureWithServerInterface:branchKey:preferenceHelper:` is new and required, because the queue now builds the operations that perform requests.

## Invariants worth a second look

These are the ones I'd most like checked, since they're the difference between describing intent and describing current state:

- **`BranchOpenRequest` vs `BranchRequestOpen`** — both live, both hit `/v3/events/open`, wired into different flows. Deliberate, or a transitional state that should collapse?
- **The init dependency is a `weak` reference.** Once the init operation finishes and deallocates, later-enqueued requests are no longer gated behind it. Intended, or a gap?
- **Attribution `NONE` drops requests inside the operation**, before the network layer — a request can be enqueued, "succeed", and never have been sent. `BranchRequestDeepLink` is deliberately exempt.
- **`getLatestReferringParamsSynchronous` waits on three locks now.** A new session-establishing request class needs its own lock added there or the synchronous getter returns early.
- **`processing_sema` / `networkCount` on `Branch` appear vestigial** — serialization is `maxConcurrentOperationCount = 1`. I described them as such; correct me if they still carry weight.
- **There is no queue-level retry/replay** on this branch — the 3.x drain-and-fail path went away with `processNextQueueItem`. Only `BNCServerInterface`'s HTTP retries remain.

## Note

`BNC_SDK_VERSION` and the podspec still read `4.0.0-alpha.0` on this branch. Flagged in the doc, not changed here.

## Review ask

@bboothe-branch / @nidhi-branch — you wrote most of this code, so you're the right check on whether I documented the intended design or just the current state. Agents will treat this file as authoritative, so a wrong "invariant" here propagates.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -10,18 +10,23 @@ Carthage, and prebuilt XCFrameworks.
 > parts of the public API all differ. Everything below describes _this_ branch. When porting a
 > fix from `master`, assume the queue and init code do not translate directly — check first.
 >
-> Note: `BNC_SDK_VERSION` and the podspec still read `4.0.0-alpha.0` on this branch.
+> Note: nothing here has been renumbered for beta, and the version locations disagree.
+> `BNC_SDK_VERSION` and the podspec read `4.0.0-alpha.0`; `scripts/version.sh` reads `4.0.0`;
+> `MARKETING_VERSION` in the pbxproj reads `3.12.1`. See "Products and distribution" before
+> touching any of them.
 
 ## What changed from `master` (3.x)
 
 The four differences that break most assumptions carried over from 3.x:
 
 1. **The queue is an `NSOperationQueue`.** `BNCServerRequestQueue` no longer stores requests in
-   an array that `Branch.m` drains. There is no `processNextQueueItem` on this branch — the
-   symbol does not exist. Each request is wrapped in a `BNCServerRequestOperation` and the
-   operation queue executes it.
-2. **New `/v3` endpoints and a two-request open flow.** `openServiceURL` is now
+   an array that `Branch.m` drains. `processNextQueueItem` is gone from `Sources/` entirely — the
+   only remaining occurrences are a stale, unimplemented category declaration in
+   `BranchSDKTests/BranchEvent.Test.m` and its twin under `Branch-TestBed/`. Each request is
+   wrapped in a `BNCServerRequestOperation` and the operation queue executes it.
+2. **New `/v3` endpoints for the open flow, and a two-request open.** `openServiceURL` is now
    `/v3/events/open` (was `/v1/open`), and there is a new `deepLinkServiceURL` → `/v3/deeplink`.
+   Nothing else moved: installs are still `/v1/install`, events still `/v2/event/{standard,custom}`.
 3. **New public deep-link API.** `requestDeepLinkData:callback:` and friends replace driving
    everything through `initSession…`. `initUserSessionAndCallCallback:…` is now marked
    `__attribute__((deprecated))`.
@@ -34,8 +39,9 @@ The four differences that break most assumptions carried over from 3.x:
 
 One source tree, several delivery vehicles — all built from `Sources/BranchSDK/**/*.{h,m}`:
 
-- **SPM** — `Package.swift`, target `BranchSDK`, `publicHeadersPath: BranchSDK/Public/`,
-  private headers reachable only via the `BranchSDK/Private` header search path.
+- **SPM** — `Package.swift`, target `BranchSDK`, `path: "Sources/BranchSDK"` with
+  `publicHeadersPath: "Public"`; private headers reachable only via `.headerSearchPath("Private")`.
+  (Note these are relative to the target `path` here, unlike on `master`.)
 - **CocoaPods** — `BranchSDK.podspec`. tvOS **excludes** `BNCContentDiscoveryManager`,
   `BNCUserAgentCollector`, and `BNCSpotlightService`.
 - **XCFramework** — `BranchSDK.xcodeproj` schemes `xcframework`, `xcframework-noidfa`,
@@ -44,8 +50,17 @@ One source tree, several delivery vehicles — all built from `Sources/BranchSDK
   `GCC_PREPROCESSOR_DEFINITIONS=BRANCH_EXCLUDE_IDFA_CODE=1`. Any new AdSupport/IDFA touch must
   be guarded by that macro.
 
-Version lives in **two** places kept in sync by `scripts/version.sh`: `BNC_SDK_VERSION` in
-`Sources/BranchSDK/BNCConfig.m` and `s.version` in `BranchSDK.podspec`. Never hand-edit one.
+Version is **not** in two places and is **not** currently consistent. `scripts/version.sh` holds
+the source of truth as a hard-coded `version=` literal (line 33, currently `4.0.0`) and rewrites
+four targets: `BNC_SDK_VERSION` in `Sources/BranchSDK/BNCConfig.m`, `s.version` in
+`BranchSDK.podspec`, `MARKETING_VERSION` in `BranchSDK.xcodeproj/project.pbxproj`, and its own
+literal. Today those hold three different values: `4.0.0-alpha.0` (BNCConfig.m, podspec), `4.0.0`
+(version.sh), `3.12.1` (pbxproj).
+
+**Do not run `version.sh -i` or `-u` on this branch.** It would drop the `-alpha.0` suffix from
+both source files, and because `prev_version` is read from the script's own stale literal, the
+`MARKETING_VERSION` rewrite silently matches nothing — leaving the repo more divergent than it
+started. `bundle exec fastlane version_bump` is not an alternative here either; see CI below.
 
 ## Source layout
 
@@ -120,11 +135,16 @@ plans, but `verify.yml` on this branch does not use them. Adding a test to
 
 ### CI
 
-- **`verify.yml`** — the PR gate: `xcodebuild test -scheme BranchSDKTests -testPlan BranchSDKTests`.
+Ten workflows exist; several are inherited from `master` and **cannot run on this branch**.
+
+- **`verify.yml`** — the only working gate: `xcodebuild test -scheme BranchSDKTests -testPlan BranchSDKTests`.
 - **`pre-release-qa.yml` / `post-release-qa.yml`** — exercise every package manager (CocoaPods,
   Carthage, SPM, XCFramework) for iOS and tvOS against `SDKIntegrationTestApps/`.
-- **`release.yml`** — builds and signs XCFrameworks with checksums.
-- **`integration-tests.yml`** — manual only (`gh workflow run integration-tests.yml`).
+- **Broken here — fastlane was deleted on this line.** There is no `fastlane/` directory and no
+  `Gemfile` (both still exist on `master`), yet `version-bump.yml`, `integration-tests.yml` and
+  `release.yml` all shell out to `bundle exec fastlane` after a `bundle check || bundle install`.
+  Treat **release, version-bump and integration-tests as non-functional** on this branch until
+  fastlane is restored or they are ported.
 - **Not present on this branch:** `layer1-logger-tests.yml` (the L1 wire-validation gate) and
   `gptdriver-release.yml`. Both exist on `master`. There is no automated wire-field assertion
   here — wire-format changes are reviewed by hand.
@@ -173,9 +193,18 @@ serialized by the OS rather than by a hand-rolled semaphore loop. Its API is now
   everything else defaults to `Normal`.
 - **Operation dependencies** — `addInitDependencyIfNeeded:` records the most recent
   `BranchOpenRequest` operation as `currentInitOperation` and makes every **non-init** operation
-  `addDependency:` on it while it is unfinished. This is what holds ordinary requests behind
-  session establishment. It is a `weak` reference, so once the init operation finishes and
-  deallocates, later requests are no longer gated.
+  `addDependency:` on it while it is unfinished. It is a `weak` reference, so once the init
+  operation finishes and deallocates, later requests are no longer gated.
+
+  **A dependency outranks priority, and "init" here means only `BranchOpenRequest`.** Only
+  `BranchOpenRequest` and its subclass `BranchInstallRequest` satisfy that `isKindOfClass:` test.
+  `BranchRequestOpen` and `BranchRequestDeepLink` are plain `BNCServerRequest` subclasses, so
+  despite being enqueued at `High` they are treated as **non-init** and given a dependency on any
+  in-flight legacy init — they wait behind it. Conversely, on a pure `/v3` session nothing ever
+  becomes `currentInitOperation`, so the dependency gate never engages at all and ordering rests
+  on `maxConcurrentOperationCount = 1` plus priority alone. Note this is the opposite grouping
+  from the operation's own session-validation step, which treats all three classes as
+  session-establishing.
 
 **`BNCServerRequestOperation`** is a concurrent (`isAsynchronous == YES`) `NSOperation` with
 manual `isExecuting`/`isFinished` KVO. Its `start` does, in order:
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,331 @@
...
+   `/v3/events/open` (was `/v1/open`), and there is a new `deepLinkServiceURL` → `/v3/deeplink`.
+   Nothing else moved: installs are still `/v1/install`, events still `/v2/event/{standard,custom}`.
+3. **New public deep-link API.** `requestDeepLinkData:callback:` and friends replace driving
+   everything through `initSession…`. `initUserSessionAndCallCallback:…` is now marked
+   `__attribute__((deprecated))`.
+4. **`+setTrackingDisabled:` / `+trackingDisabled` are removed** from the public header _and_
+   from `Branch.m`. Attribution gating is entirely `setConsumerProtectionAttributionLevel:`
+   plus the new `+attributionLevelNone`. Code that references `Branch.trackingDisabled` will
+   not compile here.
+
```
bboothe-branch, line 37: 5. Branch.Initialize and applyConfiguration replacing the public setter APIs which are now marked `__attribute__((deprecated))`.
6. BranchConfiguration added to replace public API setters
7. BranchOpenRequest to be replaced with BranchRequestDeepLink and BranchRequestOpen files

PR For 5 and 6: https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/pull/1614

CLAUDE.md
```diff
@@ -0,0 +1,331 @@
...
+| HTTP transport, retries, timeouts                       | `BNCServerInterface.m` (`genericHTTPRequest:retryNumber:callback:retryHandler:`), `BNCNetworkService.m`                                                                                                                                                                          |
+| Device/hardware signals on requests                     | `BNCDeviceInfo.m`, `BNCSystemObserver.m`, `BNCDeviceSystem.m`, `BNCNetworkInterface.m`                                                                                                                                                                                           |
+| Link creation (short/long URLs)                         | `BranchShortUrlRequest.m`, `BranchShortUrlSyncRequest.m`, `BNCLinkData.m`, `BNCLinkCache.m`; content model in `BranchUniversalObject.m`, `BranchLinkProperties.m`                                                                                                                |
```
bboothe-branch, line 97: I think we're removing long URLs. Can confirm with @gdeluna-branch

CLAUDE.md
```diff
@@ -0,0 +1,331 @@
...
+| Referring-URL query params (gclid, gbraid, sccid, Meta) | `BNCReferringURLUtility.m`, `BNCUrlQueryParameter.m`                                                                                                                                                                                                                             |
+| URL ignore/skip list                                    | `BNCURLFilter.m`                                                                                                                                                                                                                                                                 |
+| `branch.json` / config flags / key source               | `BranchJsonConfig.m`, `BranchConfigurationController.m`, `Branch.m` `+branchKey`                                                                                                                                                                                                 |
```
bboothe-branch, line 104: Add `BranchConfiguration.m`

Relevant PR: https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/pull/1614

CLAUDE.md
```diff
@@ -0,0 +1,331 @@
...
+  fastlane is restored or they are ported.
+- **Not present on this branch:** `layer1-logger-tests.yml` (the L1 wire-validation gate) and
+  `gptdriver-release.yml`. Both exist on `master`. There is no automated wire-field assertion
+  here — wire-format changes are reviewed by hand.
+
+## Architecture
+
+### Singleton wiring (`Branch.m`)
+
+`Branch` is a process singleton reached through `+[Branch getInstance]`. The designated
```
bboothe-branch, line 156: Replaced with `[Branch initialize:config]`

PR: https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/pull/1614

CLAUDE.md
```diff
@@ -0,0 +1,331 @@
...
+- **`NSOperationQueuePriority`** — session-critical requests are enqueued with
+  `NSOperationQueuePriorityHigh` (init open/install, `BranchRequestOpen`, `BranchRequestDeepLink`);
+  everything else defaults to `Normal`.
+- **Operation dependencies** — `addInitDependencyIfNeeded:` records the most recent
+  `BranchOpenRequest` operation as `currentInitOperation` and makes every **non-init** operation
+  `addDependency:` on it while it is unfinished. It is a `weak` reference, so once the init
+  operation finishes and deallocates, later requests are no longer gated.
+
+  **A dependency outranks priority, and "init" here means only `BranchOpenRequest`.** Only
+  `BranchOpenRequest` and its subclass `BranchInstallRequest` satisfy that `isKindOfClass:` test.
```
bboothe-branch, line 200: I think we'll need to update the SDK at some point to run through `BranchRequestOpen` so we don't have to deal with confusing ourselves with the `BranchOpenRequest` class too.

CLAUDE.md
```diff
@@ -0,0 +1,331 @@
...
+**`BNCServerRequestOperation`** is a concurrent (`isAsynchronous == YES`) `NSOperation` with
+manual `isExecuting`/`isFinished` KVO. Its `start` does, in order:
+
+1. Bail if already cancelled.
+2. **Attribution gate** — if `attributionLevel == NONE`, drop the request silently, _except_
+   for `BranchRequestDeepLink`, which is always allowed through (deep linking must keep working
+   with attribution off).
+3. **Session validation** — install requests skip it; `BranchOpenRequest`, `BranchRequestOpen`,
+   and `BranchRequestDeepLink` skip it (they are what _establishes_ the session); everything
+   else requires `randomizedDeviceToken` **and** `sessionID` **and** `randomizedBundleToken`,
```
bboothe-branch, line 218: sessionID removed in a newer PR: https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/pull/1603

bboothe-branch (commented):
Added some changes and suggestions based on the public API changes: https://docs.google.com/document/d/1DD3gtM60zm9TkwM02tcuqcystC9-hyFd6yRzKnMFll0/edit?usp=sharing

CLAUDE.md
```diff
@@ -0,0 +1,331 @@
...
+| Queue behavior, ordering, dependencies, cancellation    | `BNCServerRequestQueue.m` **and** `BNCServerRequestOperation.m` — execution now lives in the operation, not in `Branch.m`                                                                                                                                                        |
+| Legacy session init (still present, deprecated)         | `Branch.m` — `initUserSessionAndCallCallback:`, `initializeSessionAndCallCallback:`, `handleInitSuccessAndCallCallback:`, `handleInitFailure:`                                                                                                                                   |
+| Universal link / scheme / push / user-activity entry    | `Branch.m` — `handleDeepLink:sceneIdentifier:`, `handleSchemeDeepLink_private:`, `handleUniversalDeepLink_private:`, `continueUserActivity:`, `handlePushNotification:`                                                                                                          |
```
bboothe-branch, line 89: `continueUserActivity` and [handlePushNotification](url) replaced with convenience functions in this PR: https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/pull/1605

gdeluna-branch:
While there's no rule, claude md shouldn't be 300 lines. Try to reduce by half.

From Claude: 

>   Practical guidance:
> 
>   - Aim for under ~100 lines / a screen or two. Many good ones are 20–50 lines.
>   - Include only what Claude can't cheaply discover itself: non-obvious build/test commands, conventions the code doesn't make obvious, gotchas, "always run X before
>   committing" rules.
>   - Leave out anything derivable from the repo — file structure, restating what the code does, generic best practices, long prose explanations.
>   - Prefer terse, imperative bullets over paragraphs. It's instructions, not documentation.
>   - If it's getting long, split it. You can add nested CLAUDE.md files in subdirectories (loaded when working in that area) or point to other docs to read on demand.
> 
>   Rule of thumb: if a line wouldn't change what Claude does, cut it. A focused 30-line file beats a comprehensive 300-line one — the long one dilutes the signal and
>   gets partially ignored anyway.

CLAUDE.md
```diff
@@ -0,0 +1,331 @@
...
+**`attributionLevel == NONE` short-circuits `sendOpen:skipCallback:`** and calls
+`clearLinkIdentifiers` so the same link identifiers are not reused on a later open.
+
+### Legacy session init (still present)
+
+`initUserSessionAndCallCallback:sceneIdentifier:urlString:reset:` still exists and still works —
+it is `deprecated`, not removed, and `continueUserActivity:` still routes through it. It funnels
+into `initializeSessionAndCallCallback:` which now **enqueues at `NSOperationQueuePriorityHigh`**
+instead of inserting at index 0/1, and no longer calls `processNextQueueItem`. Init status is
+still `BNCInitStatus { Uninitialized, Initializing, Initialized }` on `Branch.initializationStatus`.
```
bboothe-branch, line 267: We will be removing Initialization Statuses: https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/pull/1603

CLAUDE.md
```diff
@@ -0,0 +1,331 @@
...
+  assert on the interface, not on enqueue.
+- **`cancelPendingDeepLinkRequests` only cancels operations that are not executing.** A deep-link
+  request already in flight will still deliver its callback.
+- **`getLatestReferringParams` reads `sessionParams`; `getFirstReferringParams` reads
+  `installParams`.** Install params are written once, on the first-ever install response.
+- **Categories must be force-loaded, but not by the function that looks like it does it.** Static
+  linking strips ObjC categories, so each category exposes a no-op `BNCForce…CategoryToLoad`
+  symbol. What registers them is `__attribute__((constructor))` on the private-header
+  declaration — present on `NSError+Branch`, `NSString+Branch`, `NSMutableDictionary+Branch` and
+  `UIViewController+Branch`, and **absent on `Branch+Validator`**. `ForceCategoriesToLoad()` in
```
bboothe-branch, line 309: Reposting @gdeluna-branch 's comment on this PR: https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/pull/1620/changes

_Branch+Validator.m:22 does define BNCForceBranchValidatorCategoryToLoad(void) — the symbol is present. What's actually absent is the
attribute((constructor)) decoration on it; that's why it needs a manual call from ForceCategoriesToLoad() instead of self-registering. As written, an agent could
grep for the symbol, find it, and wrongly conclude this file doesn't need the fix.

Suggested fix:

- and absent on Branch+Validator.
+ and not __attribute__((constructor))-decorated on Branch+Validator (it defines
+ BNCForceBranchValidatorCategoryToLoad, but only ForceCategoriesToLoad() calls it manually)._

CLAUDE.md
wpinho-branch: Covered by the new `## In flight` section in 014df94c, which lists #1614, #1603 and #1605 and what each changes. Keeping the body on current state: all three are still open, and `4.0.0-beta.0` today still has the old shape — a doc describing unmerged API sends an agent looking for symbols that do not exist yet.

CLAUDE.md
wpinho-branch: Covered by the new `## In flight` section in 014df94c, which lists #1614, #1603 and #1605 and what each changes. Keeping the body on current state: all three are still open, and `4.0.0-beta.0` today still has the old shape — a doc describing unmerged API sends an agent looking for symbols that do not exist yet. Confirmed on the branch: only `BranchConfigurationController.m` exists today, no `BranchConfiguration.m`.

CLAUDE.md
wpinho-branch: Covered by the new `## In flight` section in 014df94c, which lists #1614, #1603 and #1605 and what each changes. Keeping the body on current state: all three are still open, and `4.0.0-beta.0` today still has the old shape — a doc describing unmerged API sends an agent looking for symbols that do not exist yet.

CLAUDE.md
wpinho-branch: Covered by the new `## In flight` section in 014df94c, which lists #1614, #1603 and #1605 and what each changes. Keeping the body on current state: all three are still open, and `4.0.0-beta.0` today still has the old shape — a doc describing unmerged API sends an agent looking for symbols that do not exist yet. `sessionID` is still live here — `BNCPreferenceHelper.h:32`, and session validation still requires it.

CLAUDE.md
wpinho-branch: Covered by the new `## In flight` section in 014df94c, which lists #1614, #1603 and #1605 and what each changes. Keeping the body on current state: all three are still open, and `4.0.0-beta.0` today still has the old shape — a doc describing unmerged API sends an agent looking for symbols that do not exist yet. `BNCInitStatus` is still defined at `Branch.m:134-137` and still used.

CLAUDE.md
wpinho-branch: Covered by the new `## In flight` section in 014df94c, which lists #1614, #1603 and #1605 and what each changes. Keeping the body on current state: all three are still open, and `4.0.0-beta.0` today still has the old shape — a doc describing unmerged API sends an agent looking for symbols that do not exist yet.

CLAUDE.md
wpinho-branch: Fixed in 014df94c — same wording as the master CLAUDE.md.

CLAUDE.md
wpinho-branch: Left as is for now. Long URLs are still live on this branch — `Branch.m:1439-1448` (`getLongURLWithParams:` and friends) plus two call sites in `BNCSpotlightService.m`. Happy to change it the moment removal is confirmed; @gdeluna-branch, is it planned for the 4.0 line?

CLAUDE.md
wpinho-branch: Agreed on the direction, and the doc already flags the trap this creates today — the routing table and the invariants both call out that `BranchOpenRequest` and `BranchRequestOpen` are different live classes, and the architecture section spells out that the init dependency only recognises `BranchOpenRequest`, so `BranchRequestOpen` and `BranchRequestDeepLink` are treated as non-init despite being enqueued High. Nothing to change here until the consolidation lands.

wpinho-branch:
Condensed to 181 lines in 014df94c — 331 before, a 45% cut, not quite the half you asked for.

What I kept is the routing table (26 lines) and the non-obvious invariants. Those are the parts an agent cannot cheaply derive by reading the source, so cutting them to hit a line count would trade away the reason the file exists. What went was the architecture prose that restated what the code already says.

Also added an `## In flight` section pointing at #1614, #1603 and #1605, which answers most of the review comments at once without making the body describe API that has not merged yet.

Say the word if you want it under 165 and I will cut the architecture section further.

PR #1623, opened 2026-08-03
EMT-4010 - Additive: Async Swift wrappers for SDK functions

bboothe-branch:
## Reference
EMT-4010 added async swift wrappers for Obj C functions

## Summary

Added Swift wrappers for Obj C functions. Can be used in Swift based apps.

## Type Of Change

## Testing Instructions

1. Open iOSReleaseTests app
2. View AppDelegate.swift and SceneDelegate.swift
3. Uncomment the async function you would like to test
4. Run app
5. View logs with changed information


Sources/BranchSDK/Public/Branch.h
```diff
@@ -541,7 +541,8 @@ extern NSString * __nonnull const BNCSpotlightFeature;
  When called with a non-nil URL, any pending (not yet executing) nil-URL deep link requests are
  cancelled so that only one callback fires per app open.
  */
-- (void)requestDeepLinkData:(nullable NSString *)branchLink callback:(nullable callbackWithParams)callback;
+- (void)requestDeepLinkData:(nullable NSString *)branchLink callback:(nullable callbackWithParams)callback
+    NS_SWIFT_ASYNC_NAME(requestDeepLinkData(_:));
```
wpinho-branch: This one has a path where the completion handler never fires, and under the async import that becomes a hang with no way out.

`BranchRequestDeepLink.m` returns early when `invokeFeatures:` succeeds, with a comment saying the callback is skipped because a weblink was launched. I stubbed a response using the real keys from BranchConstants.m to reach that branch and called `requestDeepLinkData:callback:` on this commit. The callback never fires and the waiter times out.

For a callback API that's the correct thing to do, the user is already on their way to a browser. For `await branch.requestDeepLinkData(url)` the continuation never resumes, so the task sits suspended with no timeout and nothing to cancel it.

Resuming with a sentinel on that branch would close it, so would leaving this method off the async surface. Which one depends on what you want the async contract to say.

Sources/BranchSDK/Public/Branch.h
```diff
@@ -1063,7 +1067,8 @@ extern BranchAttributionLevel const BranchAttributionLevelNone;
  */
 - (void)logout;
 
-- (void)logoutWithCallback:(nullable callbackWithStatus)callback;
+- (void)logoutWithCallback:(nullable callbackWithStatus)callback
+    NS_SWIFT_ASYNC_NAME(logout());
```
wpinho-branch: The collision here is real, but the mechanism isn't ambiguity. Swift won't let anyone reach the sync one by accident, it just refuses to compile.

`callbackWithStatus` carries an `NSError *`, so the importer doesn't synthesise `func logout() async`. It synthesises `func logout() async throws -> Bool`. I built the framework from this commit and type-checked against it: calling `b.logout()` inside an async function fails for both the missing `await` and the missing `try`, and adding only `await` still fails on the `try`. It compiles with `try await`, and sync callers are untouched.

So the effect isn't someone calling the wrong method, it's that async-context code that exists today stops building until it adds `try await`. The same goes for `logEvent()` in BranchEvent.h, and for any annotated method whose completion carries an error.

Renaming the async variant avoids it at the cost of an unidiomatic Swift name. Keeping `logout()` and putting the break in the changelog is the other way. Either is fine, it just shouldn't land by accident.

wpinho-branch (changes requested):
Two findings, both measured on this branch. One is a hang with no way out under the async import, the other is a source break in a PR labelled additive.

Worth a line in the description too: EMT-4010 asks for async wrapper variants written in Swift, and this delivers importer annotations instead. That may well be the better call, it just isn't what the ticket says.

I built the framework from this branch and type-checked Swift call sites against it, and ran an XCTest against the deep link path. I didn't try it against a live Branch link.

Sources/BranchSDK/Public/Branch.h
bboothe-branch: Hi @wpinho-branch I think this deserves its own ticket for how we want to solve this. If we want to keep it async we have to do the sentinel. If we disable async that goes against the expanded API spec. I'd like to discuss with the team.

gdeluna-branch (commented):
Please add tests.

Sources/BranchSDK/Public/Branch.h
bboothe-branch: Renamed the async variant to fix

bboothe-branch:
Tests Added

NidhiDixit09 (approved):

PR #1624, opened 2026-08-04
EMT-4011 - isRetryable key for BNCError

bboothe-branch:
Set up error handling retryability to determine whether a request should be retried or not.

## Reference
EMT-4011-- isRetryable key for BNCError

## Summary

- Added Retry key that checks for type of error
- Added tests to ensure retry works

## Type Of Change

## Testing Instructions


Sources/BranchSDK/NSError+Branch.m
```diff
@@ -73,9 +75,73 @@ + (NSError *)branchErrorWithCode:(BNCErrorCode)errorCode error:(NSError *)error
...
+
     return [NSError errorWithDomain:[self bncErrorDomain] code:errorCode userInfo:userInfo];
 }
 
++ (BOOL)branchErrorIsRetryableForCode:(BNCErrorCode)errorCode {
+    switch (errorCode) {
+        // Transient failures — retrying may succeed once conditions improve.
+        case BNCServerProblemError:            // HTTP 5xx / trouble reaching Branch servers
+        case BNCNetworkServiceInterfaceError:  // timeout / connection reset / underlying network error
+            return YES;
```
wpinho-branch, line 88: `BNCNetworkServiceInterfaceError` isn't a transient failure. Every place it gets raised is inside `verifyNetworkOperation:` in BNCServerInterface.m, and each one means the host app's own `BNCNetworkServiceProtocol` conformer is broken: it returned nil, or doesn't conform, or left `startDate`, `timeoutDate` or `request` unset. Retrying can't fix any of those.

I checked what a caller actually receives. Driving a request through `BNCServerInterface` with a conformer that returns nil delivers that code with the retryable key set to YES, and the network service only gets asked for an operation once, so the SDK never retried it at all.

That second part is what makes it worth changing. The doc you added to Branch.h tells apps that a retryable error means the automatic retries were already exhausted, and here there weren't any. An app following that contract ends up retrying a permanently broken configuration on its own.

```suggestion
        case BNCServerProblemError:            // HTTP 5xx / trouble reaching Branch servers
            return YES;
```

Sources/BranchSDK/BNCServerInterface.m
```diff
@@ -273,6 +283,17 @@ - (BNCServerResponse *)genericHTTPRequestSynchronous:(NSURLRequest *)request {
 
 #pragma mark - Internals
 
+// Exponential backoff for the Nth retry (0-based). Returns the configured base retryInterval
+// scaled by 2^retryNumber, so successive retries wait interval, 2*interval, 4*interval, …
+// A base interval of 0 always returns 0 (immediate retry), matching the historical default.
+- (NSTimeInterval)backoffDelayForRetryNumber:(NSInteger)retryNumber {
+    NSTimeInterval base = self.preferenceHelper.retryInterval;
```
wpinho-branch, line 290: EMT-4011 describes the backoff as one second, then two, then four. This scales `preferenceHelper.retryInterval`, which defaults to zero, so out of the box nothing waits at all and `testBackoffIsZeroWhenIntervalIsZero` in this PR asserts exactly that.

Is the idea to keep the current default and let apps opt in, or to make the ticket's one-second base the new default? Either is fine, they just say different things right now.

wpinho-branch (changes requested):
The classification idea is good and the allowlist in `branchErrorByAnnotatingRetryable:` is the right shape for raw NSURLErrors. One code in the retryable list doesn't belong there, and the backoff doesn't do what the ticket describes.

I ran the existing suite plus a couple of throwaway checks on this branch. I didn't exercise a real 5xx against staging.

Sources/BranchSDK/NSError+Branch.m
bboothe-branch: Updated this by removing the BNCNetworkServiceInterfaceError

Sources/BranchSDK/BNCServerInterface.m
bboothe-branch: Set the retry interval to 1 to remove the opt-in.

NidhiDixit09 (approved):

PR #1625, opened 2026-08-04
EMT-4013 - Branch Interface Protocol

bboothe-branch:
BranchInterface.h allows developers to run Unit Tests and separate production and test uses of Branch SDK

## Reference
EMT-4013 -- iOS - Branch Interface Protocol

## Type Of Change

## Testing Instructions


Sources/BranchSDK/Public/BranchSDK.h
```diff
@@ -15,6 +15,7 @@ FOUNDATION_EXPORT const unsigned char BranchSDKVersionString[];
 
 // In this header, you should import all the public headers of your framework using statements like #import <BranchSDK/PublicHeader.h>
 #import "Branch.h"
+#import "BranchInterface.h"
```
wpinho-branch, line 18: The protocol only made it into this umbrella. `Framework/BranchSDK.h` and `Sources/Resources/BranchSDK.h` don't import it, and #1614 updated all three when it added `BranchConfiguration.h`. Carthage and manual-framework consumers read those two, so as it stands they'd get `Branch` conforming to a protocol they can't name.

Sources/BranchSDK/Public/BranchInterface.h
```diff
@@ -0,0 +1,84 @@
...
+
+- (void)requestDeepLinkDataWithLaunchOptions:(nullable NSDictionary *)options
+                                    callback:(nullable callbackWithParams)callback
+    NS_SWIFT_ASYNC_NAME(requestDeepLinkData(withLaunchOptions:));
+
+#if !TARGET_OS_TV
+- (void)requestDeepLinkDataWithSceneOptions:(nullable UISceneConnectionOptions *)connectionOptions
+                                      scene:(UIScene *)scene
+                                   callback:(nullable callbackWithParams)callback
+    NS_SWIFT_ASYNC_NAME(requestDeepLinkData(withSceneOptions:scene:))
```
wpinho-branch: This block covers the string, launch options and scene options entry points, but not `requestDeepLinkDataWithUserActivity:completion:`, which EMT-4013 has in its protocol definition. That's the Universal Links path, and for a protocol whose whole point is unit-testing deep link handling it's probably the one people most want to fake. Left out on purpose, or did it fall out along the way?

Separately, `BranchAttributionLevel` is redeclared here behind a macro guard while Branch.h declares it without one. It compiles, since the typedefs are identical, but two declarations of the same type in two headers will drift eventually. A small shared header both import would avoid that.

wpinho-branch (commented):
The protocol does what EMT-4013 is after, and the Swift test that injects a fake is a good way to prove it's actually mockable. Two gaps in what got exported, both mechanical.

One note for anyone reading this on its own: the `resetInitializationGuardForTesting` selector the Swift test probes for does exist here, it comes from #1614 further down the stack, so the `responds(to:)` guard resolves and the reset runs. A fair amount of what this PR leans on lands in the PRs below it.

In passing, the pbxproj entries use hand-written object IDs rather than Xcode-generated ones.

I didn't run the test target on this branch.

Sources/BranchSDK/Public/BranchInterface.h
bboothe-branch: Updated the interface with new convenience functions and added a shared import for BranchAttributionLevel

Sources/BranchSDK/Public/BranchSDK.h
bboothe-branch: This was pulled into each of those files on line 18 of each.

NidhiDixit09 (approved):

PR #1630, opened 2026-08-07
EMT-4026: make the deprecated device-token fallback reachable when the modern key is absent

wpinho-branch:
## Reference
[EMT-4026](https://branch.atlassian.net/browse/EMT-4026) -- make the deprecated device-token fallback reachable when the modern key is absent.

## Summary
`BranchRequestOpen.m` guarded the `device_fingerprint_id` fallback behind the presence of the modern `randomized_device_token` key, so the fallback was unreachable in exactly the case it exists for and no device token was stored. `BranchOpenRequest.m` carried the same defect as an early `return nil`. Both now follow the pattern the bundle token already uses in these files: read the modern key, fall back when empty, no outer guard, assign only when a value was obtained. `BNCStringFromWireFormat` is applied to both reads, so a numeric `device_fingerprint_id` cannot reach the `NSString` property and crash the setter on the next open.

## Motivation
A fresh install completes its open and then drops every request behind it -- `BNCServerRequestOperation` drops any non-open request when the device token is missing. `/v3/events/open` answers a fresh install with the legacy identity shape (`device_fingerprint_id`, `identity_id`, `session_id`), so this line hits the guard; `master` never does, because `/v1/open` always returns the modern key. Latent pre-existing defect exposed by the endpoint change, not a regression introduced here. A warm launch never reaches it, which is why it went unseen: every capture until now had been a warm open.

## Type Of Change

## Testing Instructions
New test class, 9 cases across both request classes: deprecated key only, modern key only, both, neither, numeric values. Before the change, 6 assertion failures in 4 cases. After, 9/9 pass; full suite `492 tests, 0 failures` -- [run 31654961566](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/actions/runs/31654961566).

Runtime, fresh install of Branch-TestBed with four requests issued behind the open, identical driver in both runs:

| | before | after |
| --- | --- | --- |
| requests on the wire | 1 | 5 |
| `Missing session items` | 4 | 0 |

The response carried `device_fingerprint_id` and zero occurrences of the modern key; after the change that value comes back out under the modern name on subsequent requests, so store-and-reuse is confirmed rather than storage alone. Reproducing the runtime capture needs the TestBed log hook from #1627; the unit tests are self-contained.

[EMT-4026]: https://branch.atlassian.net/browse/EMT-4026?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BranchSDKTests/BranchOpenResponseDeviceTokenTests.m
+++ b/BranchSDKTests/BranchOpenResponseDeviceTokenTests.m
@@ -12,7 +12,7 @@
 #import "BranchOpenRequest.h"
 #import "BranchRequestOpen.h"
 
-static NSString * const kDeprecatedDeviceTokenKey = @"device_fingerprint_id";
+static NSString * const deprecatedDeviceTokenKey = @"device_fingerprint_id";
 
 @interface BranchOpenResponseDeviceTokenTests : XCTestCase
 @property (nonatomic, copy) NSString *savedDeviceToken;
@@ -51,7 +51,7 @@ static NSString * const kDeprecatedDeviceTokenKey = @"device_fingerprint_id";
 // The regression: a response carrying only the deprecated key stored no device token,
 // because the fallback was nested inside a guard requiring the modern key.
 - (void)testRequestOpenStoresDeviceTokenFromDeprecatedKeyOnly {
-    [self processOpenResponseData:@{ kDeprecatedDeviceTokenKey: @"legacy_device_token" }];
+    [self processOpenResponseData:@{ deprecatedDeviceTokenKey: @"legacy_device_token" }];
 
     XCTAssertEqualObjects([BNCPreferenceHelper sharedInstance].randomizedDeviceToken, @"legacy_device_token");
 }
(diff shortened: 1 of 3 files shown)


BranchSDKTests/BranchOpenResponseDeviceTokenTests.m
```diff
@@ -0,0 +1,122 @@
...
+//
+
+#import <XCTest/XCTest.h>
+#import "BranchConstants.h"
+#import "BNCPreferenceHelper.h"
+#import "BNCServerResponse.h"
+#import "BranchOpenRequest.h"
+#import "BranchRequestOpen.h"
+
+static NSString * const kDeprecatedDeviceTokenKey = @"device_fingerprint_id";
```
bboothe-branch: Any particular reason this is kDeprecatedDeviceTokenKey instead of deprecatedDeviceTokenKey?

bboothe-branch (approved):

BranchSDKTests/BranchOpenResponseDeviceTokenTests.m
wpinho-branch: No reason beyond the usual Objective-C habit for constants. Nothing else in either test tree uses the `k` prefix and the SDK sources spell theirs in upper snake case, so it was carrying no local convention. Renamed in 3a8d43e9.

PR #1631, opened 2026-08-10
EMT-4034: make a request operation complete exactly once

wpinho-branch:
## Reference
[EMT-4034](https://branch.atlassian.net/browse/EMT-4034) -- make a request operation complete exactly once.

## Summary
Completion now belongs to `start` and `finishOperation` only, both on the queue's thread. `cancel` no longer writes state -- the queue starts every enqueued operation and `start`'s `isCancelled` path finishes it. A lock-guarded claim elects one owner, setters are equality-guarded, the KVO-observed ivars are atomic, and the response callback gets a re-entry guard.

## Motivation
`cancel` on the main thread and `start` on the queue's thread could both observe `executing=0, cancelled=1`, so `NSOperationQueue` received two completions for one operation:

```
*** BNCServerRequestOperation 0x... went isFinished=YES without being started by the queue it is in
```

The host then crashes, `xcodebuild` restarts and blames whichever test was running, which is why `verify` is red across the beta line for every author, on tests they do not touch. Introduced by the `NSOperationQueue` migration (#1572), so no customer is exposed -- `master` is structurally incapable of it and only `4.0.0-alpha` is published. Still a release gate: the testbed reaches it through `application:openURL:` on the main thread.

## Also in this PR
`BranchDeepLinkQueueDrainTests.m` carries [EMT-4174](https://branch.atlassian.net/browse/EMT-4174). It references `sessionID` and, through KVC, `initializationStatus`, both removed by #1603, so the unit target has not compiled since #1632 merged. The race fix above is not verifiable without it, so the two ship together.

## Type Of Change

## Testing Instructions
Branch-TestBed, 10 launches per build on one simulator, five deep links per launch delivered while install/open are in flight, AddressSanitizer on:

| build | Foundation warning | ASan | crashes |
| --- | --- | --- | --- |
| pristine | 20 (in 10/10 runs) | 8 | 2 |
| patched | 0 | 0 | 0 |

Both builds delivered all 50 deep links. Unit suite, alternating pristine and patched runs on one pinned simulator: ASan 10 crashes / 15 runs to 0 / 15; unsanitized 2 / 20 to 0 / 20; Foundation warning 35 / 35 runs to 0 / 35.

Reproducible without the soak, `xcodebuild test -scheme Branch-TestBed-CI` on base `c0475277` against this head, one run each on a freshly erased iPhone 16e: both sides `Executed 507 tests, with 19 failures` and an identical set of 9 failing tests, so the change is regression-free; `went isFinished=YES` goes **2 to 0**.

Review round: the first version also moved `verifyNetworkOperation:` ahead of `[operation start]`, which rejected every request because the validator reads `startDate`, set only by `-start`. Removed in 66690f80, with a regression test for `genericHTTPRequest:`.

[EMT-4034]: https://branch.atlassian.net/browse/EMT-4034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[EMT-4174]: https://branch.atlassian.net/browse/EMT-4174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gdeluna-branch (changes requested):
Awaiting offline feedback.

wpinho-branch:
Reordering removed in 66690f80, with a regression test for `genericHTTPRequest:`. `NSRecursiveLock` → `NSLock` in d333b5f1. PR body trimmed.

wpinho-branch:
Ran it on Branch-TestBed: 10 launches per build, deep links delivered while install/open are in flight, ASan on. Pristine gave 20 Foundation warnings across 10/10 runs and 2 crashes, faulting in `__NSOQSchedule` under the queue's own KVO observer. Patched gave zero on every count. Numbers and stack are in the description.

bboothe-branch (approved):
Hey Willian. Closing [my PR on this](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/pull/1641/changes) same fix as yours is more robust. Would consider looking at my setup where the early exit just calls finishOperation instead of repeating executing = NO; finished = YES;  at each one. Might give a slightly cleaner look. Otherwise your callback guards make this much better.

NidhiDixit09 (approved):

wpinho-branch:
Applied in 064d7fae — the three early exits in -start now call finishOperation. 499 unit tests pass.

PR #1634, opened 2026-08-10
EMT-4029: consult the URL filters before resolving a deep link

wpinho-branch:
## Reference
[EMT-4029](https://branch.atlassian.net/browse/EMT-4029) -- consult the URL filters before resolving a deep link.

Stack 3 of 4. Base is #1633 -- review #1632, then #1633, first.

## Summary
Both URL filters now run through one predicate shared by `-requestDeepLinkData:callback:` (`Branch.m:2477`) and `handleDeepLink:`, so the two entry points cannot drift apart again. A filtered URL is recorded, sets `dropURLOpen`, and completes the session unattributed via `-sendOpen`, matching `handleDeepLink:` rather than merely skipping the request. The check runs before `-cancelPendingDeepLinkRequests`, so a URL that will not be resolved cannot discard a resolution already in flight. The caller's callback receives `{"+clicked_branch_link": 0}` and no error -- nothing failed, and never calling back would hang the caller. `handleDeepLink:` keeps its behaviour exactly.

## Motivation
Both filters were read in exactly one place, inside the legacy `handleDeepLink:`. The documented 4.0 entry point posted to `/v3/deeplink` unconditionally, so a URL the host app passed to `setUrlPatternsToIgnore:`, or one the server skiplist marks as credential-bearing, still went on the wire. That is a regression against the shipping line, not a gap in a new feature.

## Type Of Change

## Testing Instructions
Three tests added to `BranchRequestDeepLinkTests.m`: one per filter asserting no `BranchRequestDeepLink` is enqueued at all, plus a control that an unfiltered URL still is -- without it both absence assertions could pass for want of a working queue. `setUp` now saves and restores `userURLFilter`, since the ignore-pattern test is the first to write one and it would otherwise follow the singleton into later tests.

`492 tests, 0 failures` -- [run 31436313382](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/actions/runs/31436313382), which is 3 over the 489 #1633 stood at when this was branched. #1633 has since gained 2 tests; this branch has not been rebased onto them yet.

[EMT-4029]: https://branch.atlassian.net/browse/EMT-4029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Sources/BranchSDK/Branch.m
```diff
@@ -2485,6 +2499,35 @@ + (void) clearAll {
 - (void) requestDeepLinkData:(NSString *)branchLink callback:(nullable callbackWithParams)callback {
     [[BranchLogger shared] logDebug:[NSString stringWithFormat:@"requestDeepLinkData called with branchLink: %@", branchLink] error:nil];
 
+    // Same question handleDeepLink: asks, asked before anything is cancelled or enqueued: a URL
+    // the skiplist or the host app told us to ignore must not reach /v3/deeplink. It is also
+    // asked before -cancelPendingDeepLinkRequests, so an ignored URL cannot discard a resolution
+    // already in flight for a link we are allowed to resolve.
+    NSString *ignoredPattern = nil;
```
NidhiDixit09, line 2071: @wpinho-branch These are already taken care by @bboothe-branch in another PR (for convenience methods.). So revert fix from here otherwise it will create a merge conflict later.

NidhiDixit09:
@wpinho-branch Dont add any fixes in Branch.m unless they are required for tests. If tests are failing, let them. Please create a ticket for any issues found. Brandon is also fixing this part of code. So align with him.
Can you please revert updates in Branch.m (or move them to other branch and get Brandon's review) so that I can approve rest of the PR. cc: @gdeluna-branch

Sources/BranchSDK/Branch.m
wpinho-branch: Reverted in 1bcf74da. The helpers and the handleDeepLink: rewrite are gone; Branch.m is now +46/-2 and the conflict cleared. The guard in requestDeepLinkData: stays, since processDeepLinkURL: is not reached by requestDeepLinkDataWithLaunchOptions: or the scene entry points.

NidhiDixit09 (approved):

PR #1642, opened 2026-08-24
EMT-4174: fix the beta unit build after the session-id removal

wpinho-branch:
The beta unit target has not compiled since #1632 merged. `BranchDeepLinkQueueDrainTests` references `BNCPreferenceHelper.sessionID`, `BRANCH_RESPONSE_KEY_SESSION_ID` and, through KVC by string, `Branch.initializationStatus`. None of the three exists on this line any more — they were removed in #1603, merged 2026-08-21. The test file arrived afterwards in #1632, merged 2026-08-24, having been written before the removal, so its CI had last run against an older base.

The session-id references are compile errors, so the target fails to build and nothing runs. The `initializationStatus` ones go through KVC by string, so they compile and then raise `NSUnknownKeyException` at runtime.

Every PR currently open against this line still displays a green `verify` from a run that predates the break — #1614 from 08-22, #1633 and #1631 from 08-18, #1639 from 08-19 — so the first branch to be rebased turns red.

**On the removed precondition.** The test asserted the session read uninitialized after resigning active. That state no longer decides anything: `applicationDidBecomeActive` skips the organic open on exactly two conditions, `attributionLevelNone` and an init request already in the queue, and `containsInstallOrOpen` recognises `BranchRequestDeepLink`. Attribution is the only unrelated reason left to rule out, and the test already asserts it, so the assertion is removed rather than ported.

**Proof.** On `e3a905b2` the target fails to build and zero tests run. With this change 498 tests execute and `testLiveResolutionSuppressesTheLifecycleOrganicOpen` passes. The 10 remaining local failures (`BNCKeyChainTests`, `BranchConfigurationControllerTests`, `BNCApplicationTests`) reproduce identically on `0eac1acf`, the last commit whose `verify` was green in CI, so they are environmental and unrelated to this change.

Test code only, one file. No SDK behaviour changes.

### DIFF
--- a/BranchSDKTests/BranchDeepLinkQueueDrainTests.m
+++ b/BranchSDKTests/BranchDeepLinkQueueDrainTests.m
@@ -69,7 +69,6 @@ static NSMutableArray<NSString *> *sPostedURLs = nil;
         response.data = @{ BRANCH_RESPONSE_KEY_SESSION_DATA: kResolvedLinkPayloadJSON };
     } else {
         response.data = @{
-            BRANCH_RESPONSE_KEY_SESSION_ID: @"session_id",
             BRANCH_RESPONSE_KEY_RANDOMIZED_BUNDLE_TOKEN: @"bundle_token",
             BRANCH_RESPONSE_KEY_RANDOMIZED_DEVICE_TOKEN: @"device_token"
         };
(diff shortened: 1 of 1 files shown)


wpinho-branch:
Absorbed into #1631, which carries this commit and goes green with the EMT-4034 fix that this branch alone cannot pass without. Closing in favour of that PR.

PR #1644, opened 2026-08-27
docs: correct the queue's init grouping and where ordering comes from

wpinho-branch:
The `Queue.` paragraph described the pre-EMT-4028 code.

It said init meant only `BranchOpenRequest` and `BranchInstallRequest`, that `BranchRequestOpen` and `BranchRequestDeepLink` counted as non-init, and that on a pure `/v3` session nothing became `currentInitOperation` so the gate never engaged at all. `isInitRequest:` has covered all three classes since #1632, and now matches the operation's own session-validation grouping rather than opposing it.

## Why this is worth a PR rather than a tidy-up

A reviewer investigating request ordering read this file, found it corroborated by a stale worktree on disk, and concluded the gate was disabled by a class-name mismatch that had already been fixed. Right process, wrong verdict, because the repo's own documentation confirmed the wrong tree. Documentation that describes behaviour is load-bearing for anyone reasoning about a defect, and stale is worse than absent — absent makes you go read the code.

## What else changed

**Where ordering actually comes from.** The paragraph led with the init dependency. On 4.0 the ordering rests on the queue being serial with session work at `High`; the dependency refines that when an init operation happens to be tracked. The second commit on this branch corrects the first, which had described the dependency not engaging on `/v3` as a hole and pointed at EMT-4207.

It is not a hole. 4.0 is token-gated rather than state-machine-driven: session identity is the persisted randomized bundle and device tokens, and a session-dependent request missing either is dropped immediately with `BNCInitError` — never held, never replayed. EMT-4207 is closed as working as designed, and the corrected text now says so rather than describing a gap.

## Scope

Documentation only. No code changes.


wpinho-branch:
Closing rather than merging. The correction this branch carried is already on 4.0.0-beta.0: docs/architecture.md states the isInitRequest: grouping and the weak-reference caveat, in more detail than this had.

What remained was a framing claim, that ordering rests on serialization rather than on priority plus dependencies. Nothing in the repo settles it: queuePriority is assigned once, at BNCServerRequestQueue.m:66, and appears in no test. EMT-4269 adds the tests that would.

PR #1647, opened 2026-08-28
EMT-4082: contract the first install via field presence, cold_firstInstall (2 of 3)

wpinho-branch:
A contract can now count field presence, not just requests. `cold_firstInstall` is why: it is the same cold Universal Link launch as `cold_https`, on a device with no prior install, and this line has no install endpoint. Install is decided client side by `randomizedBundleToken == nil` and posts to `/v3/events/open` like anything else, so the counts and the order of the two scenarios are identical. The install is the one open of the two carrying no token, and `{"/v3/events/open": {"randomized_bundle_token": 1}}` against `2` is the only wire signal separating them. Presence only, never a value comparison, so the layer stays the layer it claims to be.

The test link is delivered on session start rather than on a timer, so the capture is the scenario and not a race.

Merges `feat/EMT-4082-cold-scenarios` at `a7bb989c`, which renamed the registry keys, and applies the same treatment to what this branch adds: `C3` becomes `cold_firstInstall`. Six fixtures moved with `git mv`, all pure renames. The iOS captures carry no stamped scenario value, so unlike the Android side there is no fixture content to move with the key.

Suite unchanged at 94 tests. Every key passes on its own fixture and fails against another, with one exception worth naming: `deeplink` accepts both cold captures, because its contract constrains only the resolve count and the order pair. That permissiveness predates this change.

Step ids are untouched, so the `if:` guards that turn a scenario failure into a job failure keep resolving.


Branch-TestBed/Branch-TestBed/TestBedDeepLinkTestHook.h
```diff
@@ -18,6 +18,10 @@ NS_ASSUME_NONNULL_BEGIN
 /// an `NSUserActivity` and passed to the delegate's
 /// `application:continueUserActivity:restorationHandler:`.
 ///
+/// Delivery waits for `BranchDidStartSessionNotification` so the link always
+/// follows the launch open, with a ten-second fallback. It used to fire on a
```
bboothe-branch: I think we can remove "It used to fire on a fixed delay, which made the order machine-dependent." from this comment. It talks about old behavior instead of new behavior.

Branch-TestBed/Branch-TestBed/TestBedDeepLinkTestHook.m
```diff
@@ -23,23 +24,53 @@ + (void)installIfRequested:(UIApplication *)application {
 
     NSLog(@"[TestHook] -testDeepLinkURL received: %@", testDeepLinkURL);
 
-    // Delay so the launch open is not still in flight when the activity
-    // arrives. Same 1.5s master uses, for a different reason: this line
-    // resolves links through requestDeepLinkData:, not initSession.
+    // Deliver once the launch session has started, never on a timer. A fixed
+    // delay makes the order machine-dependent: on a slow first install the
```
bboothe-branch: I believe we can remove "A fixed delay makes the order machine-dependent... etc." Talks about the old behavior instead of new behavior.

Branch-TestBed/Branch-TestBed/TestBedDeepLinkTestHook.m
```diff
@@ -23,23 +24,53 @@ + (void)installIfRequested:(UIApplication *)application {
...
+                usingBlock:^(NSNotification *note) {
+                    NSLog(@"[TestHook] Session started; delivering the link");
+                    deliver();
+                }];
+
+    // A session that never starts must not hang the test. Ten seconds is well
+    // past a healthy launch and still inside the driver's settle window.
     dispatch_after(
-        dispatch_time(DISPATCH_TIME_NOW, (int64_t)(1.5 * NSEC_PER_SEC)),
+        dispatch_time(DISPATCH_TIME_NOW, (int64_t)(10.0 * NSEC_PER_SEC)),
```
bboothe-branch, line 64: The count here is raised to 10 seconds but `settleSeconds` on ColdLinkWireValidationTest.swift:34 is set to 12 seconds. Can we raise to 12 seconds or lower `settleSeconds` to match?

Also ColdLinkWireValidationTest.swift:8 has a comment still saying delivery happens 1.5 seconds after launch.

bboothe-branch (approved):
Approved. Added some comments for some slight improvements to the comments and one about matching the new delay. Code overall looks good.

Branch-TestBed/Branch-TestBed/TestBedDeepLinkTestHook.h
wpinho-branch: Removed in a8cf74fe. Comments only, no code changed.

Branch-TestBed/Branch-TestBed/TestBedDeepLinkTestHook.m
wpinho-branch: Removed in a8cf74fe. Comments only, no code changed.

Branch-TestBed/Branch-TestBed/TestBedDeepLinkTestHook.m
wpinho-branch: Kept the fallback below the settle window instead of equal to it, since a fallback at 12s would leave no time for the resolve and the open. In a8cf74fe the fallback stays at 10s, settleSeconds is now 15s, and the stale 1.5s comments in ColdLinkWireValidationTest.swift, DeepLinkColdOpenHybridTest.swift and TESTING_GUIDE.md are updated.

PR #1648, opened 2026-08-28
EMT-4082: run the cold scenarios in the gate and report the Cold row (3 of 3)

wpinho-branch:
Runs the cold scenarios in the L1 gate and reports them in the Cold row of the step summary, by name rather than by plan ID.

Merges `feat/EMT-4082-field-contracts` at `45cf8f0c`, which carries the rename, and applies it to the workflow. Three conflicts, all resolved in favour of this branch's structure with the base's naming applied: the validate-step comment, the check block and the summary rows. The base's note about driving the default harness run was dropped, since this branch drives three.

Each capture filename moves in all five of its places: the producer, the check call, the summary grep, the failure `cat` and the artifact path. Three of those four consumers swallow a miss in silence, so a stale name drops a scenario from the public endpoint table without failing anything. That is why the rename reaches the workflow here and not only the script.

The uninstall note in `run_l1_instrumented.sh` now cites what actually decides install from open. It used to cite `/v1/install`, an endpoint this line never emits, since `BranchInstallRequest` is never instantiated under `Sources/`. The live criterion is `Branch.m:2226`, `BOOL isInstall = !self.preferenceHelper.randomizedBundleToken;`.

Suite unchanged at 94 tests. Step ids untouched, so the `if:` guards that turn a scenario failure into a job failure keep resolving.


wpinho-branch:
False positive: both cold test classes are compiled into the TestBed-GPTDriverTests target, and validate_l1_logs.py implements --scenario with C1 and C3 contracts.

bboothe-branch (approved):
Has some merge conflicts, so those need fixing. Otherwise code looks good.

wpinho-branch:
Merged feat/EMT-4082-field-contracts forward in 40ddfa67; no conflicts left.

wpinho-branch:
The gate failure on 5f59eda3 was the driver, not the contract: the install launch was terminated before BNCPreferenceHelper flushed randomizedBundleToken to disk, so the captured relaunch reran the install path and produced a cold_firstInstall shape. Pre-terminate settle raised to 20s in bd676bc0.

Caveat worth a reviewer: this is a timing budget, not a determinism guarantee. There is no exposed completion signal for "install response persisted" that the XCUITest host can poll, so the race could recur if this runner is ever slower than the L1WireValidationTest baseline.

PR #1649, opened 2026-09-02
EMT-4069 - Collapse link creation into BranchLinkBuilder

bboothe-branch:
## Reference
EMT-4069 -- Collapse link creation into BranchShortLinkBuilder.

## Summary

All link generation methods are removed from Branch.m and Branch.h. In their place, we have added the BranchLinkBuilder class to build URLs with a builder pattern in order to clean up our code.

## Motivation

We previously had 36 telescoping link-generation methods that funneled into five private implementation. This replaces all of it with one object.

 `BranchLinkBuilder` collects every link-generation option into a single object.

 This class is a plain mutable builder: create it, set the properties you care about, then call a
 terminal.

     BranchLinkBuilder *builder = [[BranchLinkBuilder alloc] init];
     builder.params = @{@"$og_title": @"Sale"};
     builder.channel = @"sms";
     builder.feature = @"share";

     [builder fetchShortURLWithCallback:^(NSString *url, NSError *error) {
         // Check `error`, not `url` — see "Choosing a terminal" below.
         if (error) return;
         // `url` is a short Branch link. This callback runs on the main queue.
     }];

 A builder is reusable: the terminals do not mutate it, so you can change a property and call
 another terminal.

 ## Choosing a terminal

 The verb tells you whether the call touches the network:

 - `-buildLongURL` — **offline**, returns immediately. The link's data is encoded into the URL, so
   it works with no connectivity, but the URL is long.
 - `-fetchShortURLWithCallback:` — **network**, non-blocking. Prefer this. The callback is delivered
   on the main queue.
 - `-fetchShortURL` — **network**, blocking. Never call it on the main thread; it freezes the UI for
   a full network round trip, or for the request timeout on a bad connection.
 - `-fetchSpotlightURLWithCallback:` — **network**, for Core Spotlight indexing. Reads `params` only.

 On a server error the two short-URL terminals hand back a **long-link fallback**, not nil — so a
 non-nil URL is not proof of success. Check `error`.

 ## Migrating from the `Branch` link methods (4.0 breaking change)

 Every method below was removed. Set the corresponding properties, then call a terminal. Options you
 do not set are simply omitted, so the short forms of the old overloads map to setting less.

 | Removed on `Branch`                                     | Replacement                                               |
 | ------------------------------------------------------- | --------------------------------------------------------- |
 | `getShortURL`, `getShortURLWithParams:…`                 | set the options, then `-fetchShortURL`                      |
 | `getShortUrlWithParams:…andCampaign:andMatchDuration:`   | also set `campaign` / `matchDuration`                       |
 | `getShortURLWithParams:…ignoreUAString:forceLinkCreation:` | also set `ignoreUAString` (see below re `forceLinkCreation`) |
 | `getShortURLWithParams:…andCallback:` (all async forms)  | `-fetchShortURLWithCallback:`                               |
 | `getLongURLWithParams:…`                                 | `-buildLongURL`                                             |
 | `getLongAppLinkURLWithParams:…`                          | set `useAppLinkDomain = YES`, then `-buildLongURL`          |
 | `getSpotlightUrlWithParams:callback:`                    | set `params`, then `-fetchSpotlightURLWithCallback:`        |

 Argument renames: `andType:` → `linkType`, `andMatchDuration:` → `matchDuration`, `andTags:` →
 `tags`, and so on for each `and…:` keyword.

 Three behavior changes to be aware of:

 1. **`forceLinkCreation:` is gone.** It was accepted and never read, so passing it changed nothing.
 2. **Long URLs now include `channel`.** `getLongURLWithParams:andChannel:…` and
    `getLongAppLinkURLWithParams:andChannel:…` accepted a channel and then dropped it before
    assembling the URL. That was a bug. `-buildLongURL` emits `channel=`, so a long link built with
    a channel set is not byte-identical to what 3.x produced.
 3. **There is no `Branch`-instance entry point.** Construct a builder directly; you do not need a
    `Branch` reference to make a link.

## Type Of Change

## Testing Instructions

- Set the appdelegate example to 1 in order to run the text examples.
- Review logs

## Testing Instructions
Success Logs: 
[Logs for BranchLinkBuilder.rtf](https://github.com/user-attachments/files/31968569/Logs.for.BranchLinkBuilder.rtf)


Sources/BranchSDK/BranchLinkBuilder.m
```diff
@@ -0,0 +1,351 @@
...
+    _feature = nil;
+    _stage = nil;
+    _campaign = nil;
+
+    // Link behavior
+    _matchDuration = 0;
+    _linkType = BranchLinkTypeUnlimitedUse;
+
+    // Terminal-specific options
+    _ignoreUAString = nil;
```
NidhiDixit09: `ignoreUAString` and `useAppLinkDomain` - both apply to exactly one terminal. So they should be part of method signature. This will reduce a lot of code -- if checks as well as warning messages. 
I would suggest to add follwoing overloaded methods - 
`buildLongURLUsingAppLinkDomain`
`fetchShortURLIgnoringUserAgent:(NSString *)userAgent;`

Sources/BranchSDK/BranchLinkBuilder.m
```diff
@@ -0,0 +1,351 @@
...
+}
+
+- (instancetype)initWithBranch:(Branch *)branch {
+    self = [super init];
+    if (!self) return self;
+
+    _injectedBranch = branch;
+
+    // Link content
+    _params = nil;
```
NidhiDixit09: Is it possible to use BranchLinkProperties data class instead of declaring each properties individually here ?

wpinho-branch (approved):
Approving. I read the full production diff: the builder is a faithful port of the funnel, the cache key and queue hops match, and dropping `forceLinkCreation` is safe since `generateShortUrl:` never read it.

Two follow-ups, neither blocking this PR.

`BNCPreferenceHelper.m`: the separator fix is load-bearing for `buildLongURL`, not incidental. Before it, `attributionLevelNone` skipped the separator entirely and a long URL came out as `.../a/<key>tags=x`. Nothing pins that, so a test at attribution NONE would stop it being undone.

`BNCServerRequestOperation.m`: exempting `BranchShortUrlRequest` from the attribution gate changes behaviour at level NONE, which is the level the N3 scenario in EMT-4088 contracts. The two need to stay in step.

Sources/BranchSDK/BranchLinkBuilder.m
```diff
@@ -0,0 +1,351 @@
...
+        [longUrl appendFormat:@"stage=%@&", [BNCEncodingUtils stringByPercentEncodingStringForQuery:self.stage]];
+    }
+
+    // Truthiness guards, not nil checks: BranchLinkTypeUnlimitedUse is 0 and a matchDuration of 0
+    // means "server default", so a default builder emits neither parameter.
+    if (self.linkType) {
+        [longUrl appendFormat:@"type=%ld&", (long)self.linkType];
+    }
+    if (self.matchDuration) {
+        [longUrl appendFormat:@"matchDuration=%ld&", (long)self.matchDuration];
```
wpinho-branch: Android ([ServerRequestCreateUrl.java](https://github.com/BranchMetrics/android-branch-deep-linking-attribution/blob/3fe6a6f9d4f89f7b72b8b14095ac48c0e2bc5824/Branch-SDK/src/main/java/io/branch/referral/ServerRequestCreateUrl.java#L242)), the [Web SDK](https://github.com/BranchMetrics/web-branch-deep-linking-attribution/blob/f8c869c86f3578a8be7c9f27b8fd16e4d99baaa0/src/1_utils.js#L122) and our own short link body ([BranchConstants.m](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/blob/b5f799f54cb38f9b4290e8ec12dcc16f47f8bc44/Sources/BranchSDK/BranchConstants.m#L29)) all spell this `duration`, so I'd converge on that instead of moving the fallbacks to `matchDuration` in 75e4a0cd.

```suggestion
        [longUrl appendFormat:@"duration=%ld&", (long)self.matchDuration];
```

Sources/BranchSDK/BranchShortUrlRequest.m
```diff
@@ -122,7 +122,7 @@ - (NSString *)createLongUrlForUserUrl:(NSString *)userUrl {
         [longUrl appendFormat:@"type=%ld&", (long)self.type];
     }
     if (self.matchDuration) {
-        [longUrl appendFormat:@"duration=%ld&", (long)self.matchDuration];
+        [longUrl appendFormat:@"matchDuration=%ld&", (long)self.matchDuration];
```
wpinho-branch: Same spelling point here. This fallback also still appends `data` without percent-encoding, unlike the sync fallback and `buildLongURL`, so wrapping it in `[BNCEncodingUtils urlEncodedString:]` a few lines below would keep a `+` in the payload from arriving as a space.

```suggestion
        [longUrl appendFormat:@"duration=%ld&", (long)self.matchDuration];
```

Sources/BranchSDK/BranchShortUrlSyncRequest.m
```diff
@@ -135,7 +135,7 @@ + (NSString *)createLongUrlWithBaseUrl:(NSMutableString *)baseUrl
     }
     
     [baseUrl appendFormat:@"type=%ld&", (long)type];
-    [baseUrl appendFormat:@"duration=%ld&", (long)duration];
+    [baseUrl appendFormat:@"matchDuration=%ld&", (long)duration];
```
wpinho-branch: Same here.

```suggestion
    [baseUrl appendFormat:@"duration=%ld&", (long)duration];
```

Sources/BranchSDK/BranchLinkBuilder.m
```diff
@@ -0,0 +1,351 @@
...
+        [longUrl appendFormat:@"channel=%@&", [BNCEncodingUtils stringByPercentEncodingStringForQuery:self.channel]];
+    }
+
+    if ([self.feature length]) {
+        [longUrl appendFormat:@"feature=%@&", [BNCEncodingUtils stringByPercentEncodingStringForQuery:self.feature]];
+    }
+
+    if ([self.stage length]) {
+        [longUrl appendFormat:@"stage=%@&", [BNCEncodingUtils stringByPercentEncodingStringForQuery:self.stage]];
+    }
```
wpinho-branch: Android emits `campaign` on long links too ([ServerRequestCreateUrl.java](https://github.com/BranchMetrics/android-branch-deep-linking-attribution/blob/3fe6a6f9d4f89f7b72b8b14095ac48c0e2bc5824/Branch-SDK/src/main/java/io/branch/referral/ServerRequestCreateUrl.java#L235)). Since long URLs already change here for `channel`, this could be the moment to add it and drop the caveat on the property in `BranchLinkBuilder.h`.

```suggestion
    if ([self.stage length]) {
        [longUrl appendFormat:@"stage=%@&", [BNCEncodingUtils stringByPercentEncodingStringForQuery:self.stage]];
    }

    if ([self.campaign length]) {
        [longUrl appendFormat:@"campaign=%@&", [BNCEncodingUtils stringByPercentEncodingStringForQuery:self.campaign]];
    }
```

wpinho-branch (commented):
A few follow-ups on the long URL format, mostly parity with Android, none of them blocking. The `data` percent-encoding is worth adding to the behavior changes in the description, since it changes long URLs the same way `channel` does.

Sources/BranchSDK/BranchActivityItemProvider.m
```diff
@@ -48,14 +49,15 @@ - (id)initWithParams:(NSDictionary *)params
...
-    NSString *url =
-        [[Branch sharedInstance]
-         getLongURLWithParams:params
-         andChannel:nil
-         andTags:tags
-         andFeature:feature
-         andStage:stage
-         andAlias:alias];
+    // No channel here, so this URL is unaffected by the builder's channel= fix -- unlike the other
+    // long-URL call sites, its output is byte-identical to what the old overload produced.
```
wpinho-branch: `buildLongURL` now percent-encodes `data`, so this URL is no longer byte-identical to the old overload. I'd drop the comment.

```suggestion
```

Branch-TestBed/Branch-TestBed/AppDelegate.m
```diff
@@ -150,9 +158,73 @@ - (BOOL)application:(UIApplication *)application
...
+    // Set user Alias Example
+
+//    [[Branch sharedInstance] setUserAlias:@"your_user_alias" completion:^(NSDictionary * _Nullable params, NSError * _Nullable error) {
+//        if (error) {
+//            NSLog(@"Error setting user alias: %@", error);
+//        } else {
+//            NSLog(@"Successfully set alias. Response: %@", params);
+//        }
+//    }];
+
```
wpinho-branch: nit: looks like a leftover from another example, unrelated to the builder.

```suggestion
```

Sources/BranchSDK/BranchLinkBuilder.m
wpinho-branch: Android has the same split: [`BranchShortLinkBuilder`](https://github.com/BranchMetrics/android-branch-deep-linking-attribution/blob/3fe6a6f9d4f89f7b72b8b14095ac48c0e2bc5824/Branch-SDK/src/main/java/io/branch/referral/BranchShortLinkBuilder.java#L29) takes one setter per field and [`BranchUniversalObject`](https://github.com/BranchMetrics/android-branch-deep-linking-attribution/blob/3fe6a6f9d4f89f7b72b8b14095ac48c0e2bc5824/Branch-SDK/src/main/java/io/branch/indexing/BranchUniversalObject.java#L384) maps `LinkProperties` onto it, which is what `linkBuilderWithLinkProperties:` does here.

Sources/BranchSDK/BranchLinkBuilder.m
```diff
@@ -0,0 +1,309 @@
...
+- (instancetype)init {
+    return [self initWithBranch:nil];
+}
+
+- (instancetype)initWithBranch:(Branch *)branch {
+    self = [super init];
+    if (!self) return self;
+
+    _injectedBranch = branch;
+    _params = nil;
```
NidhiDixit09: _params will be replaced by control params of BranchLinkProperties.

bboothe-branch:
@wpinho-branch Can you please me more descriptive here? I looked at EMT-4088 on Jira and I couldn't find an N3 scenario. Is the change in the google attached google doc? I didn't see it there. Specifically what are you suggesting and how does it get affected by EMT-4088?

"BNCServerRequestOperation.m: exempting BranchShortUrlRequest from the attribution gate changes behaviour at level NONE, which is the level the N3 scenario in EMT-4088 contracts. The two need to stay in step."

Sources/BranchSDK/BranchLinkBuilder.m
```diff
@@ -0,0 +1,309 @@
...
+
+    [[BranchLogger shared] logVerbose:@"Requesting Branch Link synchronously" error:nil];
+    BNCServerResponse *serverResponse = [req makeRequest:branch.serverInterface key:[Branch branchKey]];
+
+    // -processResponse: caches on a 200. Nothing is cached here, so the long-URL fallback it
+    // returns on a non-200 does not displace the short link a later call can still fetch.
+    return [req processResponse:serverResponse];
+}
+
+- (void)getShortURLWithParamsWithLinkProperties:(BranchLinkProperties *)linkProperties
```
NidhiDixit09: WithParams substring can be removed. It does not take params anymore.

Sources/BranchSDK/BranchActivityItemProvider.m
```diff
@@ -48,14 +49,16 @@ - (id)initWithParams:(NSDictionary *)params
...
 
-    NSString *url =
-        [[Branch sharedInstance]
-         getLongURLWithParams:params
-         andChannel:nil
-         andTags:tags
-         andFeature:feature
-         andStage:stage
-         andAlias:alias];
+    BranchLinkProperties *longURLProperties = [[BranchLinkProperties alloc] init];
```
NidhiDixit09, line 52: Copy constructor can be added in BranchLinkProperties class.

Sources/BranchSDK/BranchLinkBuilder.m
```diff
@@ -0,0 +1,317 @@
...
+//
+// @param linkProperties The link properties to read. May be nil.
+// @return The control params, or nil when there are none.
+- (NSDictionary *)paramsFromLinkProperties:(BranchLinkProperties *)linkProperties {
+    NSDictionary *controlParams = linkProperties.controlParams;
+    return controlParams.count ? controlParams : nil;
+}
+
+- (BNCLinkData *)linkDataWithLinkProperties:(BranchLinkProperties *)linkProperties
+                             ignoreUAString:(NSString *)ignoreUAString {
```
NidhiDixit09: Can we move this function easily to BNCLinkData/ BNCLinkCache class -- to Look up in the cache using Linkproperties.

Sources/BranchSDK/BranchLinkBuilder.m
```diff
@@ -0,0 +1,317 @@
...
+                // on a hit nothing else would, and callers expect one consistent queue.
+                dispatch_async(dispatch_get_main_queue(), ^{
+                    callback(cachedURL, nil);
+                });
+            }
+            return;
+        }
+
+        BranchShortUrlRequest *req =
+            [[BranchShortUrlRequest alloc] initWithTags:tags
```
NidhiDixit09: We can replace tags, alias etc directly with linkProperties.tags, linkProperties.alias etc.

wpinho-branch:
> @wpinho-branch Can you please me more descriptive here? I looked at EMT-4088 on Jira and I couldn't find an N3 scenario. Is the change in the google attached google doc? I didn't see it there. Specifically what are you suggesting and how does it get affected by EMT-4088?
> 
> "BNCServerRequestOperation.m: exempting BranchShortUrlRequest from the attribution gate changes behaviour at level NONE, which is the level the N3 scenario in EMT-4088 contracts. The two need to stay in step."

Sorry, that was unclear and it overstated things. Nothing needs to change here. N3 is in the scenarios list of the EMT-4088 description, and its contract lives in `scripts/validate_l1_logs.py`, whose comment says `BNCServerRequestOperation` drops everything at NONE except `BranchRequestDeepLink`. This PR makes that comment stale but not the contract, since N3 only counts `/v3/deeplink` and `/v3/events/open`. Your change also matches `BNCServerInterface`, which already allows `/v1/url` at NONE. I'll fix the comment on my side.

NidhiDixit09 (approved):

bboothe-branch:
Hey @wpinho-branch thanks for your explanation. I connected with Gabe on this and we're going to talk to Justin about how we want to handle this going forward. Will set up a new PR eventually to discuss this.

PR #1652, opened 2026-09-04
docs: cut CLAUDE.md to the rules and move reference into docs/

wpinho-branch:
Improvement to how this repo carries context for Claude Code, applying practices from the Flight School cohort together with Anthropic's CLAUDE.md guidance.

CLAUDE.md is read at the start of every conversation, so its length is a cost paid on every turn, and a file long enough to bury its own rules gets them ignored. This one was 316 lines.

What stays is what the guidance says earns a permanent slot: commands that cannot be guessed, conventions that differ from the defaults, and the gotchas whose absence causes a mistake. Reference material moves to `docs/` (`code-map`, `architecture`, `testing`, `releasing`), pointed at by plain path rather than `@`-import, since `@`-imports load recursively and would defeat the reduction.

Two stale claims are corrected against this tip: the Xcode pin in `layer1-logger-tests.yml` reads `^16.4`, not `"16"` (049f2530), and `gptdriver-release.yml` no longer exists (da692783), so the L1 workflow is now the only pinned one. Every other claim was re-verified against the source, and no content was dropped silently.

Docs only. A companion PR does the same on `4.0.0-beta.0`, where the architecture differs enough that none of this file transfers.


docs/architecture.md
```diff
@@ -0,0 +1,137 @@
...
+  `ForceCategoriesToLoad()`, so grepping for the symbol will find it and mislead).
+  `ForceCategoriesToLoad()` in `Branch.m` aggregates all five calls but has zero call sites in
+  the repo, so adding a line to it changes nothing. A new category needs the
+  `__attribute__((constructor))` form. Nothing in CI verifies category presence in the static
+  XCFramework, so this is convention, not a guarded invariant.
+- **`BNCInitSessionResponse` is built fresh on every callback path.** There are three separate
+  construction sites in `Branch.m`; a new field means updating all of them.
+- **`BNCURLFilter` self-updates from the server after a successful init**
+  (`updatePatternListFromServerWithCompletion:`), so the runtime skip list can differ from the
+  compiled-in default. Tests that assert on filtering must pin the pattern list.
```
jason-litton-branch, line 137: consider adding a diagram or two

jason-litton-branch (approved):
This is very nice cleanup

docs/architecture.md
wpinho-branch: Good call, I'll add two Mermaid diagrams (init into the queue, and the queue drain loop) in a follow-up PR so this one stays a straight split.

PR #1655, opened 2026-09-09
EMT-4287 - Move Public APIs on BranchConfiguration to private class

bboothe-branch:
## Reference
EMT-4287 - Move Public APIs on BranchConfiguration to private class.

## Summary

Moving the previously deprecated APIs now used in BranchConfiguration to private files.

## Type Of Change

## Testing Instructions

- Verify BranchConfiguration example in appDelegate still works by running app.

### DIFF
--- a/Sources/BranchSDK/Branch.m
+++ b/Sources/BranchSDK/Branch.m
@@ -168,9 +168,6 @@ void ForceCategoriesToLoad(void) {
 // Private method used internally
 - (void)clearLinkIdentifiers;
 
-+ (void)applyDeprecatedSettersFromConfiguration:(BranchConfiguration *)configuration
-                                        toBranch:(Branch *)branch;
-
 @end
 
 @implementation Branch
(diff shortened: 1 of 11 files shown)


wpinho-branch (approved):
Approving. The removals are all previously-deprecated setters and the private category keeps them reachable from the SDK and tests.

One optional follow-up, not a blocker: `+networkServiceClass` stays public while its setter moves private, so the surface can now be read but not written.

Sources/BranchSDK/Branch.m
```diff
@@ -830,12 +791,6 @@ + (void)setReferrerGbraidValidityWindow:(NSTimeInterval)validityWindow{
     }
 }
 
-+ (void) setDMAParamsForEEA:(BOOL)eeaRegion AdPersonalizationConsent:(BOOL)adPersonalizationConsent AdUserDataUsageConsent:(BOOL)adUserDataUsageConsent{
```
wpinho-branch, line 833: The TestBed copy of the class tests still calls this ([BranchClassTests.m](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/blob/e87deba91585c5f126cf34d12c86e38416bf77f4/Branch-TestBed/Branch-SDK-Tests/BranchClassTests.m#L231)) and other setters now in `Branch+Configuration.h`, so the `Branch-SDK-Tests` target stops compiling even though CI, which doesn't build it, is green. Mirroring the `BranchSDKTests` changes there would fix it.

PR #1656, opened 2026-09-11
EMT-4313: resolve at launch in the TestBed and correct the install contract

wpinho-branch:
The TestBed calls `[Branch initialize:config]` and wires the foreground overloads, but never calls `requestDeepLinkDataWithLaunchOptions:` in `didFinishLaunchingWithOptions:`. Both 4.0.0-beta integration documents state that call is required rather than optional, because empty launch options still mean "does this device have a pending link from a prior click".

So the harness has been measuring an integration the docs do not describe, and the merged contract for that run was written to match the wrong shape. It declares `{"/v3/events/open": 1, "/v3/deeplink": 0}`, which forbids the resolve a correct integration issues on every launch.

## What changed

One line in the TestBed AppDelegate, the resolve at launch. The contract for the harness run is corrected to one resolve then one open, with the order asserted, and its key is `install` rather than a plan ID, because `run_l1_instrumented.sh` uninstalls the bundle before every run, so no `randomizedBundleToken` persists and what the harness drives is an install. The fixtures and the validator self-tests that encoded the old shape are updated to match, along with one comment in the workflow that still described it. `organic_open` stays uncontracted here: this harness cannot produce it, and the two design sources disagree on what it should emit.

The two halves cannot be split. With the resolve wired and the old contract, the gate fails; with the new contract and no resolve, it fails the other way.

## Naming the scenarios in full

`N3` becomes `attribution_none`, the name EMT-4041 gives it. The `install` key was already in place, from the same evidence above, so what moves around it now is the surfaces: three fixtures, the test class, and one method.

That method is worth a line. `test_a_clean_organic_open_passes` becomes `test_a_clean_install_capture_passes`, because the old name told the next reader that organic opens are contracted here, which is the opposite of what this PR establishes, and a review comment had flagged the pairing twice.

The note above the registry said plan scenarios use their plan ID, which is false once the IDs are gone, so it now names the members of each class instead.

## Proof

Recaptured end to end on this branch: `xcodebuild build-for-testing` then `scripts/run_l1_instrumented.sh`, iPhone simulator, 210 lines pulled from the app container. The same capture judged by both contracts: this branch's validator with `--scenario install` exits 0 at `VALIDATION PASSED (2/2 requests valid)`, and the validator from `4.0.0-beta.0` exits 1 with `FAILED: '/v3/deeplink' must not be captured for this scenario, but appeared 1 time(s)`. That pair is the argument.

Suite unchanged at 74 tests, and `Sources/` is untouched.

## What this unblocks

EMT-4088, the iOS negative scenarios, is blocked on this. Authoring them against the merged contract would contract a TestBed defect.


.github/workflows/layer1-logger-tests.yml
bboothe-branch:
```suggestion
            echo "The install scenario is authored along with the endpoint table assertion."
```

We can say that the install scenario is authored now.

bboothe-branch (approved):
Great catch here by adding requestDeepLinkDataWithLaunchOptions to appDelegate. Approved. One comment on the  layer1-logger-tests.yml added

.github/workflows/layer1-logger-tests.yml
wpinho-branch: Applied in 08559775.

Branch-TestBed/Branch-TestBed/AppDelegate.m
```diff
@@ -146,6 +146,9 @@ - (BOOL)application:(UIApplication *)application
     // ── Initialize ───────────────────────────────────────────────────────
     [Branch initialize:config];
 
+    // Required. Covers the deferred check and push cold launch.
+    [[Branch sharedInstance] requestDeepLinkDataWithLaunchOptions:launchOptions callback:nil];
```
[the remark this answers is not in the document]
wpinho-branch: The install contract was re-measured earlier in this PR. The deeplink contract predated the launch resolve and the gate does not run it; re-measured it on this AppDelegate in fb07b35c (deeplink, open, deeplink, open) and updated it to two resolves and two opens.

wpinho-branch:
Merged 4.0.0-beta.2 forward in 897eea05. The cold contracts that landed with #1646 and #1647 were measured before this PR adds the launch resolve, so both cold captures now emit four requests instead of three: deeplink, open, deeplink, open. Re-measured on the merged tree and corrected both to two resolves in 78231b16, with the fixtures updated to match.

wpinho-branch:
Re-requesting review: three commits landed after the approval on 88b51e1d. fb07b35c re-measures the deeplink contract, 897eea05 merges beta.2 forward, and 78231b16 corrects cold_https and cold_firstInstall to two resolves, since the launch resolve this PR adds emits its own pair ahead of the link-driven one.

PR #1658, opened 2026-09-14
EMT-4356: isolate the drain tests from the host app's lifecycle notifications

wpinho-branch:
`testLiveResolutionSuppressesTheLifecycleOrganicOpen` failed intermittently in CI and passed on re-run with identical code. The Branch singleton registers its `WillResignActive` and `DidBecomeActive` observers once per process, so a real activation of the test host app reaches the SDK mid-test. When it lands before `requestDeepLinkData:` enqueues the resolution, `containsInstallOrOpen` sees an empty queue and an organic `BranchRequestOpen` is enqueued, which fails the precondition, the post-foreground check, and the open count.

The drain tests now remove those two observers in `setUp` and restore them in `tearDown`, so only the test's own lifecycle calls reach the SDK.

Reproduced by posting `UIApplicationDidBecomeActiveNotification` before `requestDeepLinkData:`: all three CI failures appeared, with the queue reading `BranchRequestOpen, BranchRequestDeepLink`. With this change and the same synthetic notification, the test passes. Full suite: 594 tests, 0 failures.

Verified in the full suite only. Running this class alone on a cold user-agent cache still deadlocks on the `dispatch_sync` isolation-queue barrier, tracked separately as EMT-4348.


wpinho-branch:
Already covered on 4.0.0-beta.0 by b5f799f5, which detaches and restores the same two lifecycle observers. The isolation-queue fix continues in #1659.

PR #1659, opened 2026-09-14
EMT-4348: bound the isolation-queue barrier in the drain tests

wpinho-branch:
Stacked on #1658, so the base is `fix/EMT-4356-drain-test-lifecycle-isolation`. It retargets to `4.0.0-beta.0` once #1658 merges.

## Defect

`waitForIsolationQueue` in `BranchDeepLinkQueueDrainTests` blocked the main thread with `dispatch_sync` onto the SDK's serial isolation queue. Any isolation-queue block that needs main deadlocks against that barrier with no timeout, so the job hangs instead of failing. The natural trigger is `-loadUserAgent`, which waits on a semaphore signalled from a WKWebView completion on main whenever the user agent is not cached.

## Fix

The helper now enqueues a flag block with `dispatch_async` and waits on it through the file's existing `waitForCondition:description:timeout:`, with a 15s bound, the file's drain ceiling. The serial queue still orders everything enqueued before the call, and main keeps servicing its own queue while it waits. One method changed, and both call sites are unchanged. There is no other main-thread `dispatch_sync` onto the isolation queue in `BranchSDKTests`.

## Evidence

- Red: a temporary probe that enqueues `dispatch_sync(dispatch_get_main_queue(), ...)` on the isolation queue and then calls the old helper exceeds the 60s execution time allowance.
- Green: the same probe passes in 1.09s with the fix. With the probe removed, `BranchDeepLinkQueueDrainTests` passes 2/2 and the full `BranchSDKTests` suite passes 594/594.
- The natural `-loadUserAgent` hang was sampled once but depends on timing and did not reproduce on demand, so the red receipt comes from the probe rather than from that path.

The probe is not committed, because it exercises a test helper rather than SDK behaviour.


BranchSDKTests/BranchDeepLinkQueueDrainTests.m
bboothe-branch, line 191: This comment can be updated to the new `dispatch_async` + poll  schema.

bboothe-branch (approved):
Approved changes. Left one comment on Line 191 of BranchDeepLinkQueueDrainTests.m that we should update before merging.

bboothe-branch:
Might be interesting to consider adding a regression test to avoid someone reintroducing a dispatch_sync based wait in the future.

BranchSDKTests/BranchDeepLinkQueueDrainTests.m
wpinho-branch: Updated in 2acd54ce to describe the dispatch_async marker and the poll. Comments only, no code changed.

wpinho-branch: After merging 4.0.0-beta.2 in 3f4d6972 this comment is gone: #1660 moved absorbPendingIsolationQueueWork into BranchLifecycleTestIsolation.m, whose wait is already dispatch_async based.

wpinho-branch:
Agreed, tracked as a follow-up in EMT-4431 so this PR stays scoped to the barrier.

PR #1661, opened 2026-09-16
EMT-4362: cover the foreground open around a deep link resolve (1 of 3)

wpinho-branch:
## Why this lands before the fix

EMT-4362 ships as three PRs. This one adds only the test class the other two need, and it is green on this branch line as it stands, so the harness can be reviewed on its own: whether the duplication is justified, and whether each new test earns its place. The fix PRs then arrive as failing-first diffs against a harness that already runs.

## What it adds

`BranchLifecycleOpenResolveTests` drives `applicationDidBecomeActive` against a stubbed server interface and counts what reaches the wire, for a launch that resolves a deep link.

- **Guard (e)**, organic resolve without `~referring_link`, foreground check after both requests drained: exactly one `/v3/events/open`.
- **Guard (d)**, nil-URL resolve whose response carries `~referring_link`, foreground check while the resolve is live: exactly one open, and it carries `link_data`.

## Why these two tests, and not the existing ones

`BranchDeepLinkQueueDrainTests` covers a link resolve with a URL. Neither case here is reachable from it: a nil `urlString` sends `attemptToSendOpen` down the response branch (`BranchRequestDeepLink.m:285` to `:298`) rather than `:282`, and no existing test asserts `link_data` on the resulting open or evaluates the foreground check against an empty queue.

Guard (d) passes today for a reason that will stop being true: it pins the wire order, and its distinctness becomes load-bearing in (2 of 3), where a deferred check that is not cancelled would occupy the queue and add a second open.

## Isolation

The class does not reuse the drain tests' helpers: their `setUp` barrier is a main-thread `dispatch_sync` onto the isolation queue, which deadlocks against `loadUserAgent` (EMT-4348, #1659). This class waits on a bounded `XCTWaiter` instead, detaches and reattaches the two lifecycle observers, and saves and restores everything `clearLinkIdentifiers:` resets.

## Receipts, on `6fead0d4`

- Targeted, iPhone 16 Plus iOS 18.4, this class plus `BranchDeepLinkQueueDrainTests`: 4 tests, 0 failures.
- Full suite, iPhone 16e iOS 26.3 (what `scripts/getSimulator` picks, and the runtime CI used on the last green `verify`): 674 tests, 0 failures.
- Full suite, iPhone 16 Plus iOS 18.4: 674 tests, 2 failures, both in `BranchConfigurationControllerTests`. They reproduce on `4.0.0-beta.2` with no local changes, on that runtime only, and are tracked as EMT-4372.

No production code is touched. Five `-Wnullability-completeness` warnings from `Private/BranchRequestOpen.h` appear in any test file that imports it, on base as well.


bboothe-branch (approved):
Approved.

PR #1662, opened 2026-09-16
EMT-4362: send one open when an organic launch resolves at launch (2 of 3)

wpinho-branch:
Stacked on #1661 (1 of 3), the harness this fix runs on.

## Defect

An app that calls `requestDeepLinkDataWithLaunchOptions:` in `didFinishLaunchingWithOptions:`, as the integration guide asks, sends no `/v3/events/open` on an organic relaunch. The launch enqueues a resolve with a nil `urlString`. The foreground check sees it in the queue and skips its open, and the resolve's response carries no `~referring_link`, so it chains none. No session callback fires for that launch.

## Fix

When the foreground check finds a live nil-URL resolve and no install or open, it adds one check operation that depends on that resolve. When the resolve finishes, the check re-reads the queue (ignoring finished and cancelled operations) and, if no open was chained, sends the open from the isolation queue, the same place the base open is sent from. Enqueuing an install or open cancels the check, as does cancelling the pending resolves. Nothing is stored: the operation leaves the queue by running or by being cancelled, and its only link to the resolve is `addDependency:`.

## Why not smaller

- Filtering the existing check alone still sees the queued resolve: 0 opens (test a).
- Excluding nil-URL resolves from the check: a deferred link sends 2 (test d).
- Sending from the resolve's no-link branch: a fresh install sends 2 (test e).

## Evidence on `ac92e947`

iPhone 16 Plus, iOS 18.4, TestBed with the #1656 launch resolve applied locally, 10 runs per row, base `8978bb26`:

| Scenario | Base | Fix |
|---|---|---|
| Organic relaunch | 1 deeplink, 0 opens | 1 deeplink, 1 open with `user_agent` |
| Organic fresh install | 1 open with `user_agent` | 1 open with `user_agent` |
| Universal Link relaunch | 1 open with `link_data` and `user_agent` | unchanged |

- Failing first on #1661's tip: (a) organic relaunch, (c) Spotlight activity with no URL, (f) failed resolve with `dropURLOpen` off, (g) replaced resolve, and a test that no open is sent while the isolation queue is held. That last one exists because sending from the request queue lost `user_agent` on a fresh install, 10 of 10.
- Guards on the operation, which no wire count sees: a chained open cancels the check, none is added while an open is queued, and two activations send one open (50 of 50).
- Full suite: 681 tests, 2 failures, both in `BranchConfigurationControllerTests`. They are pre-existing on iOS 18.4 and tracked as EMT-4372.

## Limits

- Spotlight and the failed resolve have unit-test evidence only. CI's iOS 26.2, physical devices and apps without the TestBed's launch link builder are unmeasured, and tvOS is covered by CI's builds only.
- A duplicate is still possible if a new link resolve and its chained open both finish inside the first-launch user-agent wait, the same class of race base already has. A nil-URL resolve enqueued after the check is added still loses its open.
- On an organic relaunch the open now goes out after the launch's Normal-priority events, where base sent none. The order was not measured.


bboothe-branch (approved):
Approved but a couple findings. Hopefully addressed in Part 3.

1. Attribution-None re-read is untested - shouldSendDeferredForegroundOpen (Branch.m:1685) re-checks both attributionLevelNone and the newly-added bnc_disableAutomaticOpenTracking, but no test flips either while a resolve is queued at the foreground. 
2. The "resolve executing, not just queued" race is untested again - this was actually built in the commit history with `testNaturallyFinishingResolveStillSendsOneOpen`, but deleted in commit ac92e947 because it didn't distinguish the fix from a broken mutant. If this omission is deliberate all good.
3. Guard (b) is still missing from the same lettering used elsewhere in the PR body - likely deferred to the not-yet-opened "3 of 3" PR, but that's an inference, not something stated in this PR's description.

wpinho-branch:
1. Covered in a0b10ec3: testAttributionLevelNoneAtTheDeferredReReadSendsNoOpen and testAutomaticOpenTrackingDisabledAtTheDeferredReReadSendsNoOpen flip each setting while the re-read is queued behind the held isolation queue; each fails with its check removed.
2. Deliberate: it passed 50 of 50 against the mutant in ac92e947, so it could not tell the fix from the defect.

PR #1665, opened 2026-09-16
EMT-4086: run hot_uriScheme in the gate and report the Hot row (3 of 4)

wpinho-branch:
Stacked on #1664. Runs `hot_uriScheme` in the L1 gate and reports a Hot row: delivers a scheme URL to the running TestBed with `scripts/run_h2_hot.sh`, then checks the delivery delta against the contract (one `/v3/deeplink`, exactly one `/v3/events/open`) and against the foreground markers.

## Review follow-up: the carve-out

The review is right that a runner without the pinned runtime should not fail the gate. The marker was already written and nothing read it; it is now read ahead of the capture-outcome guard.

The literal `exit 0` alone would have been worse than the bug. It makes `steps.h2.outcome` equal `success`, and the summary if-chain tested that **before** the marker. Running the unreordered chain verbatim, marker present, outcome forced to `success`:

```
| Hot | hot_uriScheme pass; ... |
Result: **PASS**
```

A green required check for a scenario that never ran. The chain is therefore reordered in the same commit to test the marker first, giving `| Hot | hot_uriScheme not run: ... |` and `Result: **PASS** (hot_uriScheme not run)`. The capture step also `rm -f`s the marker first, so a stale one cannot disarm the gate while still rendering "not run".

## Verification

Guard block extracted verbatim from the patched step, run under `bash -e`:

| marker  | capture | exit | output                                          |
| ------- | ------- | ---: | ----------------------------------------------- |
| present | failure |    0 | `not run: no simulator on the expected runtime` |
| absent  | failure |    1 | `FAILED: [hot_uriScheme] driver failed`         |
| absent  | success |    0 | falls through to the validators                 |

Row 2 is the one that matters: a genuine failure is still red.

End to end on scratch branch `test/EMT-4086-h2-carveout-proof` (SHAs differ from this head; each run needs a different pinned value):

- Missing runtime (`iOS-99-9`), `51fd320a`: [run 35796375033](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/actions/runs/35796375033), **success**. Capture logged `ERROR: no available 'iPhone 16 Plus' on runtime iOS-99-9`; validate logged `not run`, exited 0; fail step skipped.
- Genuine failure past runtime selection (uninstalled bundle), `b8058625`: [run 35797522525](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/actions/runs/35797522525), **failure**. Capture logged `RUNTIME=iOS-18-5` then `ERROR: driver exited 4`, so no marker; validate exited 1 and the fail step failed the job.

Green alone proves nothing here: green is what the gate already gave whenever the runtime was present.

## Merge forward

`4.0.0-beta.2` at `31948b5c` merged in. Three workflow conflicts, none "take ours": the endpoint-table preamble is dropped (it named one capture and two contracts, now three and four); the Hot block loses its trailing `Negative and guards` echo, which beta already emits from the Cold if/else; the wire-log block takes beta's three-file render inside this branch's wrapper.


bboothe-branch (approved):

.github/workflows/layer1-logger-tests.yml
```diff
@@ -129,6 +150,36 @@ jobs:
...
+      # proves the app stayed foreground across it. Both print payloads or
+      # counts, so only their verdict lines reach the log.
+      - name: Validate hot_uriScheme
+        id: h2
+        continue-on-error: true
+        env:
+          OUTPUT_DIR: /tmp/h2
+        run: |
+          tag="[hot_uriScheme]"
+          if [ "${{ steps.h2capture.outcome }}" != "success" ]; then
```
bboothe-branch, line 212: This treats a missing simulator the same as a real validation failure. h2capture already sets /tmp/h2-runtime-missing when the driver's error was "no available simulator on runtime" (see the capture step above), but this guard doesn't check it — any h2capture failure unconditionally becomes exit 1 here, which makes steps.h2.outcome == 'failure', which then fails the whole job via "Fail the job if hot_uriScheme failed" below. That contradicts this PR's own description: "so an image change is not read as an SDK failure." The summary row correctly prints "not run," but the required check still goes red.

Suggested fix:

if [ -f /tmp/h2-runtime-missing ]; then
   echo "$tag not run: no simulator on the expected runtime" | tee /tmp/h2-output.txt
   exit 0
 fi
 if [ "${{ steps.h2capture.outcome }}" != "success" ]; then
   echo "FAILED: $tag driver failed; see the capture step log" | tee /tmp/h2-output.txt
   exit 1
 fi

bboothe-branch (changes requested):
Found one issue.

.github/workflows/layer1-logger-tests.yml
wpinho-branch: Carved out in c6993435: a missing runtime now skips the gate instead of failing the job.

PR #1668, opened 2026-09-17
EMT-4084: run warm_uriScheme in the gate and report the Warm row (2 of 2)

wpinho-branch:
Stacked on #1667. Runs `warm_uriScheme` in the L1 gate and reports the Warm row: backgrounds the TestBed, brings it forward, delivers the scheme URL, then checks the delivery delta against the warm contract and marker rule.

This is the explicit `--scenario` caller: the marker checker gets the warm value rather than its default, and a test asserts that argument, so dropping it fails the suite instead of silently validating a warm capture against the foreground rule.

**An honest gap:** the hot capture passes the warm contract, by design, since that contract is bounded rather than exact while the open count is unpinned. `cold_https` proves it binds.

## Review follow-up: the carve-out

A runner without the pinned runtime should not fail the gate. Both capture steps already wrote a marker and neither was read; each is now read ahead of its capture-outcome guard. This branch carries the hot step too, so both are fixed.

A bare `exit 0` alone would be worse than the bug: it makes the outcome `success`, and both if-chains tested that **before** the marker, so a runner with no iOS 18.5 would render `pass` and `Result: **PASS**` for a scenario that never ran. Both chains now test the marker first. Each capture step also `rm -f`s its own marker, so a stale one cannot disarm the gate while rendering "not run".

## Verification

Each guard block extracted verbatim from its patched step and run under `bash -e`:

| marker  | capture | exit | output, identical for both, tagged per scenario              |
| ------- | ------- | ---: | ------------------------------------------------------------ |
| present | failure |    0 | `[<scenario>] not run: no simulator on the expected runtime` |
| absent  | failure |    1 | `FAILED: [<scenario>] driver failed`                         |
| absent  | success |    0 | falls through to the validators                              |

Row 2 is the one that matters: a genuine failure is still red.

The summary step was rendered for all four marker combinations, each exiting 0, giving `PASS`, `PASS (hot_uriScheme not run)`, `PASS (warm_uriScheme not run)` and `PASS (hot ... and warm ... not run)`. The explicit `if` form is deliberate: `[ -f x ] && y=1` returns 1 as the last command under `bash -e`.

End to end in CI, on the hot half, scratch branch `test/EMT-4086-h2-carveout-proof` off #1665 (SHAs differ from this head):

- Missing runtime (`iOS-99-9`), `51fd320a`: [run 35796375033](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/actions/runs/35796375033), **success**. Validate logged `not run`, exited 0, fail step skipped.
- Genuine failure past runtime selection, `b8058625`: [run 35797522525](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/actions/runs/35797522525), **failure**. Capture logged `RUNTIME=iOS-18-5` then `ERROR: driver exited 4`, so no marker, and the fail step failed the job.

The warm half is identical in shape, covered by the rows above, not its own CI run.


.github/workflows/layer1-logger-tests.yml
```diff
@@ -182,6 +205,36 @@ jobs:
...
+      # transition marker in pre, which is what tells a warm delivery from a hot one.
+      - name: Validate warm_uriScheme
+        id: w2
+        continue-on-error: true
+        env:
+          OUTPUT_DIR: /tmp/w2
+          H2_URL: branchtest://open?scenario=W2
+        run: |
+          tag="[warm_uriScheme]"
+          if [ "${{ steps.w2capture.outcome }}" != "success" ]; then
```
bboothe-branch, line 234: Same gap as the h2 step in #1665 (flagged there): this doesn't check /tmp/w2-runtime-missing before treating any w2capture failure as exit 1, so a missing simulator fails the whole job here too, despite the PR description saying "an image change is not read as an SDK failure." Needs the same carve-out:

```
 if [ -f /tmp/w2-runtime-missing ]; then
   echo "$tag not run: no simulator on the expected runtime" | tee /tmp/w2-output.txt
   exit 0
 fi
```
 
Since this is now the second copy of the same pattern (and Cold will likely be a third), might be worth factoring the runtime-missing check into a shared step/script instead of duplicating it per scenario.

bboothe-branch (changes requested):

.github/workflows/layer1-logger-tests.yml
wpinho-branch: Carved out in ee2fb09d, same as #1665.

